PR #18, opened 2019-10-09
initial commit

nekufa:
@bobstrecansky please review current code state

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,33 @@
+{
+    "name": "open-telemetry/opentelemetry",
+    "type": "library",
+    "license": "Apache-2.0",
+    "require": {
+        "php": "^7.1"
+    },
+    "authors": [
+        {
+            "name": "Bob Strecansky",
+            "email": "bob.strecansky@gmail.com"
+        },
+        {
+            "name": "Dmitry Krokhin",
+            "email": "nekufa@gmail.com"
+        }
+    ],
+    "autoload": {
+        "psr-4": {
+            "OpenTelemetry\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "": "tests"
+        },
+        "classmap": ["tests"]
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^7.5.16",
+        "composer/xdebug-handler": "^1.3"
+    }
+}
--- a/src/Exporter.php
+++ b/src/Exporter.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry;
+
+use OpenTelemetry\Tracing\Span;
+use OpenTelemetry\Tracing\Tracer;
+
+abstract class Exporter
+{
+    abstract public function convertSpan(Span $span) : array;
+
+    public function flush(Tracer $tracer, Transport $transport) : int
+    {
+        $data = [];
+
+        foreach ($tracer->getSpans() as $span) {
+            $data[] = $this->convertSpan($span);
+        }
+
+        if (count($data)) {
+            $transport->write($data);
+        }
+
+        return count($data);
+    }
+}
--- a/src/Tracing/Span.php
+++ b/src/Tracing/Span.php
@@ -1,0 +1,162 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tracing;
+
+use Exception;
+
+class Span
+{
+    private $name;
+    private $spanContext;
+    private $parentSpanContext;
+
+    private $start;
+    private $end;
+    private $status;
+
+    private $attributes = [];
+    private $events = [];
+
+    public function __construct(string $name, SpanContext $spanContext, SpanContext $parentSpanContext = null)
+    {
+        $this->name = $name;
+        $this->spanContext = $spanContext;
+        $this->parentSpanContext = $parentSpanContext;
+        $this->start = microtime(true);
+    }
+
+    public function getSpanContext() : SpanContext
+    {
+        return $this->spanContext;
+    }
+
+    public function getParentSpanContext() : ?SpanContext
+    {
+        return $this->parentSpanContext;
+    }
+
+    public function setParentSpanContext(SpanContext $parentSpanContext = null) : self
+    {
+        $this->parentSpanContext = $parentSpanContext;
+        return $this;
+    }
+
+    public function end(Status $status = null) : self
+    {
+        $this->end = microtime(true);
+        if (!$this->status && !$status) {
+            $status = new Status(Status::OK);
+        }
+        if ($status) {
+            $this->setStatus($status);
+        }
+        return $this;
+    }
+
+    public function getStart()
+    {
+        return $this->start;
+    }
+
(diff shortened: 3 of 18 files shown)


src/Exporter.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry;
+
+use OpenTelemetry\Tracing\Span;
+use OpenTelemetry\Tracing\Tracer;
+
```
bobstrecansky: not necessarily for this PR, but:

Should we be adding docstrings to our classes?  I feel as if they are very helpful (and might be even more helpful for future committers): 

http://phpdocu.sourceforge.net/howto.php

src/Exporter.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
```
bobstrecansky: I'm assuming we are planning on declaring strict types for each of our files?  This seems reasonable to me at first glance, just wanted to get a temperature check.

bobstrecansky:
You'll also probably need to sign the CLA. ^

composer.json
```diff
@@ -0,0 +1,33 @@
...
+    "name": "open-telemetry/opentelemetry",
+    "type": "library",
+    "license": "MIT",
+    "require": {
+        "php": "^7.1"
+    },
+    "authors": [
+        {
+            "name": "Bob Strecansky",
+            "email": "bob@mailchimp.com"
```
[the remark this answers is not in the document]
nekufa: Sure, fixed

src/Exporter.php
nekufa: i'm not sure that is really needed right now, i think we will get back to this later, because this can be helpful for api generation.

src/Exporter.php
nekufa: Yes, i think that gives us some advantages for non-typed language

nekufa:
> You'll also probably need to sign the CLA. -----v

I've registered on https://identity.linuxfoundation.org/projects/cncf and make electronic signature by email. Right now i see an information block on the website
```
Linux Foundation ID is migrating to an updated platform.
For more details, please read our FAQ.
```
Is there any else i should do? Or everything is in it's right place?

src/Exporter.php
bobstrecansky: It can also be helpful for new maintainers of the project!

composer.json
```diff
@@ -0,0 +1,33 @@
+{
+    "name": "open-telemetry/opentelemetry",
+    "type": "library",
+    "license": "MIT",
```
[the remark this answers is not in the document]
nekufa: Thanks, that was invalid line. I think license should be Apache-2.0 as other open-telemetry libraries

SergeyKanzhelev:
@nekufa I'm creating an issue in CNCF support tool. Can you please confirm that you've been using the same e-mail to sign off CLA as used in commits. I see `<first-name>.<last-name>@basis.company` in the commit details.

SergeyKanzhelev:
I will also close and reopen the PR so CLA will be recalculated

idvoretskyi:
@nekufa have you signed the individual's CLA or a corporate one?

Also, please note that all CLA-related requests should go to the LF HelpDesk (https://github.com/kubernetes/community/blob/master/CLA.md#troubleshooting).

nekufa:
@idvoretskyi i signed with my business account - dmitry.krokhin@basis.company

src/Tracing/Span.php
```diff
@@ -0,0 +1,162 @@
...
+    private $parentSpanContext;
+
+    private $start;
+    private $end;
+    private $status;
+
+    private $attributes = [];
+    private $events = [];
+
+    public function __construct(string $name, SpanContext $spanContext, SpanContext $parentSpanContext = null)
```
morrisonlevi: Nit: the technically correct type is: `?SpanContext`. The `?` is allowed to be omitted for now, but I would not be surprised at all if in the future it was required, or if linters/PSR type things started requiring it.

src/Exporter.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenTelemetry;
+
+use OpenTelemetry\Tracing\Span;
+use OpenTelemetry\Tracing\Tracer;
+
+abstract class Exporter
+{
+    abstract public function convertSpan(Span $span) : array;
```
morrisonlevi: I would prefer that we don't have to convert to an array of arrays, and then almost certainly have to do *another* conversion to transport it. Is this approach mandated/suggested by the specification? I don't see it in the Tracing API -- is it somewhere else maybe?

morrisonlevi: After thinking about it more, I realized an implementation could override the `flush` function as well, so maybe it's a non-issue. This way it could detect concrete `Transports` and see if they can handle direct conversions from `Span`s in a vendor specific API.

morrisonlevi: The [Tracing SDK Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/c9d3d439492e07ca900cf28f98552a8aff919ff5/specification/sdk-tracing.md) says this about the [Span Exporter](https://github.com/open-telemetry/opentelemetry-specification/blob/c9d3d439492e07ca900cf28f98552a8aff919ff5/specification/sdk-tracing.md#span-exporter):

> The exporter must support two functions: Export and Shutdown. In strongly typed languages typically there will be 2 separate Exporter interfaces, one that accepts spans (SpanExporter) and one that accepts metrics (MetricsExporter).

This doesn't seem to follow the specification, unless there is yet another place I should be looking. I wouldn't discount that possibility, given how new I am to the project.

src/Tracing/Span.php
bobstrecansky: +1 for this; following the spec here would be smart for future compatibility.

src/Exporter.php
bobstrecansky: I don't think there is another place.  I think we may want to follow the spec a little more deliberately here, which we could update in subsequent PRs.

bobstrecansky:
@idvoretskyi - wanted to make sure you saw @nekufa's comment above!

idvoretskyi:
@nekufa may I ask you to reach out to the LF HelpDesk directly - https://github.com/kubernetes/community/blob/master/CLA.md#troubleshooting

nekufa:
> @nekufa may I ask you to reach out to the LF HelpDesk directly - https://github.com/kubernetes/community/blob/master/CLA.md#troubleshooting

@idvoretskyi 

### the registration can't be complete, error has no details, just occurred and that's all
image

### i can't register because the email or username is registered (should be unique).
image

### i receive password restore link, but when i set it i see an error that email is not registered
image

this dissapoint me, i still have no chance to contribute to opentelemetry
is there another way to see that CLA is signed at hellosign and relate the sign  with my cncf account? can you reach me via email or telegram, so we can solve this issue shortly?

idvoretskyi:
@nekufa as mentioned above, please email the LF HelpDesk via the dedicated email address listed is the "troubleshooting" subsection - https://github.com/kubernetes/community/blob/master/CLA.md#troubleshooting.

They are happy to assist.

PR #23, opened 2019-11-07
Refactor initial implementation

morrisonlevi:
This is a partial improvement to the implementation based on a reading of the spec between Levi Morrison and Bob Strecansky.

### DIFF
--- a/src/Tracing/Status.php
+++ b/src/Tracing/Status.php
@@ -46,6 +46,7 @@ class Status
 
     private $code;
     private $description;
+    static private $ok;
 
     public function __construct(int $code, string $description = null)
     {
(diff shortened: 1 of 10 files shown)


src/Tracing/Status.php
```diff
@@ -46,6 +46,7 @@ class Status
 
     private $code;
     private $description;
+    static private $ok;
 
     public function __construct(int $code, string $description = null)
```
morrisonlevi: This should probably be marked private now, so everything uses `new`.

morrisonlevi:
Just now I believe I've signed the CLA -- let me know if there is anything else needed from me.

PR #27, opened 2019-11-15
Adding phan and a Docker hook for phan

bobstrecansky:
Adding phan, the PHP static analyzer, to OpenTelemetry-php. 

TODO: clean up phan errors :)


resources/Dockerfile.phan
```diff
@@ -0,0 +1,11 @@
+FROM php:7.2-buster
```
morrisonlevi: We should probably run these on PHP 7.1 if possible, since that's the oldest we support.

resources/Dockerfile.phan
```diff
@@ -0,0 +1,11 @@
...
+RUN apt-get -y update && apt-get -y install git zip && \
+curl -sS https://getcomposer.org/installer | php && \
+mv composer.phar /usr/local/bin/composer && \
+chmod +x /usr/local/bin/composer && \
+pecl install ast-1.0.4 && \
```
morrisonlevi: What is the ast extension for?

bobstrecansky: It was a requirement for phan.

resources/Dockerfile.phan
bobstrecansky: Sounds like a good plan.  I'll swap this out.

bobstrecansky:
I've updated this PR with a couple things:

Using PHP 7.1 (as per @morrisonlevi)
Adding some phan optimization for Docker (to improve performance)
Updating the test cases and a couple individual files so that both phan and PHPUnit pass successfully.

resources/Dockerfile.phan
bobstrecansky: From the `composer install` output:
`phan/phan suggests installing ext-ast (Needed for parsing ASTs (unless --use-fallback-parser is used). 1.0.1+ is needed.)`

composer.json
```diff
@@ -3,7 +3,8 @@
     "type": "library",
     "license": "Apache-2.0",
     "require": {
-        "php": "^7.1"
+        "php": "^7.1",
+        "phan/phan": "^2.4"
```
morrisonlevi: I think phan should be a require-dev dependency, not require. It's not needed for using our project, just developing for it.

bobstrecansky: I was planning on having phan run as part of the CI, so that's why I was including this here.  Happy to change it if you aren't of the same opinion?

PR #30, opened 2019-11-29
Performing Always and Never Sample Steps

bobstrecansky:
Creating a PR in order to include `Always Sample` and `Never Sample` samplers, as per the SDK spec:
This is a very rudimentry, blunt version of : https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#sampler

Performing this for now with the intention of updating it later to more closely follow the spec so work on the exporter can begin.

It makes sense to me to add the probability sampler after this is more explicitly defined:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#probability-sampler-algorithm


Dockerfile
```diff
@@ -1,8 +0,0 @@
-FROM php:7.2-buster
```
bobstrecansky: This Dockerfile was vestigial; we are coordinating them all in the `resources` folder now.

examples/AlwaysOffTraceExample.php
```diff
@@ -0,0 +1,28 @@
+<?php
+require __DIR__.'/../vendor/autoload.php';
+use OpenTelemetry\Tracing\Builder;
+use OpenTelemetry\Tracing\SpanContext;
+use OpenTelemetry\Tracing\Sampler\AlwaysOffSampler;
+$sampler = AlwaysOffSampler::shouldSample();
+if($sampler) {
+    $spanContext = SpanContext::generate(); // or extract from headers
+    $tracer = Builder::create()->setSpanContext($spanContext)->getTracer();
```
AustinSchoen: Was the builder removed in another PR?

bobstrecansky: Yup; this gets removed in the next PR:
https://github.com/open-telemetry/opentelemetry-php/pull/31

examples/AlwaysOnTraceExample.php
```diff
@@ -0,0 +1,28 @@
+<?php
+require __DIR__.'/../vendor/autoload.php';
+use OpenTelemetry\Tracing\Builder;
+use OpenTelemetry\Tracing\SpanContext;
+use OpenTelemetry\Tracing\Sampler\AlwaysOnSampler;
+$sampler = AlwaysOnSampler::shouldSample();
+if($sampler) {
+    $spanContext = SpanContext::generate(); // or extract from headers
+    $tracer = Builder::create()->setSpanContext($spanContext)->getTracer();
```
bobstrecansky: We should change this after the `Builder` goes away in the subsequent PR

PR #36, opened 2019-12-21
interfaces improvements

beniamin:
- moved Export interface to Exporter directory and renamed it from Exporter to ExporterInterface because I think that interfaces should be in the same directory with the implementations
- rename Transport interface into TransportInterface. I think that interface classes should have Interface prefix
- added SpanProcessorInterface as described in specifications
- improved tests autoloading in order to avoid importing require '/../../../../vendor/autoload.php' in every test file

### DIFF
--- a/src/Tracing/Sampler/AlwaysOffSampler.php
+++ b/src/Tracing/Sampler/AlwaysOffSampler.php
@@ -1,5 +1,6 @@
 <?php
 namespace OpenTelemetry\Tracing\Sampler;
+
 /**
  * This implementation of the SamplerInterface always returns false.
  * Example:
@@ -15,8 +16,13 @@ class AlwaysOffSampler implements SamplerInterface
      *
      * @return bool
      */
-    public function shouldSample()
+    public function shouldSample(): bool
     {
         return false;
     }
-} 
+
+    public function getDescription(): string
+    {
+        return self::class;
+    }
+}
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+/vendor
+/.git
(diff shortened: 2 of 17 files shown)


beniamin:
I signed it

beniamin:
I signed it

beniamin:
I sign it

src/Tracing/Sampler/AlwaysOffSampler.php
```diff
@@ -15,8 +16,13 @@ class AlwaysOffSampler implements SamplerInterface
      *
      * @return bool
      */
-    public function shouldSample()
+    public function shouldSample(): bool
```
bobstrecansky: Gah I completely missed this return and the AlwaysOn one too.  Thanks for the update!

beniamin: Tests are failing because of SpanContext::export method missing. It was remove in [6c9b99cfcdbf3d5ef4d4efc1595f6b7e7f5a1d80](https://github.com/open-telemetry/opentelemetry-php/commit/6c9b99cfcdbf3d5ef4d4efc1595f6b7e7f5a1d80#diff-b0672e6dc8ef53bdc61451969c5ecfe0L26). I don't know the reason it was removed. Should I change tests to pass without this method or should I add the method back in SpanContext?

bobstrecansky: I'd say add it back in! Thank you!

beniamin: Added exporter method back and fixed tests and phan issues

PR #38, opened 2020-01-02
TracerFactory implementation

beniamin:
- implement TracerFactory
- begin migration to the recommended library layout


src/Context/SpanContext.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace OpenTelemetry\Context;
+
+class SpanContext extends \OpenTelemetry\Tracing\SpanContext
```
bobstrecansky: I think you were renaming Tracing to Trace.  Did you mean to change it here too?

src/Trace/Span.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace OpenTelemetry\Trace;
+
+class Span extends \OpenTelemetry\Tracing\Span
```
bobstrecansky: I think you were renaming Tracing to Trace.  Did you mean to change it here too?

src/Trace/Tracer.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace OpenTelemetry\Trace;
+
+class Tracer extends \OpenTelemetry\Tracing\Tracer
```
bobstrecansky: I think you were renaming Tracing to Trace.  Did you mean to change it here too?

Makefile
```diff
@@ -0,0 +1,13 @@
+DC_RUN_PHP = docker-compose run php
```
bobstrecansky: This Makefile will probably make things easier.  We will probably want to update the travis.yaml to use some of these functionalities.

src/Trace/TracerFactory.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    protected $tracers;
+
+    /**
+     * @var SpanProcessorInterface[]
+     */
+    protected $spanProcessors;
+
+    private final function __construct()
+    {
```
bobstrecansky: What are you planning on adding in this constructor?  I would assume you are planning on creating an object containing some of the included vars?

tests/unit/Trace/TracerFactoryTest.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * @test
+     * @expectedException Error
+     */
+    public function shouldNotBeAbleToInstantiateDirectly()
+    {
+        new TracerFactory();
+    }
+}
```
bobstrecansky: What other tests are you considering including for the TracerFactory?  A `getTracer` and `getInstance` test might be prudent?

src/Context/SpanContext.php
beniamin: I followed instructions from here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md and I want to rename Tracing to Trace.
For now, I just extended the previous implementation to make sure that it's ok for you to rename. After we decide this is the right path, I will move the implementation here.

For this case, I moved SpanContext in OpenTelemtry\Context.

src/Trace/Span.php
beniamin: I followed instructions from here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md and I want to rename Tracing to Trace.
For now, I just extended the previous implementation to make sure that it's ok for you to rename. After we decide this is the right path, I will move the implementation here.

src/Trace/Tracer.php
beniamin: I followed instructions from here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md and I want to rename Tracing to Trace.
For now, I just extended the previous implementation to make sure that it's ok for you to rename. After we decide this is the right path, I will move the implementation here.

src/Trace/TracerFactory.php
beniamin: I didn't plan to do anything, just wanted to implement TracerFactory as a singleton. I took the suggestion from specification: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#obtaining-a-tracer

Now I think that I can use the constructor to inject SpanProcessors and other configuration in the TracerFactory object and not in getInstance() method.

tests/unit/Trace/TracerFactoryTest.php
beniamin: Yes, of course. This is work in progress.

src/Context/SpanContext.php
bobstrecansky: Yeah; following the spec here is perfect.  I don't believe this existed when we implemented originally (or was inadvertent).  This change sounds good.

src/Trace/TracerFactory.php
bobstrecansky: Sounds like a good plan.

bobstrecansky:
Thank you for the travisci update as well!

src/Trace/TracerFactory.php
```diff
@@ -0,0 +1,73 @@
...
+        if (self::$instance instanceof TracerFactory) {
+            return self::$instance;
+        }
+
+        $instance = new TracerFactory($spanProcessors);
+
+        return self::$instance = $instance;
+    }
+
+    public function getTracer(string $name, string $version = ""): Tracer {
```
bobstrecansky: Should this `{` be on the next line?

examples/AlwaysOffTraceExample.php
```diff
@@ -1,14 +1,15 @@
...
+use OpenTelemetry\Trace\Sampler\AlwaysOffSampler;
+use OpenTelemetry\Trace\Tracer;
+use OpenTelemetry\Trace\TracerFactory;
 
 $sampler = AlwaysOffSampler::shouldSample();
 if ($sampler) {
-    $spanContext = SpanContext::generate(); // or extract from headers
-    $tracer = new Tracer($spanContext);
+    $tracer = TracerFactory::getInstance()
+        ->getTracer('io.opentelemetry.contrib.php');
```
bobstrecansky: Did this naming scheme come from somewhere in particular?  I didn't see it in the spec, but I could just be missing it.

beniamin: Yes, I found something similar in the specs: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#obtaining-a-tracer

src/Trace/TracerFactory.php
beniamin: Yes, it should. I think we should add somethinh like CodeSniffer.

examples/AlwaysOffTraceExample.php
bobstrecansky: Good enough for me!

PR #47, opened 2020-01-22
Add exporter Status class

nenad:
Creates a `Exporter\Status` class that holds the constants for possible responses from the exporter and replaces usages of the constants to this class.

This is similarly done for the `Trace` package.

### DIFF
--- a/src/Exporter/BasisExporter.php
+++ b/src/Exporter/BasisExporter.php
@@ -22,6 +22,6 @@ class BasisExporter implements ExporterInterface
 
     public function export(iterable $spans): int
     {
-        return ExporterInterface::SUCCESS;
+        return Status::SUCCESS;
     }
 }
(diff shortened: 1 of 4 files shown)


bobstrecansky:
Could we possibly use a class name a bit more descriptive? `Status` feels a bit too generic to me. Maybe we could use something like `ExporterStatus` or something similar? Thoughts?

nenad:
> Could we possibly use a class name a bit more descriptive? `Status` feels a bit too generic to me. Maybe we could use something like `ExporterStatus` or something similar? Thoughts?

I intentionally left it only as `Status` to prevent package "stutter", since it's already in the `Exporter` package. If you think this is a nuisance, I'll rename it.

bobstrecansky:
Avoided stutter is always a good thing 🤠

nenad:
I suppose we can live with them being in the interface. Closing the PR.

PR #50, opened 2020-01-24
Add php-cs-fixer

nenad:
Adds the [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and fixes all PHP files.


beniamin:
Good idea.

PR #51, opened 2020-01-24
Remove the composer.lock file

nenad:
Since this is a package meant for public usage, we can live without the lock file.

https://github.com/composer/composer/issues/2461


nenad:
Good spot, it slipped my mind. Thanks!

beniamin:
> Since this is a package meant for public usage, we can live without the lock file.
> 
> [composer/composer#2461](https://github.com/composer/composer/issues/2461)

I don't think we should. Composer.lock help all maintainers to work with same libraries version.

nenad:
Fair point, I'll close the PR. 

Related to the dependency versions, should we test with lowest and highest version of dependencies in the CI to handle more cases and avoid a scenario like the one you described?

beniamin:
We already do this. We are running travis for php 7.1, 7.2, 7.3.

Do you think on something else?

nenad:
Yes, I was referring to using different flags for composer deps download:

```bash
--prefer-stable: Prefer stable versions of dependencies.
--prefer-lowest: Prefer lowest versions of dependencies. Useful for testing minimal versions of requirements, generally used with --prefer-stable.
```

This would give us more confidence that the tool is very compatible, however, the drawback is that we have multiple test suites to run.

beniamin:
On the other hand, yesterday I ran in a different issue. After merging the PR with php-cs-fixer I was unable to install the library, because, I suppose, the php-cs-fixer was installed in an environment with php 7.2 and I was running php 7.1.

Right now I am thinking that may be better to completly remove the composer.lock and leave the issue that I was talking in the previous comment on the developer side.

nenad:
In my experience, I have never worked with a `composer.lock` file on a library. If errors like the one you described pop up while development with the lock file, we could remove it and as suggested, leave the developer to deal with the dependencies. This would also increase the compatibility and would detect errors with various versions of libraries we use in this package earlier. Feel free to re-open the PR and merge it if you feel we'll benefit more without the lock :)

bobstrecansky:
Reopened for discussion. I'm still happy to merge this if that becomes the consensus

beniamin:
I think that removing composer.lock it is a better ideea.

nenad:
Apologies for the force-push, I had to resolve the conflicts. I too am in favor of removing the `composer.lock` file. :+1:

PR #53, opened 2020-01-26
Clean up Trace\Status

nenad:
Code cleanup on the `Trace\Status` class.

### DIFF
--- a/src/Trace/Status.php
+++ b/src/Trace/Status.php
@@ -50,11 +50,9 @@ class Status
     public function __construct(int $code, string $description = null)
     {
         $this->code = $code;
+        $this->description = $description;
         if (!$description && array_key_exists($code, self::DESCRIPTION)) {
-            $description = self::DESCRIPTION[$code];
-        }
-        if (null !== $description) {
-            $this->description = $description;
+            $this->description = self::DESCRIPTION[$code];
         }
     }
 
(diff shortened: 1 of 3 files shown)


bobstrecansky:
Agreed with @morrisonlevi - we should work towards the (overwhelmingly) common use case of $ok.

nenad:
I've made the constructor `private` so that we can enforce reusing the `Status`es from the `$map` object pool. Additionally, I've added a `Status::ok()` function that returns OK status, since it seems to be a common case based on your input. About the `isStatusOK` method, I would work on it in another PR, didn't want to make this one too big. What do you think?

bobstrecansky:
I'm going to defer to @morrisonlevi on this one.  I know his refactor with interfaces touches this, and he would probably have some additional insight here!

bobstrecansky:
Looks like this broke CI: https://travis-ci.org/open-telemetry/opentelemetry-php/jobs/645774130?utm_medium=notification&utm_source=github_status

nenad:
> Long term, we need to make that not return a bash error code (in this particular instance it returned an 8 exit code).

It was intentional to return error code so that we can make sure the style is not broken. Although, I'm not sure why the PR didn't report it, maybe it was not rebased? We could also turn on the repository setting to enforce the PR to be always up-to-date with master.

PR #54, opened 2020-01-27
More style

nenad:
Adds two more nice style checks. Additionally, sets up a style check on Travis. Also, it cleans up any container spun up when a Makefile command was executed, to prevent polluting the Docker daemon.


Makefile
```diff
@@ -1,4 +1,4 @@
-DC_RUN_PHP = docker-compose run php
+DC_RUN_PHP = docker-compose run --rm php
```
beniamin: good job

PR #55, opened 2020-01-27
Disable XDebug on CI

nenad:
Only enables XDebug when running PHPUnit. This should speed up PHP tasks like Phan and in some cases Composer if Travis doesn't optimize for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,5 @@ script:
     - vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
     - vendor/bin/phan
 
-after_script:
-    - |
-        if [[ "$TRAVIS_PHP_VERSION" != 'hhvm' ]]; then
-          wget https://scrutinizer-ci.com/ocular.phar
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,19 +14,20 @@ cache:
...
+    - vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
 
 after_script:
-    - |
-        if [[ "$TRAVIS_PHP_VERSION" != 'hhvm' ]]; then
-          wget https://scrutinizer-ci.com/ocular.phar
-          php ocular.phar code-coverage:upload --format=php-clover coverage.clover
-        fi
+    - xdebug-disable
+    - wget https://scrutinizer-ci.com/ocular.phar
```
bobstrecansky: I think we have gotten rid of scrutinizer?

bobstrecansky:
We will also need to change your conflicting `.travis.yml` changes from your other PR.  Other than those two tiny things; this looks good!

.travis.yml
nenad: Just did :+1:

PR #57, opened 2020-02-03
Added Propagators API

Wizofgoz:
Implements Propagators API up to specifications: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-propagators.md

I couldn't find any mention of what the binary format should be so I used what the [Go library](https://github.com/open-telemetry/opentelemetry-go/blob/master/api/propagation/binary_propagator.go) had as a starting point.


morrisonlevi:
I don't like how deep the namespaces are. We really only need (and should probably only have) 2 levels: one for the OpenTelemetry umbrella, and maybe just one more to separate things like trace vs metrics. That's it.

I'm not sure how pervasive this is in the rest of our codebase, and changing it right now in this PR is not required; I just wanted to start a discussion on it and here seemed as good of place as any.

src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Propagation\HTTPTextFormat;
+
+use OpenTelemetry\Context\SpanContext;
+use OpenTelemetry\Propagation\Getter\GetterInterface;
+use OpenTelemetry\Propagation\Setter\SetterInterface;
+
+interface HTTPTextFormatInterface
```
morrisonlevi: I don't like this situation:
- We have a triple segmented namespace when I think we should have exactly two segments 1) for the OTEL umbrella and 2) for Trace vs Metrics and so on.
- The name `HTTPTextFormatInterface` is really long, and the `Interface` suffix is unnecessary. Almost entirely the point of OTEL is to have common interfaces for tools to target, so the clean name should be the interface name, not the implementation.

So taking these `use` statements:

```php
use OpenTelemetry\Propagation\Getter\GetterInterface;
use OpenTelemetry\Propagation\Setter\SetterInterface;
```

and, without having read the spec on the subject to propose better names than just `Getter` and `Setter`, I think it should be more like this:

```php
use OpenTelemetry\Propagation\Getter;
use OpenTelemetry\Propagation\Setter;
```

What do the rest of you think?

src/Propagation/BinaryFormat.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Propagation;
+
+use OpenTelemetry\Context\SpanContext;
+
+final class BinaryFormat
```
morrisonlevi: Does the spec expect us to provide an implementation of the binary format? If so it's odd that there isn't a description or spec for the format. If it doesn't expect us to provide an impl, then this should probably just be an interface, I think?

bobstrecansky:
One other minor nit I saw with php-cs-fixer too:

```docker-compose run --rm php php ./vendor/bin/php-cs-fixer fix

Fatal error: Declaration of Symfony\Contracts\EventDispatcher\EventDispatcherInterface::dispatch(Symfony\Contracts\EventDispatcher\object $event, ?string $eventName = NULL): Symfony\Contracts\EventDispatcher\object must be compatible with Psr\EventDispatcher\EventDispatcherInterface::dispatch(Psr\EventDispatcher\object $event) in /usr/src/myapp/vendor/symfony/event-dispatcher-contracts/EventDispatcherInterface.php on line 19
make: *** [Makefile:16: style] Error 255```

File: src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php (hunk 1 above)
Comment (bobstrecansky) [reply src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php]: Your implementation idea seems a bit cleaner to me too @morrisonlevi .  A lot more readable.

File: src/Propagation/BinaryFormat.php (hunk 2 above)
Comment (bobstrecansky) [reply src/Propagation/BinaryFormat.php]: The binary format of the spec is pretty ambiguous now:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-propagators.md#binary-format

I'd expect that might change in the future?

Comment (Wizofgoz) [reply src/Propagation/BinaryFormat.php]: I agree, it is a bit odd that it's not in the spec. I based what is done here on what is used for the Go library as I said in my initial message but if you guys want to be a bit more cautious, I could swap it with a simple interface for now.

File: src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php (hunk 1 above)
Comment (Wizofgoz) [reply src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php]: I agree, the `Interface` suffix is unneeded and your suggestion is much easier to read. I was just following how other Interfaces were named in the project.

As for the extra namespaces, I expected there to be quite a few implementations for each of them for all the different carriers that could be used and was concerned the folder might get too crowded but I could be persuaded otherwise.

**Comment by Wizofgoz** [comment]
@bobstrecansky I noticed that as well. It could be a PHP version issue because when I changed the version in the Dockerfile to 7.2 and rebuilt, it ran correctly. I could be wrong though, never used the tool before.

**Comment by bobstrecansky** [comment]
ahh understand @Wizofgoz - We are currently attempting to build for 7.1 and later.  We should try and follow that pattern.  There was just another PR that had the same issue and it broke CI.  More context here: 

https://github.com/open-telemetry/opentelemetry-php/pull/53#issuecomment-581733323

File: src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php (hunk 1 above)
Comment (nenad) [reply src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php]: True, it's more readable without the `Interface` suffix. About the carriers - do we really need them if we have the `Setter` interface? One could implement a setter that would modify the carrier to inject the variables. [This](https://github.com/nenad/opencensus-php/blob/master/src/Trace/Propagator/HeaderSetter.php#L11) is how I've dealt with this situation in the OpenCensus implementation.

Comment (Wizofgoz) [reply src/Propagation/HTTPTextFormat/HTTPTextFormatInterface.php]: @nenad I believe that's what the spec is meant to do. As you can see in the interface for a `Setter`
```php
public function put($carrier, string $key, string $value) : void;
```
The carrier (Guzzle for example) is passed into `put` and is modified by applying the key/value. So the carriers wouldn't need to be implemented by this library, just the `Getter`s and `Setter`s for them.

After looking over the code again, I would expect there shouldn't be many implementations of `HTTPTextFormat` since the `Getter`s and `Setter`s will be so flexible so I wouldn't be opposed to moving that back to the `\OpenTelemetry\Propogation` namespace.

tests/integration/BinaryFormatTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->assertNotEmpty($serialized);
+        $unserializedSpan = $formatter->fromBytes($serialized);
+        $this->assertEquals($span->getTraceId(), $unserializedSpan->getTraceId());
+        $this->assertEquals($span->getSpanId(), $unserializedSpan->getSpanId());
+        $this->assertEquals($span->isSampled(), $unserializedSpan->isSampled());
+        $this->assertNotEquals($span->isRemote(), $unserializedSpan->isRemote());
+    }
+
+    /**
+     * @test
```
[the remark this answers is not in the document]
Wizofgoz: Those have now been removed

Wizofgoz:
@morrisonlevi I removed `BinaryFormat` since it was just removed from the spec.

I also moved all interfaces to the `\OpenTelemetry\Propagation` namespace and removed noop implementations.

morrisonlevi:
I'm a bit shaky on what a `$carrier` is. What's the reason it is `mixed`?

Wizofgoz:
@morrisonlevi as I interpreted the spec, an example of a carrier would be a [PSR7 Request/Response](https://www.php-fig.org/psr/psr-7/) where the propagation attributes can be set with a `Psr7RequestSetter` and extracted with a `Psr7RequestGetter` or something similar

src/Propagation/Setter.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
```
morrisonlevi: I don't think `declare(strict_types=1)` do anything in files that only contain an interface declaration. Will our linter complain?

Wizofgoz: I haven't seen any warnings and all previous interfaces seem to have `declare(strict_types=1)` as well.

morrisonlevi: Can you try removing it and running the linter? I don't think it does anything lol

Wizofgoz: I removed it and ran `make style` and it re-applied `declare(strict_types=1)` to the interfaces

nenad:
> @morrisonlevi as I interpreted the spec, an example of a carrier would be a [PSR7 Request/Response](https://www.php-fig.org/psr/psr-7/) where the propagation attributes can be set with a `Psr7RequestSetter` and extracted with a `Psr7RequestGetter` or something similar.
> 
> Because of this, the interface needs to be pretty permissive to accept all carriers but implementations can be more strict.

I'm with you on the `Psr7RequestSetter` bit, and with this I believe we can make the `$carrier` variable not required. If the `Psr7RequestSetter` takes the PSR7 Request as a parameter either in the constructor or a static function, we can encapsulate the injection logic in the `Setter` without the need for carriers. Example:
```php
$propagator->inject($context, new Psr7RequestSetter($request));
```
or

```php
$propagator->inject($context, Setter::fromPSR7($request));
```

The `inject` method of the propagator would just need to iterate through the headers  and use the `Setter::put()` to set them. The same principle could be applied to the `Getter` interface.

With this approach, we keep type safety and don't obstruct new implementations.

Wizofgoz:
@nenad good point, that would be cleaner. Those changes should be done now.

morrisonlevi:
My reading of the spec is that the carrier must be a parameter because `Getter` and `Setter` MUST be stateless.

nenad:
With the current interface we are not keeping state in the `Getter`s and `Setter`s. They are just processors used for extracting/injecting data from/to various carriers. 

If you agree on it, I propose to merge the PR and iterate more quickly on implementations and future modifications on the interfaces if necessary.

bobstrecansky:
@morrisonlevi - does this interfere with the interfaces work you're doing?  I don't want either of y'all to be blocked.

morrisonlevi:
It doesn't interfere. I am on mobile and looking at code is kind of
difficult right now, but IIRC we need to figure out why carriers are a
required parameter in the spec.

bobstrecansky:
@nenad - is that something you can look into?

nenad:
I've looked into this scenario when I was working on OpenCensus PHP implementation, and I've chosen to omit the carrier in the `inject` and `extract` methods as it seemed like a good idea.

However, after re-reading the specs, I might have a different opinion.

The specs say this about the `carrier` parameter in `inject` and `extract` which are methods defined on a `HTTPTextFormat`:

> the carrier that holds propagation fields. For example, an outgoing message or http request.

As a parameter to `inject` and `extract`, we need to pass `Setter` or `Getter` correspondingly. 

Now, let's take at the `Setter` and `Getter` interfaces and what do they do:

> Setter allows a HTTPTextFormat to set propagated fields into a carrier.

> Getter allows a HttpTextFormat to read propagated fields from a carrier.

As @morrisonlevi pointed out, the specs explicitly say that the `Setter`s and `Getter`s must be stateless so that we can initialize them before runtime to minimize memory/CPU costs. After some re-evaluation on the solution I proposed, I understand that the `$carrier` parameter is required in both the `HTTPTextFormat` and `Setter/Getter` objects to fulfill the requirements in the specs. 

However, this leaves us with having the carrier parameter as `mixed`, which I personally prefer to avoid, especially in interfaces. As the scope of the `HTTPTextFormat` is very narrow, it can only be a HTTP Request, could we use PSR-7 definition of a Request as the type of the carrier instead? I am aware that this might limit the amount of frameworks that OTel might be compatible with and I'm wondering if this is something we want to do.

bobstrecansky:
I agree this is a bit of a conflict.  I've asked for more clarification in https://gitter.im/open-telemetry/opentelemetry-specification

bobstrecansky:
Y'all wanna move forward with this?

nenad:
If @Wizofgoz is willing to revert the changes (apologies for misguiding you), I'm in favor of merging it.

Wizofgoz:
No problem, I should be able to do it tonight

Wizofgoz:
This now includes the update to the spec that the method on the `Setter` interface was renamed to `set` instead of `put` but doesn't include Composite or Global Propagators. IMO, that can probably be handled by another PR.

bobstrecansky:
I'm good with this if @morrisonlevi is!  I'll give him until Friday and then merge it.

morrisonlevi:
Lgtm

PR #58, opened 2020-02-04
Fixing PHP for CI using php-cs-fixer

bobstrecansky:
Attempting to fix CI for the time being using php-cs-fixer


bobstrecansky:
Thanks for the review Nenad!  I'll also need to get a reviewer to review this.  @AustinSchoen should be able to do that today.

PR #62, opened 2020-02-16
Remove PHP 7.1, add PHP 7.4

bobstrecansky:
PHP 7.1 has been EoL'ed, and doesn't have security support anymore.  I've removed it from CI and moved our default Dockerfile to use PHP7.2.  

I've also added in testing for PHP7.4, as php 7.4.2 is the current stable PHP version.

### DIFF
--- a/src/Trace/TracerFactory.php
+++ b/src/Trace/TracerFactory.php
@@ -65,7 +65,8 @@ class TracerFactory
 
     public function getTracer(string $name, string $version = ''): Tracer
     {
-        if ($this->tracers != null && $this->tracers[$name] instanceof Tracer) {
+
+        if (isset($this->tracers[$name]) && $this->tracers[$name] instanceof Tracer) {
             return $this->tracers[$name];
         }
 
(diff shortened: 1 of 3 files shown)


src/Trace/TracerFactory.php
```diff
@@ -65,7 +65,7 @@ public static function getInstance(array $spanProcessors = []): self
 
     public function getTracer(string $name, string $version = ''): Tracer
     {
-        if ($this->tracers[$name] instanceof Tracer) {
+        if ($this->tracers != null && $this->tracers[$name] instanceof Tracer) {
```
bobstrecansky: @beniamin - I had to change this implementation in the `TracerFactory` because PHP 7.4 was complaining about `Trying to access array offset on value of type null`

You okay with this implementation or do you have other ideas of how you'd like to handle this?

morrisonlevi:
Something like this is slightly more robust:


isset($this->tracers[$name]) && $this->tracers[$name] instanceof Tracer

As the isset covers both tracers and tracers[$name] being null.

bobstrecansky:
@AustinSchoen @morrisonlevi - bump please!

PR #64, opened 2020-03-12
Add API Interfaces

morrisonlevi:
The goal is to add the API as pure interfaces so that implementations have a lot of freedom on how they implement these things. A few things specifically worth talking about:

- The Span extends SpanStatus and SpanKind. The latter two APIs are very light on data, so storing them in a separate object seems too heavy for the common case; this way they can be queried and used directly on the Span and it seems prudent to guarantee this capability by the API.
- The Attributes, Events, and Links have the same pattern; they are a data-structure that implements PHP's IteratorAggregate API so they can return an Iterator, and the specific iterator returns objects of a given type. If PHP had generics we probably wouldn't do this and would just return `Iterator<T>` where T is one of `Attribute`, `Event`, or `Link`. My past experience says we need to specify these types or people will end up screwing it up and then implementations will be incompatible.
    - There's also a pattern that can't be specified in interface requirements, so I left it in the code comments in a few places. The iterator objects should be permitted to call `.rewind()` as many extra times as desired _until_ `.next()` has been called. This allows a caller using a function designed for abstract  `Iterator`s to know if there is at least one value (i.e. if it's not empty) by rewinding the iterator and checking if it is `.valid()` without invalidating the iterator.
- ~I chose not to delete or modify any existing code at this time. We need to discuss what to do with our existing code as it seems useful. Maybe it belongs in the SDK?~ It's moved to the SDK, found in `sdk/`.
- I put the code into `api/Trace` as recommended by the OpenTelemetry project layout, which deviates from traditional PHP package layouts. I'm not sure which layout is better to follow.


morrisonlevi:
Current CI failure: `SpanKind` and `SpanStatus` both have an `INTERNAL` constant, so `Span` fails to inherit them both.

beniamin:
I was looking at `symfony/contracts` library and at `laravel/contracts` library and, for me, it seems a good way to go if we want two distinct libraries, one for SDK and one for API.

For me, using just `Opentelemtry\` namespace for api and `Opentelemetry\Sdk` for sdk it's a little bit confusing. It takes me to the ideea that sdk is a part of api, which I think is wrong.

I suggest moving the API in its own namespace like Opentelemtry\Api or Opentelemetry\Contract, with a preference for the second form.

api/Trace/Attributes.php
```diff
@@ -0,0 +1,19 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+interface Attributes extends \IteratorAggregate, \Countable
+{
+    /**
+     * Setting event should not invalidate nor change any existing iterators.
+     * @param string $name
+     * @param $value
```
beniamin: type hinting should be added here, and should be the same with Attribute::getValue()

api/Trace/Events.php
```diff
@@ -0,0 +1,20 @@
...
+interface Events extends \IteratorAggregate, \Countable
+{
+    /**
+     * Adding an event should not invalidate nor change any existing iterators.
+     * @param string $name
+     * @param Attributes|null $attributes
+     * @param string|null $timestamp
+     * @return Events Must return $this to allow setting multiple attributes at once in a chain.
+     */
+    function addEvent(string $name, ?Attributes $attributes = null, ?string $timestamp = null): Events;
```
beniamin: I think that we should not accept null for Attributes and timestamp. I think we should use an empty array and an empty string as default values.
Having both empty array/string and null will make the implementation harder because both cases should be treated.

api/Trace/Links.php
```diff
@@ -0,0 +1,19 @@
...
+    function count(): int;
+    function getIterator(): LinksIterator;
+
+    /**
+     * Adding a link should not invalidate nor change any existing iterators.
+     * @param SpanContext $context
+     * @param Attributes|null $attributes
+     * @return Links Return $this to allow setting multiple links at once in a chain.
+     */
+    function addLink(SpanContext $context, ?Attributes $attributes = null): Links;
```
beniamin: same thing like in the Event case with null as default value

api/Trace/Span.php
```diff
@@ -0,0 +1,53 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+interface Span extends SpanStatus, SpanKind
+{
+    function getSpanName(): string;
+    function getContext(): SpanContext;
+    function getParent(): ?SpanContext;
+    function getStartTimestamp(): string;
```
beniamin: Do you think we should use DateTime objects instead of strings?

api/Trace/Span.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    function addEvent(string $name, ?Attributes $attributes = null, ?string $timestamp = null): Span;
+
+    /**
+     * @param SpanContext $context
+     * @param Attributes|null $attributes
+     * @return Span Must return $this to allow setting multiple links at once in a chain.
+     */
+    function addLink(SpanContext $context, ?Attributes $attributes = null): Span;
+
```
beniamin: I think you missed the SetStatus API. https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#set-status

api/Trace/SpanContext.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace OpenTelemetry\Trace;
+
+interface SpanContext
+{
+    function getTraceId(): string;
+    function getSpanId(): string;
+    function getTraceFlags(): int;
+    function getTracestate(): array;
+    function IsValidContext(): bool;
```
beniamin: since we are in SpanContext, do we need Context suffix for IsValid and IsRemote?

api/Trace/SpanKind.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+interface SpanKind
+{
+    const KIND_INTERNAL = 0;
```
beniamin: Where did the values come from? It couldn't find anything in the specs.

api/Trace/Span.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Calling this method is highly discouraged; the name should be set on creation and left alone.
+     * @param string $name
+     * @return Span Must return $this
+     */
+    function updateName(string $name): Span;
+
+    // TODO: addLazyEvent
+    // TODO: end(), though why is this allowed on the span? I thought Tracers were responsible for this?
```
beniamin: Shouldn't ::end() method allow the possibility to set endTimeStamp? If this method is not available, how should one set a Span as ended?

api/Trace/SpanOptions.php
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+/**
+ * The SpanOptions implementation is intended to be tightly coupled with the Span and Tracer implementations.
+ * Hopefully the Span object can implement SpanOptions so that the toSpan() call is almost a no-op, but prevents
+ * the rest of the SpanOptions API from being used after toSpan() is called.
+ */
+interface SpanOptions
```
beniamin: If I understand it right, this is some kind of Span Builder?
I assumed that if one would want to create a span, then he should pass all parameters in the Span constructor.

api/Trace/Events.php
morrisonlevi: We can't pass an array for Attributes; it's a type mismatch. I think supporting NULL is important because I would assume that most of the time you aren't adding any attributes to the event.

Similar reasons for timestamp, although we should consider validating the passed timestamp.

api/Trace/Attributes.php
morrisonlevi: Good point.

api/Trace/Span.php
morrisonlevi: No. At least in my opinion this is a case where the cost of an object is too high:

```php
<?php
function cost_date() {
    $before = memory_get_usage();
    $date = new DateTime('now');
    $after = memory_get_usage();
    return [$date, $after - $before];
}

function cost_string() {
    $before = memory_get_usage();
    $date = microtime();
    $after = memory_get_usage();
    return [$date, $after - $before];
}

echo cost_date()[1], "\n";
echo cost_string()[1], "\n";
```
> 960
48

The object is just so heavy, and keep in mind you need a start and stop for every span.

morrisonlevi: Although, "string" is pretty ambiguous, so I don't know. I'll think more on this.

api/Trace/SpanContext.php
morrisonlevi: Yes! I hope that SDKs will implement SpanContext on the Span not use a separate object for the context unless specifically asked for. Again, for memory reasons. Originally I made this a requirement, but relaxed it. What do you think?

api/Trace/SpanKind.php
morrisonlevi: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#spankind

api/Trace/Span.php
morrisonlevi: I need to reread the spec here, but there is a line about Tracer's being the one to manage spans, which is why you can't do `new Span()` or something.

api/Trace/SpanOptions.php
morrisonlevi: Yes, it's some sort of builder. However, you can't use a span's constructor; it's forbidden in the spec (citation needed).

api/Trace/Attributes.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+interface Attributes extends \IteratorAggregate, \Countable
+{
+    /**
+     * Setting event should not invalidate nor change any existing iterators.
```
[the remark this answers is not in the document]
morrisonlevi: I would be 👍  except the spec and anchors have changed over the time I've been contributing, so I don't know if they are stable enough yet.

api/Trace/Attribute.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+interface Attribute
+{
+    function getKey(): string;
+
+    /**
+     * @return bool|int|float|string|iterable Note: the iterable MUST be homogeneous, i.e. it MUST NOT contain values
```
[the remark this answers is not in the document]
morrisonlevi: I'll have to learn more about this.

api/Trace/Attribute.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
```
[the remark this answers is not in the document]
morrisonlevi: I would be 👍  except the spec and anchors have changed over the time I've been contributing, so I don't know if they are stable enough yet.

morrisonlevi:
I pulled in latest master and squashed. Current state is that `make test` should work, but there are still gaps in the SDK:

- Links
- Some Tracer methods, and SpanOptions
- Noop implementations

api/Trace/Span.php
morrisonlevi: Fixed in a recent commit 👍

api/Trace/Span.php
morrisonlevi: I've added `end()`, which optionally takes a timestamp (still need to rework timestamps in general, though).

morrisonlevi:
Phew, I think this PR is finally "done". There are still things the project needs done; I even added some todos in code. In particular, I think we should leave changing the timestamps to a future PR, as this one has already changed a ton of stuff and we need to get it merged in, and that PR may change requirements to PHP 7.3.

So let me know what issues you see!

api/Trace/Attribute.php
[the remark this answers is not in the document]
bobstrecansky: @beniamin - I started looking more into this.  Can you give me an example of what you're expecting here?  I'll go through and update the codebase accordingly, just want to make sure that I'm not missing something.  Phan currently isn't complaining about this?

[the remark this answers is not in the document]
beniamin: I was thinking to use something like ‘list<int>|list<string>...’ as a more strict type for iterables.

PR #70, opened 2020-03-30
Jaeger exporter

beniamin:
Added new exporter for Jaeger using Zipkin like HTTP protocol.

- [x] implement exporter
- [x] add new HTTP request client using curl,  becuase file_get_contents return warning and connot be catched - done in #72 
- [x] refactor Zipkin and Jaeger to use HTTP request client - done in #72 
- [x] unit test for all new code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ install:
 update:
 	$(DC_RUN_PHP) composer update
 test:
-	$(DC_RUN_PHP) php ./vendor/bin/phpunit --colors=always
+	$(DC_RUN_PHP) php ./vendor/bin/phpunit --colors=always --coverage-text
 phan:
 	$(DC_RUN_PHP) env PHAN_DISABLE_XDEBUG_WARN=1 php ./vendor/bin/phan
 examples: FORCE
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ phan:
 	$(DC_RUN_PHP) env PHAN_DISABLE_XDEBUG_WARN=1 php ./vendor/bin/phan
 examples: FORCE
 	$(DC_RUN_PHP) php ./examples/AlwaysOnTraceExample.php
+	$(DC_RUN_PHP) php ./examples/AlwaysOffTraceExample.php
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -10,6 +10,8 @@ phan:
 	$(DC_RUN_PHP) env PHAN_DISABLE_XDEBUG_WARN=1 php ./vendor/bin/phan
 examples: FORCE
 	$(DC_RUN_PHP) php ./examples/AlwaysOnTraceExample.php
+	$(DC_RUN_PHP) php ./examples/AlwaysOffTraceExample.php
+	$(DC_RUN_PHP) php ./examples/JaegerExporterExample.php
```
bobstrecansky, line 14: This is great for now - eventually we may want to turn these into integration tests.

bobstrecansky: Also, perhaps we want to perform a `docker-compose up -d` in `make examples` before we execute the examples?  That may take away some of the confusion of running these examples.

beniamin: yes, totally make sense to add `docker-compose up -d` in make examples. I will do it in this PR.

bobstrecansky:
@beniamin - I think I fixed bullet points 2 and 3 (at least partially) in this PR: 
https://github.com/open-telemetry/opentelemetry-php/pull/72

Makefile
bobstrecansky: We should also probably update the examples section in the README with this change:

https://github.com/open-telemetry/opentelemetry-php/#examples

beniamin:
> @beniamin - I think I fixed bullet points 2 and 3 (at least partially) in this PR, as long as we decide this is a decent pattern:
> #72

Then this PR should be considered completed and ready for review.

bobstrecansky:
@beniamin - This looks good to me.  You mentioned you wanted to add tests for these, are you planning on doing that in a subsequent PR/

beniamin:
I fully tested SpanConverter, but the exporter methods of both exporter are impossible to test because of file_get_contents.
I can say that I tested everything I could.

PR #71, opened 2020-04-05
Adding Explicit Reference for Attribute Iterables

bobstrecansky:
@beniamin mentioned that we should add explicit references for attribute iterables.  I performed this action for the api and sdk trace attribute `getValue` functions; we should discuss whether or not there are other places where we would like to do this.

### DIFF
--- a/api/Trace/Attribute.php
+++ b/api/Trace/Attribute.php
@@ -9,8 +9,8 @@ interface Attribute
     public function getKey(): string;
 
     /**
-     * 
-     * @return bool|int|float|string|list<int>|list<float> 
+     *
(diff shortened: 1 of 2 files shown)


bobstrecansky:
Yup, it should.  I misread the spec.  Updating now.

morrisonlevi:
As far as I can tell, `list<bool>` should also be supported, right?

This is one of those times where I wonder what the impact would be to use a explicitly typed structure or API for it. I'll poke around and see if I have any ideas that aren't bloody annoying

bobstrecansky:
No problem.  I thought about adding a list of booleans.  It felt weird but I guess I should follow the spec.  Will update and let you poke around :)

PR #72, opened 2020-04-05
Using Guzzle for Zipkin HTTP Requests

bobstrecansky:
In our gitter channel, we discussed using Guzzle as our HTTP request client rather than file_get_contents.  

Using [Guzzle](http://docs.guzzlephp.org/en/stable/) seemed like a sane choice here.  This is an initial thought about how to implement this for the zipkinExporter.  I also have changed this for @beniamin 's open PR in [this branch](https://github.com/open-telemetry/opentelemetry-php/compare/jaegertest?expand=1) should we choose to go this route.

I'll be happy to write tests for this, but I wanted to make sure we were all on board with this implementation before I wrote them.

### DIFF
--- a/sdk/Trace/ZipkinExporter.php
+++ b/sdk/Trace/ZipkinExporter.php
@@ -54,9 +54,10 @@ class ZipkinExporter implements Exporter
 
         try {
             $json = json_encode($convertedSpans);
+            $url = $this->getEndpointUrl();
             $client = new \GuzzleHttp\Client();
             $headers = ['content-type' => 'application/json'];
-            $request = new Request('POST', $this->endpointUrl, $headers, $json);
+            $request = new Request('POST', $url, $headers, $json);
             $response = $client->send($request, ['timeout' => 2]);
         } catch (Exception $e) {
             return Exporter::FAILED_RETRYABLE;
--- a/sdk/Trace/ZipkinExporter.php
+++ b/sdk/Trace/ZipkinExporter.php
@@ -58,7 +58,7 @@ class ZipkinExporter implements Exporter
             $client = new \GuzzleHttp\Client();
             $headers = ['content-type' => 'application/json'];
             $request = new Request('POST', $url, $headers, $json);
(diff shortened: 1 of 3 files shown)


sdk/Trace/ZipkinExporter.php
```diff
@@ -53,15 +54,11 @@ public function export(iterable $spans) : int
...
-                    'content' => $json,
-                ],
-            ];
-            $context = stream_context_create($contextOptions);
-            file_get_contents($this->getEndpointUrl(), false, $context);
+            $url = $this->getEndpointUrl();
+            $client = new \GuzzleHttp\Client();
+            $headers = ['content-type' => 'application/json'];
+            $request = new Request('POST', $url, $headers, $json);
+            $response = $client->send($request, ['timeout' => 2]);
```
bobstrecansky, line 61: This is the default timeout for guzzle; it's probably much too long but we can change it later if we deem prudent.

morrisonlevi: The [SDK/Tracing spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md) says this for a parameter to batching processors:

> exporterTimeoutMillis - how long the export can run before it is cancelled. The default value is 30000.

That default looks insanely high in traditional PHP SAPIs, but the point is that it's configurable, I think?

bobstrecansky: Yeah; I can think of a couple ways we could change this.  30000ms (30s) does feel insanely long, but we should follow the spec I guess and change it if we deem important later.  Updating.

beniamin: I would suggest using some kind of dependency injection approach. I would add the possibility to inject the http-client in the Exporter, and only if no client is injected, then I would instantiate a default http-client. I would use PSR18 http-client interface.

bobstrecansky:
Gonna lump this in with https://github.com/open-telemetry/opentelemetry-php/pull/77 to make things easier.

PR #76, opened 2020-04-06
Adding HTTP Tracing to Jaeger

bobstrecansky:
Same change as https://github.com/open-telemetry/opentelemetry-php/pull/72, only for Jaeger.  I think we could eventually DRY these two out as they are the same, but figured I would leave them for now in case we need to diverge them.


bobstrecansky:
Gonna lump this in with https://github.com/open-telemetry/opentelemetry-php/pull/77 to make things easier.

PR #77, opened 2020-04-07
WIP - Initial hrtime spike swapout

bobstrecansky:
Here is the work that I've done to try and get us using hrtime.  I think it's almost completed.

I've done the PHP7.3 upgrade locally, swapped out the time implementations we have with `hrtime` and fixed all the tests and phan.  Now I just need to see why the timestamps aren't perfectly correct for the jaeger and zipkin examples.  I think it has to do with the md5sum examples here:

I added logging to JaegerExporter.php in order to try and troubleshoot a little bit more.  Will pick it back up tomorrow.


sdk/Trace/Event.php
```diff
@@ -13,11 +12,10 @@ class Event implements API\Event
     private $timestamp;
     private $attributes;
 
-    // todo: pick datatype for timestamp
-    public function __construct(string $name, ?API\Attributes $attributes = null, string $timestamp = null)
+    public function __construct(string $name, ?API\Attributes $attributes = null, int $timestamp = null)
     {
         if (null === $timestamp) {
-            $timestamp = (new Clock())->millitime();
+            $timestamp = hrtime(true);
```
morrisonlevi, line 18: It makes sense for events to use timestamps, not durations -- these would be `microtime`s, right?

sdk/Trace/Span.php
```diff
@@ -74,28 +73,28 @@ public function setSpanStatus(int $code, ?string $description = null): API\Span
         return $this;
     }
 
-    public function end(string $timestamp = null): API\Span
+    public function end(int $timestamp = null): API\Span
```
morrisonlevi, line 76: I think we need to come up with a name for what we call the result of an `hrtime` -- `$timestamp` is misleading, and at least as written as this uses a duration (which makes sense, but we may want both).

tests/unit/Exporter/ZipkinSpanConverterTest.php
```diff
@@ -32,8 +32,8 @@ public function shouldConvertASpanToAPayloadForZipkin()
         $this->assertSame($span->getSpanName(), $row['name']);
 
         $this->assertIsInt($row['timestamp']);
-        // timestamp should be in microseconds
-        $this->assertGreaterThan(1e15, $row['timestamp']);
+        // timestamp should be in nanoseconds
```
beniamin, line 35: pay attention that Zipkin needs timestamps in microseconds not nanoseconds. Also, if you want to test for nanoseconds it should be 1e18 not 1e12, because seconds timestamp already has 10 digits + another 9 to reach nanoseconds. For microseconds you should test against 1e15.
More info about the Zipkin Spans can be found here: https://zipkin.io/zipkin-api/#/definitions/Span

sdk/Trace/Event.php
bobstrecansky: They would be.  Sorry for my mess here, hoping to clean it up today.

sdk/Trace/Span.php
bobstrecansky: Agreed; I think we need to have some better naming scheme.  I saw you asked in #open-telemetry, let's see what they say.

bobstrecansky:
I don't have a strong feeling about having an internal clock either way.  

I'm not sure we should try and optimize for 32 bit systems, but I'm happy to hear counterarguments here.

beniamin:
I dont have counterarguments for letting 32 bit version behind. I tried to find some statistics about 32 bit marketshare on linux servers, but I couldn’t find anything relevant. 
I just wanted to know your opinion.

sdk/Internal/Clock.php
```diff
@@ -12,10 +12,10 @@ public function __construct()
     {
     }
 
-    public function millitime(): string
+    public function millitime(): int
     {
-        $millitime = microtime(true) * 1000;
+        $millitime = microtime(true) * 1e+6;
```
morrisonlevi, line 17: Microseconds to milliseconds; should that be something like `microtime() / 1000`? There are 1000 microseconds to 1 millisecond, right?

bobstrecansky: Yeah I was futzing with this for a couple hrtime implementation thoughts and never changed it back.  Will change it back and validate.

bobstrecansky: Also I think you mean * 1000, as there are 1000 microseconds in a millisecond?

sdk/Internal/Clock.php
```diff
@@ -12,10 +12,10 @@ public function __construct()
     {
     }
 
-    public function millitime(): string
+    public function zipkinFormattedTime(): int
```
beniamin, line 15: I don't know if this is close to a final version, but for me, it's very odd that the Internal Clock "knows" about Zipkin time format.
In my opinion, the internal clock should return time at various resolutions (seconds, milliseconds, nanoseconds) and the ZipkinConverter should be the only one that will know how to convert the timestamps for Zipkin.

bobstrecansky: I agree with you.  I was going to make that change during the hrtime() addition, but I think Levi also had some thoughts.  Letting him share them.

bobstrecansky:
Closed in favor of https://github.com/open-telemetry/opentelemetry-php/pull/83

PR #78, opened 2020-04-19
Add clocks to the API/SDK

morrisonlevi:
I think this is the direction we want to go towards in our tracing package. We need two clocks: a monotonic and realtime.

A monotonic clock is used for calculating durations _only_. This should be used on Spans to calculate the duration of the span.

A realtime clock is used for getting timestamps _only_. This is should be used on Events to note the time of the event.

However, the spec currently says that a Span has a begin and end _timestamp_, so I suppose we have to use both a realtime clock and a monotonic clock on each span begin/end.

-----

I would like feedback on this direction before changing APIs and fixing tests. Notably, how do we want to represent the pair of clock values? We can pass them individually to every function, or we could store them in a `[$timestamp, $monotonic]` pair, or maybe something else?

### DIFF
--- a/api/Trace/MonotonicClock.php
+++ b/api/Trace/MonotonicClock.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+/**
+ * The API Specification says:
+ * OpenTelemetry can operate on time values up to nanosecond (ns) precision.
+ * The representation of those values is language specific.
+ * A duration is the elapsed time between two events.
+ *   - The minimal precision is milliseconds.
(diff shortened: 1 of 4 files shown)


bobstrecansky:
This feels like a very smart implementation to me.  The only reason CI is failing is because your PR still has PHP7.2 in the .travis.yaml (which we can't have if you want to implement `hrtime()`)

I think representing clock values as a pair is fine.  Theoretically we could have a monotonic clock time and a real clock time acting independently throughout our implementation, but the vast majority of use cases probably will need them both.

bobstrecansky:
It also looks like @lalex has thoughts on this in their PR:
https://github.com/open-telemetry/opentelemetry-php/pull/80

PR #83, opened 2020-05-11
Change to timestamps and weave the hrtime web

bobstrecansky:
**Sorry for this whopper PR; will be happy to break it down into pieces if needed.**
I've covered a bit of ground in this PR:

1. Changed our targeted PHP version to PHP 7.3 and removed PHP 7.2 as a tested version (`hrtime()` has a php7.3 hard requirement).

2. Updated our `travis.yml` to reflect this PHP 7.2 version removal.

3. Added `Trace/Clock` - this has a function called `moment()` to create an array that includes both a Monotonic (`hrtime()`) and Realtime clock (`microtime()`)  - building off of @morrisonlevi 's PR: https://github.com/open-telemetry/opentelemetry-php/pull/78 - I added the `Clock` wrapper around the realtime and monotonic clocks to make it easier, but I'm happy to separate them if we decide this convenience isn't worthwhile.

4. Swapped out all the string time definitions to ints.

5. Used the guzzle http library rather than `file_get_contents` to ship the traces to the zipkin and jaeger instances - based on my closed PRs : https://github.com/open-telemetry/opentelemetry-php/pull/72 and https://github.com/open-telemetry/opentelemetry-php/pull/76

Next steps for discussion:

1.  Where do we expect to use the monotonic clock in this library?


bobstrecansky:
Sorry for the force push, pulled in upstream and had to fix the tests that I had merged earlier today.

api/Trace/Clock.php
```diff
@@ -0,0 +1,16 @@
...
+
+interface Clock
+{
+    /**
+     * A combination of the Monotonic and Realtime Clocks
+     * Monotonic time value in the first slot, as it'll get accessed more frequently in duration calculations.
+     * @return array
+     */
+    public function moment(): array;
+    public function now(): int;
```
morrisonlevi, line 15: Let me make sure I understand this. Calling `moment()` will return an array like `[$monotonic, $real]` where `$monotonic` comes from the monotonic clock and `$real` comes from the realtime clock, right? Calling `now()` will give you the value from a realtime clock too?

composer.json
```diff
@@ -5,7 +5,8 @@
     "license": "Apache-2.0",
     "require": {
         "php": "^7.2",
```
morrisonlevi, line 7: This should be bumped to `7.3`, right?

sdk/Trace/Clock.php
```diff
@@ -0,0 +1,33 @@
...
+namespace OpenTelemetry\Sdk\Trace;
+
+use OpenTelemetry\Trace as API;
+
+class Clock implements API\Clock
+{
+    public function moment(): array
+    {
+        $realtime_clock = new RealtimeClock();
+        $monotonic_clock = new MonotonicClock();
```
morrisonlevi, line 14: These should be made in the constructor, and possibly passed in instead of calling `new` in the constructor. Maybe as a compromise, accept nullable clocks and if null then make new ones?

sdk/Trace/BatchSpanProcessor.php
```diff
@@ -85,7 +84,7 @@ public function onEnd(API\Span $span): void
         if ($this->bufferReachedExportLimit() && $this->enoughTimeHasPassed()) {
             $this->exporter->export($this->queue);
             $this->queue = [];
-            $this->lastExportTimestamp = (int) \round((float) $this->clock->millitime());
+            $this->lastExportTimestamp = $this->clock->timestamp();
```
morrisonlevi, line 87: I think this is supposed to be `now()`?
```suggestion
            $this->lastExportTimestamp = $this->clock->now();
```

sdk/Trace/Event.php
```diff
@@ -13,18 +12,17 @@ class Event implements API\Event
     private $timestamp;
     private $attributes;
 
-    // todo: pick datatype for timestamp
-    public function __construct(string $name, ?API\Attributes $attributes = null, string $timestamp = null)
+    public function __construct(string $name, ?API\Attributes $attributes = null, int $timestamp = null)
     {
         if (null === $timestamp) {
-            $timestamp = (new Clock())->millitime();
+            $timestamp = (new Clock())->timestamp();
```
morrisonlevi, line 18: I think the timestamp should be required; this way `Event` doesn't have to know anything about clocks.

sdk/Trace/Event.php
```diff
@@ -36,12 +34,12 @@ public function getAttributes(): API\Attributes
...
     }
 
-    public function getName() : string
+    public function getName(): string
     {
         return $this->name;
     }
 
-    public function getTimestamp(): string
+    public function getTimestamp(): ?int
```
morrisonlevi, line 42: I don't think this can be null.

api/Trace/Clock.php
bobstrecansky: That was my initial intention.  If it works with you I think we can roll with it (and it might be worth adding additional documentation around these design decisions).

composer.json
bobstrecansky: Yup; this was just missed when I was porting this over.  Will fix.

sdk/Trace/Event.php
bobstrecansky: I agree with you.

sdk/Trace/Clock.php
bobstrecansky: I think adding these in the constructor does make more sense.  I'll change that.

sdk/Trace/Event.php
bobstrecansky: Seems like a reasonable ask.  I think that is a fair expectation.

api/Trace/Clock.php
bobstrecansky: We also talked in our meeting about making this `Clock` call a singleton.

api/Trace/Events.php
```diff
@@ -10,10 +10,10 @@ interface Events extends \IteratorAggregate, \Countable
      * Adding an event should not invalidate nor change any existing iterators.
      * @param string $name
      * @param Attributes|null $attributes
-     * @param string|null $timestamp
+     * @param int|null $timestamp
      * @return Events Must return $this to allow setting multiple attributes at once in a chain.
      */
-    public function addEvent(string $name, ?Attributes $attributes = null, ?string $timestamp = null): Events;
+    public function addEvent(string $name, ?Attributes $attributes = null, ?int $timestamp = null): Events;
```
bobstrecansky, line 16: I'm wondering if we should make this `addEvent` timestamp required as well.

morrisonlevi: Not as important to me, but I'd at least put its parameters in the same order in case we decide to make it required.

PR #94, opened 2020-05-19
Change how SpanProcessors are added

morrisonlevi:
SpanProcessors are now added using the `addSpanProcessor()` method on the `TracerProvider`. This is what Python does, and the updated spec vaguely recommends this.

To simplify this somewhat, I added a new class `MultiSpanProcessor` that forwards all its methods to a list of `SpanProcessors`. Naming suggestions welcome; this is what Python named it but I don't love it. 

Also removed the singleton-ness of the `TracerProvider` in favor the Just Make One idiom. This is because once Sampling is added to the TracerProvider (not in this PR) it no longer makes sense to use the same object.

Also added `isRecording()` to the `Span` API; previously only existed in the SDK but the spec requires it.

-----

I did not update any examples.


sdk/Trace/Span.php
```diff
@@ -108,8 +108,6 @@ public function getStatus(): API\SpanStatus
         return SpanStatus::new($this->statusCode, $this->statusDescription);
     }
 
-    // I think this is too simple, see: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#isrecording
-    // -> This had an update this past month: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#isrecording
```
morrisonlevi, line 112: These links were broken.

sdk/Trace/MultiSpanProcessor.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace OpenTelemetry\Sdk\Trace;
+
+use OpenTelemetry\Trace as API;
+
+/**
+ * Class MultiSpanProcessor is a SpanProcessor that forwards all events to an
+ * array of SpanProcessors.
+ */
+final class MultiSpanProcessor implements SpanProcessor
```
beniamin, line 13: Do you think SpanProcessorsCollection will be a better name?

bobstrecansky: Is there a reason you're interested in changing this @beniamin?

beniamin: MultiSpanProcessor is taking me to the idea that this is a processor that processes multiple spans. In reality this class is a collection of processors.

bobstrecansky: Sounds like a decent compromise to me; you have thoughts @morrisonlevi?

michaelorr: I tend to agree, the `MultiSpanProcessor` name is a bit misleading. Also, in this design the `MultiSpanProcessor` isn't **actually** a `SpanProcessor` at all. Although it fits the API/interface of a `SpanProcessor` it is really just a meta-object used by the `TracerProvider` to delegate the calling of `onStart` `onEnd` and `shutdown` to the actual `SpanProcessors`.

It makes sense to delegate this behavior outside of the `TracerProvider` proper and let a meta object handle the actual work of dispatching calls to all the registered `SpanProcessors`. I scratched my head for too long to think of a name that matched this delegated fan-out logic but they were all clunkier than `SpanProcessorsCollection`

sdk/Trace/TracerProvider.php
```diff
@@ -6,69 +6,44 @@
...
+        $this->spanProcessors->addSpanProcessor($processor);
 
-        $spanContext = SpanContext::generate();
+        return $this;
+    }
 
-        return $this->tracers[$name] = new Tracer($this->spanProcessors, $spanContext);
+    public function getSpanProcessor(): MultiSpanProcessor
+    {
+        return $this->spanProcessors;
```
michaelorr, line 47: This is super petty but actually made me re-read things three times to make sure I was following...
The cardinality of the vars and funcs don't line up here and I think it's partly due to holdover naming from before your PR.

Previously there was a class level var that held an array of `SpanProcessors` aptly named `$spanProcessors`. In your new implementation, the var has changed to be a single instance of the meta class `MultiSpanProcessor` (or `SpanProcessorCollection` or whatever it ends up being called) but the var name was unchanged from `$spanProcessors`.

Then we have `getSpanProcessor()` (singular) and returns `$spanProcessors` (plural).

All that to suggest: If we renamed `MultiSpanProcessor` to `SpanProcessorCollector` then the class level var here could be renamed from `$spanProcessors` to `$spanProcessorCollector` and `getSpanProcessor()` could either become `getSpanProcessorCollector()` or at a minimum `getSpanProcessors()` (plural)

sdk/Trace/MultiSpanProcessor.php
morrisonlevi: SpanMultiProcessor?

michaelorr: I don't *love* it ... but that's alright, something about the use of `Multi` that throws my brain off. It still works though. Honestly, naming is hard and I don't want to beleaguer the point too much when I don't think the end result will be substantially better. `SpanMultiProcessor` works for me @bobstrecansky @beniamin thoughts on `SpanMultiProcessor`?

bobstrecansky: That's totally fine with me.

beniamin: I’m ok with it.

bobstrecansky:
I fixed the merge conflicts in this PR.  @michaelorr had some holdover questions but we could make further changes in another PR if we choose to do so.  Y'all have thoughts here?

bobstrecansky:
well, attempted to fix them.  Going to fix them for real now.

bobstrecansky:
The test coverage change here is expected as part of the change.  I think this is ready for merge if we deem prudent.

PR #95, opened 2020-05-21
Initial Correlation Context API Definition

bobstrecansky:
This is my initial stab at the Correlation Context API Definition.  I have a couple questions for the other contributors:

1.  How are we planning on passing the `$Context` variable around?  I have it earmarked here as an array, as that feels like the right data type to me.  The spec for Context is delivered [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md) - I'm planning on stubbing it out in a subsequent PR.

2.  Where should we implement this?  It looks like other sigs have implemented it in different places within their repositories.  As an example, [Python](https://github.com/open-telemetry/opentelemetry-python/tree/master/opentelemetry-api/src/opentelemetry) implemented `Context` right next to `CorrelationContext` - this doesn't feel natural for our repo setup.  I'm open to options here.  Context is defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md)

3.  We'll need to discuss how and where we want to implement [Serialization](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/correlationcontext/api.md#serialization).

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -357,6 +357,7 @@ return [
     // your application should be included in this list.
     'directory_list' => [
         'api',
+        'Context',
         'sdk',
         'vendor/composer/xdebug-handler/src',
         'vendor/guzzlehttp',
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -9,7 +9,7 @@ class Context
     /**
      * @var ContextKey|null
      */
-    private $key;
+    protected $key;
 
     /**
      * @var mixed|null
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -34,6 +34,7 @@ class Context
      * @param ContextKey $key
      * @param mixed $value
      *
+     * @param Context|null $parent
      * @return Context
      */
     public static function setValue(ContextKey $key, $value, ?Context $parent=null): Context
@@ -87,6 +88,7 @@ class Context
     }
 
     /**
+     * @param Context $ctx
      * @return callable
      */
     public static function attach(Context $ctx): callable
@@ -100,19 +102,11 @@ class Context
     }
 
     /**
+     * @param callable $token
+     * @return Context
      */
     public static function detach(callable $token): Context
     {
         return self::$current_context = call_user_func($token);
-    }
-
-    /**
-     * @param Context $parent
-     *
-     * @return null
-     */
-    protected function setParent(Context $parent)
-    {
-        $this->parent = $parent;
     }
 }
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -105,4 +105,14 @@ class Context
     {
         return self::$current_context = call_user_func($token);
     }
+
+    /**
+     * @param Context $parent
+     *
+     * @return null
+     */
+    protected function setParent(Context $parent)
+    {
+        $this->parent = $parent;
+    }
 }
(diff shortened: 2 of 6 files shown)


beniamin:
> 1. How are we planning on passing the `$Context` variable around?  I have it earmarked here as an array, as that feels like the right data type to me.  The spec for Context is delivered [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md) - I'm planning on stubbing it out in a subsequent PR.

An array indexed by Tokens sounds good to me.

> 2. Where should we implement this?  It looks like other sigs have implemented it in different places within their repositories.  As an example, [Python](https://github.com/open-telemetry/opentelemetry-python/tree/master/opentelemetry-api/src/opentelemetry) implemented `Context` right next to `CorrelationContext` - this doesn't feel natural for our repo setup.  I'm open to options here.  Context is defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md)

I see that the specs keep the `Context` and `CorrelationContext` in different namespaces. For me this is a more natural way to keep them separated.

api/CorrelationContext/Correlations.php
```diff
@@ -0,0 +1,78 @@
...
+    * The order of name/value pairs MUST NOT be significant.
+    * Based on the language specification, the returned value can be either an immutable collection
+    * or an immutable iterator to the collection of name/value pairs in the CorrelationContext.
+    * OPTIONAL parameters:
+    * Context: the context containing the CorrelationContext from which to get the correlations.
+    *
+    * @param ?array $Context
+    * @return array Returns the name/value pairs in the CorrelationContext.
+    */
+    public function getCorrelations(?array $Context): array;
```
beniamin, line 27: I think that we must define a new API for Context and use that API here. I don't think that the Context can be modeled properly using an array, because the specs say that the Context should be immutable and any change on a Context should result in a new Context.

api/CorrelationContext/Correlations.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\CorrelationContext;
```
beniamin, line 5: I think is a good thing to keep CorrelationContext as a separate namespace.

api/CorrelationContext/Correlations.php
bobstrecansky: Agreed with you; I started implementing this and then started reading through the context specification.  I'll change this to use the context definition after we've created one.

beniamin: Also, the specs say that the getCorrelations method may receive a Context and if so, then it must return the  Correlations tied to that Context.

bobstrecansky: Will change that as well when the time comes.

bobstrecansky:
@kneitinger - I believe this is the PR you talked about in our meeting yesterday that you'd like to continue to feed off after @michaelorr 's [context PR](https://github.com/open-telemetry/opentelemetry-php/pull/111) is merged

bobstrecansky:
Closing in favor of https://github.com/open-telemetry/opentelemetry-php/pull/137

Issue #85, opened 2020-05-12
Implement the shutdown function


bobstrecansky:
We should implement a `shutdown()` function

The specification defines the shutdown SDK as so: [shutdown](https://github.com/open-telemetry/opentelemetry-specification/blob/e0bd41739a92304c0be5d68840f434e281b5f8b2/specification/trace/sdk.md#shutdown)

We have a placeholder defined for this in a couple places in our codebase:

* our BatchSpanProcessor, defined [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/BatchSpanProcessor.php#L109)
* our JaegerExporter, defined [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/JaegerExporter.php#L93)
* our SimpleSpanProcessor, defined [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SimpleSpanProcessor.php#L44)
* our ZipkinExporter, defined [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/ZipkinExporter.php#L87)

Success criteria: implement a shutdown method.  We should most likely start with a static shutdown timeout and determine if we eventually would like to make it end user configurable.  The spec allows us to make our own determination on the timeout:

```
Shutdown should not block indefinitely. Language library authors can decide if they want to make the shutdown timeout configurable.
```

PR #100, opened 2020-05-27
Implement SpanProcessor::shutdown()

michaelorr:
The diff isn't crazy big here, but unfortunately there are several logical changes combined. None of the individual moving parts would make sense without the bigger picture of what they enabled, so it may actually be hard to review in chunks without seeing the end state but I'm happy to break the PR up into logical chunks if desired.

This change also:
- Implements `shutdown()` for `SimpleSpanProcessor` and `BatchSpanProcessor`
- Implements `shutdown()` for the Jaeger and Zipkin Exporter implementations
- Makes Exporters optional for `SpanProcessors` (per spec)
- Changes the `BatchSpanProcessor` such that if *either* the configured batch size OR
the configured delay is reached, then a batch is sent, both are no longer required
- Adds `SpanProcessor::forceFlush()` (per spec)

I didn't see any logging in our existing code. Some of the other language libraries do some logging amount of logging (for example when `SpanProcessor`'s `$maxQueueSize` is reached. I've never authored a php library so I wasn't sure if good solutions exist to enable us to log that don't stomp on the app developers integration.

This would close: https://github.com/open-telemetry/opentelemetry-php/issues/85

### DIFF
--- a/sdk/Trace/BatchSpanProcessor.php
+++ b/sdk/Trace/BatchSpanProcessor.php
@@ -11,7 +11,7 @@ use OpenTelemetry\Trace as API;
 class BatchSpanProcessor implements SpanProcessor
 {
     /**
-     * @var Exporter
+     * @var Exporter|null
      */
     private $exporter;
     /**
@@ -44,8 +44,13 @@ class BatchSpanProcessor implements SpanProcessor
      */
     private $clock;
 
+    /**
+     * @var bool
+     */
+    private $running = true;
+
     public function __construct(
-        Exporter $exporter,
+        ?Exporter $exporter,
         Clock $clock,
         int $maxQueueSize = 2048,
         int $scheduledDelayMillis = 5000,
@@ -77,11 +82,23 @@ class BatchSpanProcessor implements SpanProcessor
      */
     public function onEnd(API\Span $span): void
     {
-        if (count($this->queue) < $this->maxQueueSize) {
-            $this->queue[] = $span;
-        }
+        if (null !== $this->exporter && $this->running) {
+            if (count($this->queue) < $this->maxQueueSize) {
+                $this->queue[] = $span;
+            }
 
-        if ($this->bufferReachedExportLimit() && $this->enoughTimeHasPassed()) {
+            if ($this->bufferReachedExportLimit() || $this->enoughTimeHasPassed()) {
+                $this->forceFlush();
+            }
+        }
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function forceFlush(): void
+    {
+        if (null !== $this->exporter && $this->running) {
             $this->exporter->export($this->queue);
             $this->queue = [];
             $this->lastExportTimestamp = $this->clock->timestamp();
@@ -105,6 +122,10 @@ class BatchSpanProcessor implements SpanProcessor
      */
     public function shutdown(): void
     {
-        // TODO: Implement shutdown() method.
+        $this->running = false;
+        if (null !== $this->exporter) {
+            $this->forceFlush();
+            $this->exporter->shutdown();
+        }
     }
 }
--- a/sdk/Trace/SimpleSpanProcessor.php
+++ b/sdk/Trace/SimpleSpanProcessor.php
@@ -9,11 +9,16 @@ use OpenTelemetry\Trace as API;
 class SimpleSpanProcessor implements SpanProcessor
 {
     /**
-     * @var Exporter
+     * @var Exporter|null
      */
     private $exporter;
 
-    public function __construct(Exporter $exporter)
+    /**
+     * @var bool
+     */
+    private $running = true;
+
+    public function __construct(?Exporter $exporter)
     {
         $this->exporter = $exporter;
     }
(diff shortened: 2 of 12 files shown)


sdk/Trace/BatchSpanProcessor.php
```diff
@@ -11,7 +11,7 @@
 class BatchSpanProcessor implements SpanProcessor
 {
     /**
-     * @var Exporter
+     * @var Exporter|null
```
michaelorr, line 14: > Each processor registered on TracerProvider is a start of pipeline that consist of span processor and optional exporter. SDK MUST allow to end each pipeline with individual exporter.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor

sdk/Trace/BatchSpanProcessor.php
```diff
@@ -77,11 +82,23 @@ public function onStart(API\Span $span): void
...
     public function onEnd(API\Span $span): void
     {
-        if (count($this->queue) < $this->maxQueueSize) {
-            $this->queue[] = $span;
+        if (null !== $this->exporter && $this->running) {
+            if (count($this->queue) < $this->maxQueueSize) {
+                $this->queue[] = $span;
+            }
+
+            if ($this->bufferReachedExportLimit() || $this->enoughTimeHasPassed()) {
```
michaelorr, line 90: The spec is very vague on the interplay between `maxExportBatchSize` and `scheduledDelayMillis` but both the python and go SDK implementations appear to send a batch for whichever occurs first, not necessarily waiting for both conditions to be met. This change brings the php sdk in line with the other implementations and it especially makes sense for us given that we aren't running a concurrent thread to monitor queue depth or timeout delay, so the `scheduledDelayMillis` is only being checked during calls to `onEnd` and there likely will be stretches of time that are far longer than this configured time between exports.

sdk/Trace/BatchSpanProcessor.php
```diff
@@ -77,11 +82,23 @@ public function onStart(API\Span $span): void
...
+                $this->forceFlush();
+            }
         }
+    }
 
-        if ($this->bufferReachedExportLimit() && $this->enoughTimeHasPassed()) {
+    /**
+     * @inheritDoc
+     */
+    public function forceFlush(): void
```
michaelorr, line 99: > ForceFlush()
Export all ended spans to the configured Exporter that have not yet been exported.
ForceFlush should only be called in cases where it is absolutely necessary, such as when using some FaaS providers that may suspend the process after an invocation, but before the Processor exports the completed spans.
ForceFlush should not block indefinitely. Language library authors can decide if they want to make the flush timeout configurable.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#forceflush

sdk/Trace/SimpleSpanProcessor.php
```diff
@@ -23,24 +28,38 @@ public function __construct(Exporter $exporter)
...
     /**
      * @inheritDoc
      */
     public function onEnd(API\Span $span): void
     {
-        // @todo only spans with SampleFlag === true should be exported according to
-        // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#sampling
-        $this->exporter->export([$span]);
+        if ($this->running) {
+            // @todo only spans with SampleFlag === true should be exported according to
```
michaelorr, line 39: I considered tackling this in this same PR but I figured *yet another* logical change in the same diff should be avoided if possible.

sdk/Trace/TracerProvider.php
```diff
@@ -43,6 +43,15 @@ class TracerProvider implements API\TracerProvider
...
         $this->spanProcessors = $spanProcessors;
+
+        register_shutdown_function([$this, 'shutdown']);
+    }
+
+    public function shutdown(): void
+    {
+        foreach ($this->spanProcessors as $spanProcessor) {
+            $spanProcessor->shutdown();
+        }
```
michaelorr, line 54: This `TracerProvider::shutdown()` helper is only needed while the `TracerProvider` is managing the array of `$spanProcessors`... once open-telemetry#94 is merged, this helper can go away and the callback registration in the constructor can look like:
`register_shutdown_function([$this->spanProcessors, 'shutdown'])` since it will handle the "fan-out" to all the other downstream `SpanProcessors`

bobstrecansky:
> I have signed the CLA now, not sure how to re-rerun that check.

You're good to go here!

bobstrecansky:
Waiting on a merge of https://github.com/open-telemetry/opentelemetry-php/pull/94 before we merge this because of dependencies.

bobstrecansky:
I think this is ready for merge after the conflict is resolved.

sdk/Trace/TracerProvider.php
bobstrecansky: #94 has been merged so we can remove this helper at will.

sdk/Trace/SimpleSpanProcessor.php
bobstrecansky: We should probably create a bite sized issue to tackle this.

Issue #92, opened 2020-05-18
Implement the resource provider


bobstrecansky:
We need to implement the (newly created) resource provider:

https://github.com/open-telemetry/opentelemetry-specification/blob/27e5c89075d3eeb5981e79bf8ab94d78aea5481d/specification/resource/sdk.md#resource-sdk

This was recently added to the specification and our code should follow along with this.

bobstrecansky:
This issue can be closed when #101 is merged.

bobstrecansky:
This has been completed.

PR #101, opened 2020-05-27
Add Resource SDK

lalex:
This is an implementation of `Resource` SDK (#92) and recently added `InstrumentationLibrary` ([TracerCreation](https://github.com/open-telemetry/opentelemetry-specification/blob/1a0395e482d4fec6faab0ee4b7bc1dc5fdd3e493/specification/trace/sdk.md#tracer-creation)).

In this PR minimal changes were made to `TracerProvider` and `Tracer` to simplify a merge with the concurrent PRs.

The class for `Resource` is called `ResourceInfo` because `resource` is the reserved word in PHP and it makes difficulties for code analysis tools. Any suggestion on name are welcome.

In specification `Resource` is proposed as immutable object. Now it is not fully immutable since it uses `Attributes` collection which is mutable.

### DIFF
--- a/sdk/Resource/ResourceInfo.php
+++ b/sdk/Resource/ResourceInfo.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Resource;
+
+use OpenTelemetry\Sdk\Trace\Attributes;
+
+/**
+ * A Resource is an immutable representation of the entity producing telemetry. For example, a process producing telemetry
+ * that is running in a container on Kubernetes has a Pod name, it is in a namespace and possibly is part of a Deployment
+ * which also has a name. All three of these attributes can be included in the Resource.
+ *
+ * The class named as ResourceInfo due to `resource` is the soft reserved word in PHP.
+ */
+class ResourceInfo
+{
+    private $attributes = [];
+
+    private function __construct(Attributes $attributes)
+    {
+        $this->attributes = $attributes;
+    }
+
+    public static function create(Attributes $attributes): self
+    {
+        return new ResourceInfo(clone $attributes);
+    }
+
+    /**
+     * Merges two resources into a new one.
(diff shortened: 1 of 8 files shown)


sdk/Resource/ResourceInfo.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace OpenTelemetry\Sdk\Resource;
+
+use OpenTelemetry\Sdk\Trace\Attributes;
+
+/**
+ * A Resource is an immutable representation of the entity producing telemetry. For example, a process producing telemetry
+ * that is running in a container on Kubernetes has a Pod name, it is in a namespace and possibly is part of a Deployment
+ * which also has a name. All three of these attributes can be included in the Resource.
+ */
```
bobstrecansky, line 13: It might be prudent to make a note about how we are using `ResourceInfo` rather than `resource` here because of the reserved PHP word.

bobstrecansky:
Lalex,

I think this is a very great start on the Resource SDK and TracerCreation Library!  You've done a great job of being conscientious of impacting changes; thank you for doing that.  

I like your recommendation of `ResourceInfo`, it seems like it still gets the point across while being far enough away from the reserved word.  It's unfortunate in this case that `resource` is a reserved word in PHP in this particular case.  We should potentially put a comment about this inline just for future reference; I made a comment where I think you may be able to add this information.

We should probably create a ticket to make sure that the Attributes collection isn't mutable if that's required to keep the Resource as an immutable object.  [The spec doesn't explicitly mention that the Attributes collection should be immutable](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes), so maybe we should ask the spec maintainers if that was intentional.

bobstrecansky:
@lalex do you have additional thoughts here?

lalex:
@bobstrecansky I've merged the latest master and resolved todos and notes.

bobstrecansky:
Verifying that no one else has any qualms in the gitter channel; will merge afterwords.

PR #111, opened 2020-06-03
Add `context` data structures.

michaelorr:
This adds a naive implementation of `context` storage appropriate for
use by higher level library constructs such as CorrelationContext and
SpanContext.

Example Static Usage:

>>> $key = new ContextKey();
>>> $ctx = Context::setValue($key, $val);
>>> Context::getValue($key);
$val

Example Instance Usage:
>>> $key = new ContextKey();
>>> $ctx = (new Context())->set($key, $val);
>>> $ctx->get($key);
$val

The above can be mixed and matched.


bobstrecansky:
I'm really happy with this pull request; I'd just like to get some other (non Mailchimp) opinions on it before I merge it.

lalex:
In #64 the API/SDK segregation was introduced. I think we should follow it with the new components.

bobstrecansky:
@lalex - I believe the reason this was written into a separate namespace is that this doesn't really fall into either of the API / SDK buckets. [The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md#overview) doesn't include Context in the API or the SDK (most likely because many languages already have a concept of a `Context`.  I think it is okay for this to live in a separate namespace.

michaelorr:
@lalex yeah, @bobstrecansky's assessment was my thought process as well. Basically the spec doesn't outline the `Context` as part of either the API or SDK but just as something that needs to exist in the language runtime _in general_... The spec is pretty sparse on detail but the OTEP is actually pretty informative in case you hadnt read it: https://github.com/open-telemetry/oteps/blob/master/text/0066-separate-context-propagation.md

Language ecosystems like go and python provide their own and can promptly ignore the `Context` in the spec, php does not provide a construct that meets our needs so this is it.
This will be used by components in the SDK (and possibly the API depending on future choices) but in theory this could even live in a separate repo and be vendored in via composer to further distinguish that the `Context` class is a separate thing. If we were to go that route, the only changes that would be needed are the division of the notion of the `current_context` along with `attach` `detach` and `getCurrent` since those are unique to OTs usage of the `Context` class.

PR #112, opened 2020-06-03
Phpcs fix in shell script

bobstrecansky:
This PR accomplishes 2 things:

* Updates the version we test with from Trusty (16.04) to Bionic (18.04).  Trusty is out of standard support, so I figured it was best to move to the next version.

* Moves the travis phpcs test from within the `.travis.yml` file to its own separate script in `.travis/phpcs.sh`.  This makes the [output from CI](https://travis-ci.org/github/open-telemetry/opentelemetry-php/jobs/694432862#L504-L509) a little easier to read and grok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 language: php
 os: linux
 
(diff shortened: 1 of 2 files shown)


bobstrecansky:
I would have loved to, but It wasn't yet available on Travis CI.

PR #125, opened 2020-06-12
rough pass at correlationcontext

michaelorr:


sdk/CorrelationContext/CorrelationContext.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\CorrelationContext;
+
+use OpenTelemetry\Context\Context;
+
+class CorrelationContext implements Context
```
bobstrecansky, line 9: I think this needs to be `extends Context` rather than `implements Context`

bobstrecansky:
A couple things I wanna bring up when we talk about this:

I ran phan and fixed a couple of the errors it was reporting about, but I want to go over them with you so we are on the same page.  I created this so I can track what I was doing:
https://github.com/open-telemetry/opentelemetry-php/commit/6c868c7a715289afde99603ed81ecc6641acd7ba


A couple key ones:

1)  I think the ContextKey value $key in Context/Context.php might need to be protected rather than private.
2)  I think the Context value $parent in Context/Context.php might need to be protected rather than private.
3) `setParent` might need to be public rather than protected since we are referencing it in CorrelationContext.
4) We need to import `ContextKey` in api/CorrelationContext/CorrelationContext.php

We need to add `Context` to the phan directory list in `.phan/config.php too.


I created https://github.com/open-telemetry/opentelemetry-php/compare/master...bobstrecansky:correlation_context_phan_appeasement?expand=1

michaelorr:
Closing in favor of https://github.com/open-telemetry/opentelemetry-php/pull/132

bobstrecansky:
This will be completed with https://github.com/open-telemetry/opentelemetry-php/pull/132

PR #126, opened 2020-06-17
Fix broken zipkin span converter test

michaelorr:


michaelorr:
With the 2 PRs that I just opened: 
Screenshot 2020-06-17 at 1 56 09 AM

michaelorr:
This is now rebased on master after @beniamin 's changes. He actually also fixed this test in a slightly different way so this PR isn't exactly necessary anymore, but I combined our two approaches and I think this reduces a lot of the boilerplate and removes an assert that is entirely unrelated to the zipkinexporter... having asserts that are unrelated to the "code under test" can cause noisy failures that aren't helpful when that behavior should be covered under its own test elsewhere.

PR #129, opened 2020-06-21
Fix - Remove debug code from Zipkin exporter

atrauzzi:
- This change prevents output from being sent to the client during the request.


atrauzzi:
I signed it!

atrauzzi:
:bow:

PR #131, opened 2020-06-23
change context to be trait

michaelorr:
Convert `Context` from `class` to `trait`
This should allow for a better approach at the CorrelationContext and `SpanContext` layer


michaelorr:
Let's hold off on merging this until we rebase the `CorrelationContext` branch on top and make sure this approach feels right when its being used.

michaelorr:
Closing in favor of https://github.com/open-telemetry/opentelemetry-php/pull/132

PR #132, opened 2020-06-26
Correlation context with trait

bobstrecansky:
This PR is creating the CorrelationContext with our newly defined trait Correlation.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -357,6 +357,7 @@ return [
     // your application should be included in this list.
     'directory_list' => [
         'api',
+        'Context',
         'sdk',
         'vendor/composer/xdebug-handler/src',
         'vendor/guzzlehttp',
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Context;
 
-class Context
+trait Context
 {
     /**
      * @var ContextKey|null
@@ -23,7 +23,24 @@ class Context
 
     private static $current_context = null;
 
-    public function __construct(ContextKey $key=null, $value=null, ?Context $parent=null)
+    /**
+     * The constructor for a Context object. Because this is a trait, a `Context` cannot be instantiated directly,
+     * this constructor is intended for use by the classes that use this trait. Throughout the docsblocks and comments
+     * in this class, we will refer to "Context instance" which really refers to an instance of the class
+     * that uses this trait.
+     *
+     * This is a general purpose read-only key-value store. Read-only in the sense that adding a new value does not
+     * mutate the existing context, but returns a new Context which has the new value added.
+     *
+     * In practical terms, this is implemented as a linked list of Context instances, with each one holding a reference
+     * to the key object, the value that corresponds to the key, and an optional reference to the parent Context
+     * (i.e. the next link in the linked list chain)
+     *
+     * @param ContextKey|null $key The key object. Should only be null when creating an "empty" context
+     * @param mixed|null $value
+     * @param Context|null $parent Reference to the parent object
+     */
+    public function __construct(?ContextKey $key=null, $value=null, $parent=null)
     {
         $this->key = $key;
         $this->value = $value;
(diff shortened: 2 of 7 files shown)


Context/Context.php
```diff
@@ -23,31 +23,75 @@ class Context
...
+     *    Context::setValue($key, $value, $ctx) is functionally equivalent to $ctx->set($key, $value)
+     * 2) Without a $parent parameter:
+     *    In this scenario, setValue() will use the `$current_context` reference as supplied by `getCurrent()`
+     *    `getCurrent()` will always return a valid Context/__CLASS__. If one does not exist at the global scope,
+     *    an "empty" context will be created.
+     *
+     * @param ContextKey $key
+     * @param mixed $value
+     * @param Context|null $parent
+     * @suppress PhanTypeInstantiateTrait
```
bobstrecansky, line 84: I had to add this Phan suppression in because Phan didn't like the trait call to Context in this function call.  Ideas welcome here!

Context/Context.php
```diff
@@ -61,48 +105,53 @@ public function get(ContextKey $key)
             return $this->value;
         }
         if (null === $this->parent) {
-            throw new ContextValueNotFoundException();
+            throw new ContextValueNotFoundException('HUH');
```
bobstrecansky, line 108: This was left from our debugging; we could probably change it to something more meaningful.

sdk/CorrelationContext.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $this->removeCorrelationHelper($key, null);
+
+        return $this;
+    }
+
+    /**
+     * @param ContextKey $key
+     * @param CorrelationContext|null $child
+     * @suppress PhanUndeclaredMethod
```
bobstrecansky, line 41: I added this to appease the CI for now, but I think we might need to address it.

sdk/CorrelationContext.php
```diff
@@ -0,0 +1,72 @@
...
+     * @suppress PhanTypeMismatchArgument
```
bobstrecansky, line 42: I added this to appease the CI for now, but I think we might need to address it.

michaelorr: One possible solution to this and the `PhanUndeclaredMethod` is to add:

```
    /**
     * @var CorrelationContext|null
     */
    private $parent;
```
to the top of `CorrelationContext` which I think kinda makes sense... within a `CorrelationContext`, we know because of the structure of the logic that `$this->parent` will always be a `CorrelationContext` but the type system cannot assert this. 

I'm looking at alternatives, specifically if adding an inline type check would be better

Another option is to change this function to look like this:
```
    private function removeCorrelationHelper(ContextKey $key, ?CorrelationContext $child)
    {
        $parent = $this->parent;
        '@phan-var CorrelationContext|null $parent';
        if ($this->key !== $key) {
            if (null === $parent) {
                return;
            }

            $this->parent->removeCorrelationHelper($key, $this);
            return;
        }

        $child->setParent($parent);
    }

```

But I kinda hate that... its very weird.


The other option is to add a type assertion to the helper:

```
    private function removeCorrelationHelper(ContextKey $key, ?CorrelationContext $child)
    {
+       assert(($this->parent instanceof CorrelationContext) || (is_null($this->parent)));
        if ($this->key !== $key) {
            if (null === $this->parent) {
                return;
            }
        ...
```

I might hate this one the least? I dunno, that or just mirroring the `$parent` into the top of `CorrelationContext` I dunno

Context/Context.php
bobstrecansky: Yeah it feels like a bug in Phan to me too.

michaelorr: https://github.com/phan/phan/issues/3988

bobstrecansky:
This will be completed with https://github.com/open-telemetry/opentelemetry-php/pull/132

.phan/config.php
```diff
@@ -357,6 +357,7 @@
     // your application should be included in this list.
     'directory_list' => [
         'api',
+        'Context',
```
morrisonlevi, line 360: So... the context is part of neither the API nor sdk?

bobstrecansky: That's correct.  The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/context.md) has correlation living as its own thing.  Context is implemented at a language level in many other languages, so they were attempting to decouple the Context implementation from CorrelationContext so that authors could either write their own Correlation implementation or adapt their languages' Correlation packages.  The CorrelationContext is part of the API.

bobstrecansky:
Closing this in favor of https://github.com/open-telemetry/opentelemetry-php/pull/137

PR #134, opened 2020-06-28
external http client on exporters

beniamin:
I added the possibility to inject external HTTP client to the exporters, in order to ease the configuration of it. I thought this will be the simplest way, to have a configured HTTP client.

Also, I improved the tests coverage and implement the exporter return based on zipkin/jaeger server response code.

### DIFF
--- a/contrib/Zipkin/Exporter.php
+++ b/contrib/Zipkin/Exporter.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Contrib\Zipkin;
 
-use Exception;
 use GuzzleHttp\HandlerStack;
 use GuzzleHttp\Middleware;
 use GuzzleHttp\Psr7\Request;
+use Http\Adapter\Guzzle6\Client;
 use InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


contrib/Zipkin/Exporter.php
```diff
@@ -51,8 +64,9 @@ public function __construct($name, string $endpointUrl, SpanConverter $spanConve
         }
 
         $this->endpointUrl = $endpointUrl;
-
+        $this->client = $client ?? $this->createDefaultClient();
         $this->spanConverter = $spanConverter ?? new SpanConverter($name);
+        $this->client = $client;
```
michaelorr, line 69: It looks like this line will over-ride the null-coalesce on line 67?
When `$client` is `null`, line 67 will set `$this->client` to `$this->createDefaultClient()` and then line 69 will set `$this->client` to `null`

beniamin: you are right, I will fix it right away.

beniamin: done

PR #137, opened 2020-07-04
Context and correlationcontext inheritance

michaelorr:
Add `Context` and `CorrelationContext` via standard inheritance.

When subclassing `Context` it is necessary to shadow the `$parent` attr to ensure that all operations on a `CorrelationContext` also return a `CorrelationContext` instead of a `Context`

### DIFF
--- a/api/CorrelationContext.php
+++ b/api/CorrelationContext.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\CorrelationContext;
+
+use OpenTelemetry\Context\ContextKey;
+
+interface CorrelationContext
+{
+    public function getCorrelations(); // TODO
+
+    public function get(ContextKey $key);
+
+    public static function getValue(ContextKey $key, ?CorrelationContext $ctx);
+
+    public function set(ContextKey $key, $value): CorrelationContext;
+
+    public static function setValue(ContextKey $key, $value, ?CorrelationContext $parent = null): CorrelationContext;
+
+    public function removeCorrelation(): CorrelationContext;
+
+    public function clearCorrelations(); // TODO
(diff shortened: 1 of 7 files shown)


api/CorrelationContext.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\CorrelationContext;
+
+use OpenTelemetry\Context\ContextKey;
+
+interface CorrelationContext
```
beniamin, line 9: After going multiple times trough the specs I understood that `CorrelationContext` it's a 'dictionary' used across multiple threads/processes. But, since PHP is single-threaded I don't think that this type of context will add real value to our library, and maybe we should drop it.

bobstrecansky: @beniamin - what would you suggest instead?  I think our implementation of CorrelationContext feels important, but I may be sitting too close to the problem.

beniamin: I can't see when the CorrelationContext  will be used. Because php beeing single thread, I dont't think CorrelationContext will help someone. What I propose here is to drop the implementation for CorrelationContext and we will reimplement it only if someone will ask for it.

bobstrecansky: It's going to be weaved throughout tracing and metrics eventually.   [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#correlation-context-into-metric-labels) is a good example of where it's implemented in the metrics spec.  I think it'd be valuable to keep this in the codebase, as I think it'll be used in a couple of our next implementation steps.

beniamin: OK, If you think it will be useful then we should keep it.

Issue #140, opened 2020-07-04
Only spans with SampleFlag == True should be exported


bobstrecansky:
We need to only export spans that have a `SampleFlag == True`.  This is located at the following [todo](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SpanProcessor/SimpleSpanProcessor.php#L41).

Success criteria: implement this flagging logic

wRLSS:
this can be closed btw

bobstrecansky:
Thanks @wRLSS !

PR #144, opened 2020-07-07
Only sampled spans should exported

wRLSS:
Sampled span should never be exported.

https://github.com/open-telemetry/opentelemetry-php/issues/140


michaelorr:
Hey @wRLSS great contribution! Looks like you'll need to sign CNCF's Contributors License Agreement which should only take a few moments.

One question about your change, the `isSampled` flag in OT is defined according to the w3c spec for TraceContext and sampling, which comes from here: https://github.com/w3c/trace-context/blob/master/spec/20-http_request_header_format.md#sampled-flag

Namely:
> When set, the least significant bit (right-most), denotes that the caller may have recorded trace data. When unset, the caller did not record trace data out-of-band.

The language here is a little vague, but I THINK the intent is for `sampled` to imply that the trace data WAS sent out of band (or will be). The word choice is unfortunate and easy to confuse because "sampled" in other software contexts could mean "sampled away" or removed from consideration, but the way I am reading the w3c spec is that they are using "sampled" to mean "has been considered/recorded or should be considered/recorded"

All that to say, I think:
```
        if ($span->isSampled()) {
            return;
        }
```
Should actually be:
```
        if ( ! $span->isSampled()) {
           ...
```

I had to read the spec a few times and then still ended up referencing the go language lib to see what the authors over there did, which I found here:
https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/simple_span_processor.go#L46-L48
(in the simple span processor, if span IS sampled, export it)
and here:
https://github.com/open-telemetry/opentelemetry-go/blob/master/sdk/trace/batch_span_processor.go#L217-L219
(in the batch span processor, if the span is NOT sampled, throw it away and do not enqueue)

Someone will also need to make a similar change to the batch span processor, would you be willing to add that change to this PR also? If not, that is totally ok, I can add that to the batch processor later this week to mirror this addition.

wRLSS:
Sure, will provide changes later.

wRLSS:
CLA requires an email history rewrite, will reopen this.

PR #145, opened 2020-07-08
Only sampled spans should be exported

wRLSS:
Recreating this due to CLA 
https://github.com/open-telemetry/opentelemetry-php/issues/140

### DIFF
--- a/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -85,14 +85,24 @@ class BatchSpanProcessor implements SpanProcessor
      */
     public function onEnd(API\Span $span): void
     {
-        if (null !== $this->exporter && $this->running) {
-            if (count($this->queue) < $this->maxQueueSize) {
-                $this->queue[] = $span;
-            }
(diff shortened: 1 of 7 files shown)


sdk/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -85,14 +85,24 @@ public function onStart(API\Span $span): void
...
 
-            if ($this->bufferReachedExportLimit() || $this->enoughTimeHasPassed()) {
-                $this->forceFlush();
-            }
+        if (!$this->running) {
+            return;
+        }
+
+        if (!$span->isSampled()) {
+            return;
```
michaelorr, line 97: I'm wondering, even if this particular `$span` is not sampled and not exported, should we still evaluate the buffer to see if we've reached the timeout?
That doesn't make sense for the first two checks, since the processor either wouldnt have an exporter or wouldnt be running, but in the scenario where this span isnt sampled/exportable, perhaps we still want to check for the min time elapsed?

wRLSS: Agree

PR #146, opened 2020-07-13
Added setSpanKind implementation and a test for it

### LINKED ISSUE #142: Implement the setSpanKind Method
The [specification for SpanKind](
https://github.com/open-telemetry/opentelemetry-specification/blob/4daee3af756be75edbd17085f6d382d963027bc8/specification/trace/api.md#spankind) has been firmed up, so we should implement the setSpanKind method.

Success criteria:  Implement the SpanKind method as defined [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SpanOptions.php#L36-L39).

artem-prozorov:
Closes #142

### DIFF
--- a/sdk/Trace/SpanOptions.php
+++ b/sdk/Trace/SpanOptions.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace OpenTelemetry\Sdk\Trace;
 
 use OpenTelemetry\Trace as API;
-use ReflectionClass;
 
 final class SpanOptions implements API\SpanOptions
 {
--- a/sdk/Trace/SpanOptions.php
+++ b/sdk/Trace/SpanOptions.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace OpenTelemetry\Sdk\Trace;
 
 use OpenTelemetry\Trace as API;
+use ReflectionClass;
 
 final class SpanOptions implements API\SpanOptions
 {
@@ -16,6 +17,7 @@ final class SpanOptions implements API\SpanOptions
     private $parent = null;
     private $attributes = null;
     private $links = null;
+    private $kind = API\SpanKind::KIND_INTERNAL;
 
     /** @var int|null */
     private $start = null;
@@ -35,7 +37,14 @@ final class SpanOptions implements API\SpanOptions
 
     public function setSpanKind(int $spanKind): API\SpanOptions
     {
-        // TODO: Implement setSpanKind() method.
+        $reflection = new ReflectionClass(API\SpanKind::class);
+
+        if (!in_array($spanKind, $reflection->getConstants(), true)) {
+            throw new \InvalidArgumentException('You must pass a valid span kind');
+        }
+
+        $this->kind = $spanKind;
+
         return $this;
     }
 
@@ -81,7 +90,7 @@ final class SpanOptions implements API\SpanOptions
             ? SpanContext::fork($span->getContext()->getTraceId())
             : SpanContext::generate();
 
-        $span = new Span($this->name, $context, $this->parent);
+        $span = new Span($this->name, $context, $this->parent, $this->kind);
 
         if (isset($this->start)) {
             $span->setStartTimestamp($this->start);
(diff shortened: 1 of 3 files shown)


artem-prozorov:
@bobstrecansky sorry I don't see any functionality within examples that breaks with these changes. Could you please share the results of your local testing here?

Anyway, what I see:
1. https://github.com/open-telemetry/opentelemetry-php/blob/master/examples/AlwaysOnTraceExample.php - this one doesn't send anything to local zipkin instance because SpanContext is generated here with this method https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SpanContext.php#L83 and so property sampled is set to false and sending results is interrupted here: https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SpanProcessor/SimpleSpanProcessor.php#L45
2. https://github.com/open-telemetry/opentelemetry-php/blob/master/examples/AlwaysOffTraceExample.php#L21 - this one cannot be true because $samplingResult is always an object. Suppose this is some legacy.
3. https://github.com/open-telemetry/opentelemetry-php/blob/master/examples/AlwaysOffTraceExample.php#L30 and L34 - need to add timestamp

Do you mean these bugs?

bobstrecansky:
@artem-prozorov - Thanks for your thorough explanation.  

* I think we should update that integration test for (1), but that shouldn't preclude us from merging this!
 * (2) should probably be slightly tweaked.  The original intention for the always off trace example was to have an example of when a trace shouldn't be sampled.  This still works as intended, but it's not abundantly clear I think?
* (3) should definitely include a timestamp.  This also shouldn't be a blocker for this PR.

sdk/Trace/SpanOptions.php
```diff
@@ -35,7 +37,14 @@ public function setSpanName(string $name): API\SpanOptions
 
     public function setSpanKind(int $spanKind): API\SpanOptions
     {
-        // TODO: Implement setSpanKind() method.
+        $reflection = new ReflectionClass(API\SpanKind::class);
```
beniamin, line 40: Taking into account that an application may create a lot of spans, this cand have an impact on performance. Usually, the reflection has an impact on performance. I made [a test](https://pastebin.com/FXBN1Baf) comparing the reflection method vs a simple array and the impact in performance is around 2.5x - 3x.

That being said, I recommend replacing the reflection with a simple array(a list of spans kind) and the current kind to be checked on that.

artem-prozorov: Yes, I thought about that but considering that right now there is no const TYPES with an array of available types, I finally decided against messing with the SpanKind interface and save the flexibility by getting available kinds via reflection dynamically.

Do you think it's okay to add const TYPES in the interface? The alternative is to create here a static getter that would get available types via reflection only once? Something like
```
protected static $types = [];

protected static function getAvailableTypes(): array
{
    if (empy(static::$types)) {
        $reflection = new ReflectionClass(API\SpanKind::class);
        static::$types = $reflection->getConstants();
    }

    return static::$types;
}
```
and so the validation would look like `if (!in_array($spanKind, static::getAvailableTypes(), true))`
The advantage of this is that we don't need to care about TYPES consistency and the perfomance impact is much less.

wRLSS: All of these type stuff is implemented really good inside `MyCLabs\Enum`. Maybe we should consider switching to it instead? The why's are covered here:  
https://github.com/myclabs/php-enum#why

beniamin: I like the ideea of having an array of Types on SpanKind API.

@wRLSS the 'php-enum' uses Reflection behind the scene, so is not an improvement on the current implementation.

wRLSS: There are a lot more to it. First of all it has internal cache, also it introduces typed value object which is much safer to pass around.

bobstrecansky:
Thanks for the changes @artem-prozorov!  We merged another PR this morning that created these conflicts.  I'm happy to resolve the conflicts and merge if that's okay with you, or you can also feel free to resolve them.

artem-prozorov:
> Thanks for the changes @artem-prozorov! We merged another PR this morning that created these conflicts. I'm happy to resolve the conflicts and merge if that's okay with you, or you can also feel free to resolve them.

No worries, I'll resolve

Issue #73, opened 2020-04-05
Replace current time implementation with hrtime


bobstrecansky:
We'd like to replace our current time implementation with php's `hrtime` function.

https://www.php.net/manual/en/function.hrtime.php

Todo for this:

- [x] update dependencies to PHP 7.3
- [ ] replace current time implementation with calls to hrtime

bobstrecansky:
Working through this first with https://github.com/open-telemetry/opentelemetry-php/pull/83

bobstrecansky:
This should be a relatively low lift ticket, and would be good for someone who is trying to get acclimated with the codebase.

We went though and replaced string instances of time with integer times.  Plumbing was also created to use a monotonic timer for Span duration, but has not yet been implemented.

bobstrecansky:
This has been completed with #153

PR #153, opened 2020-07-20
timestamp should be span start timestamp

mattiabasone:
As reported here https://zipkin.io/zipkin-api/#/default/post_spans: 'Epoch microseconds of the start of this span, possibly absent if incomplete.'

### DIFF
--- a/contrib/Zipkin/SpanConverter.php
+++ b/contrib/Zipkin/SpanConverter.php
@@ -69,8 +69,9 @@ class SpanConverter
             if (!array_key_exists('annotations', $row)) {
                 $row['annotations'] = [];
             }
+            // Maybe $event->getTimestamp() (if not null)?
             $row['annotations'][] = [
-                'timestamp' => (int) ($timestamp / 1e3), // RealtimeClock in microseconds
+                'timestamp' => (int) ($start_realtime / 1e3), // RealtimeClock in microseconds
                 'value' => $event->getName(),
             ];
         }
--- a/contrib/Zipkin/SpanConverter.php
+++ b/contrib/Zipkin/SpanConverter.php
@@ -42,10 +42,6 @@ class SpanConverter
 
     public function convert(Span $span)
     {
-        $start_realtime = $span->getStartTimestamp();
-        $end_realtime = $span->getEndTimestamp();
-        $elapsed_realtime = (int) (($end_realtime - $start_realtime) / 1e3); // diff in microseconds
-
         $row = [
             'id' => $span->getContext()->getSpanId(),
             'traceId' => $span->getContext()->getTraceId(),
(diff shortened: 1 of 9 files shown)


contrib/Zipkin/SpanConverter.php
```diff
@@ -56,7 +54,7 @@ public function convert(Span $span)
                 'serviceName' => $this->serviceName,
             ],
             'name' => $span->getSpanName(),
-            'timestamp' => (int) ($timestamp / 1e3), // RealtimeClock in microseconds
+            'timestamp' => (int) ($start_realtime / 1e3), // RealtimeClock in microseconds
```
bobstrecansky, line 57: I'm thinking we should probably use the monotonic clock rather than the realtime clock for this one:

https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/Clock.php#L40-L44

This is an artifact left from https://github.com/open-telemetry/opentelemetry-php/issues/73

@mattiabasone - are you interested in taking on that work alongside this PR?

contrib/Zipkin/SpanConverter.php
```diff
@@ -43,7 +42,6 @@ private function sanitiseTagValue($value)
 
     public function convert(Span $span)
     {
-        $timestamp = Clock::get()->timestamp();
         $start_realtime = $span->getStartTimestamp();
```
bobstrecansky, line 45: I'm thinking we should probably use the monotonic clock instead of the realtime clock for this one:

https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/Clock.php#L40-L44

This is an artifact left from https://github.com/open-telemetry/opentelemetry-php/issues/73

@mattiabasone - are you interested in taking on that work alongside this PR?

mattiabasone: @bobstrecansky so these changes don't fix the problem?

Before this "fix" I have some issues with trace representation because root span (last closed) had the greatest timestamp value and jaeger tries to adjust clock skew

Before
tracing_before

After
tracing_after

bobstrecansky: @mattiabasone - I think your fixes do fix the imminent problem.  I think we need to use a monotonic clock in a couple places in these calls as well, as that'll be a more accurate time call.

mattiabasone: @bobstrecansky probably missing something, but if all the monotonic clocks are sent to API, isn't there the risk of having many different timestamps (based on server/container system clocks) if you work with different microservices?

Surely it can be used to measure the total duration of the span, saving in two properties startTime/endTime the value of monotonic clock at two different times.

bobstrecansky: Agree with you that there is a risk of clock skew based on different services. Also agree that the monotonic clock should be used to measure the duration of the span in the way perscribed.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#time

bobstrecansky:
@mattiabasone - this looks like an awesome improvement!  Validating it locally and will be ready to merge after you sign the CLA.

PR #155, opened 2020-07-22
Metrics api

artem-prozorov:
#7 
This PR adds API interfaces, partial implementation for meter provider, global meter provider and abstract exporter


sdk/Metrics/AbstractExporter.php
```diff
@@ -0,0 +1,62 @@
...
+use OpenTelemetry\Sdk\Metrics\Exceptions\RetryableExportException;
+
+abstract class AbstractExporter implements API\Exporter
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function export(iterable $metrics): int
+    {
+        // todo: we need to implement Logger?
```
bobstrecansky, line 19: The logger project will come after the Tracing and Metrics API/SDK pairs are completed.  They are still working through the implementations.  I think this isn't a primary concern right now.

tests/Sdk/Unit/Metrics/Exporters/AbstractExporterTest.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return new class() extends AbstractExporter {
+            protected function getFormatted(API\Metrics $metric)
+            {
+                return $metric;
+            }
+
+            protected function send(array $preparedMetrics): void
+            {
+                // No need to implement here
```
bobstrecansky, line 39: Now or in the future?

artem-prozorov: I meant no need to implement at all. Maybe it's worth removing this comment as it may be confusing

bobstrecansky:
@artem-prozorov - is this PR ready for review?

api/Metrics/UpDownCounter.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Metrics;
+
+interface UpDownCounter extends Counter
+{
```
bobstrecansky, line 8: Should we add `@access` and `@return` doc blocks here?

api/Metrics/Counter.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Metrics;
+
+interface Counter extends Metric
+{
```
bobstrecansky, line 8: Should we add `@access` and `@return` doc blocks here?

api/Metrics/MeterProvider.php
```diff
@@ -0,0 +1,20 @@
...
+interface MeterProvider
+{
+    /**
+     * @param string $name - (required) - This name must identify the instrumentation library
+     * (e.g. io.opentelemetry.contrib.mongodb) and not the instrumented library.
+     * In case an invalid name (null or empty string) is specified, a working default Meter implementation is returned
+     * as a fallback rather than returning null or throwing an exception.
+     * A MeterProvider could also return a no-op Meter here if application owners configure the SDK to suppress
+     * telemetry produced by this library.
+     * @param ?string $version - (optional) - Specifies the version of the instrumentation library (e.g. semver:1.0.0)
```
bobstrecansky, line 16: Should we follow the `@access` and `@return` doc block patterns here too?

docker-compose.prometheus.yaml
```diff
@@ -0,0 +1,33 @@
...
+    volumes:
+      - ./docker/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml
+  redis:
+    image: redis
+    ports:
+        - 6379:6379
+  web:
+    image: nginx:1.17-alpine
+    ports:
+      - 8083:80
```
bobstrecansky, line 16: Did you pick port 8083 for a particular reason?   This isn't a port pattern I'm familiar with.

sdk/Metrics/Exceptions/CantBeExported.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Metrics\Exceptions;
+
+class CantBeExported extends ExportException
```
bobstrecansky, line 7: Could we potentially just have `CantBeExported` extend `Exception` rather than the nested `CantBeExported -> ExportException -> Exception` tree?  Maybe I'm missing the reason you're doing it this way.

Maybe we could have multiple Exception classes defined in one `sdk/Metrics/Exceptions` file rather than in individual files.

sdk/Metrics/Exporters/PrometheusExporter.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->registry = $registry;
+        $this->namespace = $namespace;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function doExport(iterable $metrics): void
+    {
+        foreach ($metrics as $metric) {
```
bobstrecansky, line 35: I think we'll want to keep an eye on this nested switch as the metrics groupings get bigger.

sdk/Metrics/Exceptions/CantBeExported.php
artem-prozorov: Well, we have 2 kind of exceptions: the ones where there is no reason to retry export (wrong data for example) and the ones where we can retry (the server is unreachable for now but we can try later).

The idea was to keep a single type of exceptions (`ExportException`) for all export operations so they can be easily catched and in the same time we could easily differ retryable ones from irretryable ones

docker-compose.prometheus.yaml
artem-prozorov: Hm... no particular reason. Just trying to avoid port collisions... I'll change to 8080

docker-compose.prometheus.yaml
```diff
@@ -0,0 +1,33 @@
...
+    depends_on:
+      - php-prometheus
+    volumes:
+      - ./docker/prometheus/default.conf:/etc/nginx/conf.d/default.conf
+      - ./docker/prometheus/index.php:/var/www/public/index.php
+  php-prometheus:
+    build:
+      context: .
+      dockerfile: docker/prometheus/Dockerfile
+    volumes:
```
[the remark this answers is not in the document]
artem-prozorov: hm... no particular reason, just a habit to mount only the folders that are required for correct functioning. Will change

docker/prometheus/prometheus.yml
```diff
@@ -0,0 +1,32 @@
+global:
```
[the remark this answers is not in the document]
artem-prozorov: Yes, they are the default values

sdk/Metrics/Exceptions/CantBeExported.php
bobstrecansky: Cool sounds good to me

Makefile
```diff
@@ -13,6 +13,11 @@ examples: FORCE
 	$(DC_RUN_PHP) php ./examples/AlwaysOnTraceExample.php
 	$(DC_RUN_PHP) php ./examples/AlwaysOffTraceExample.php
 	$(DC_RUN_PHP) php ./examples/JaegerExporterExample.php
+prometheus-example:
+	@docker-compose -f docker-compose.prometheus.yaml up -d web
```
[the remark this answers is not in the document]
artem-prozorov: Fixed

api/Metrics/Counter.php
artem-prozorov: done

api/Metrics/MeterProvider.php
artem-prozorov: done

api/Metrics/UpDownCounter.php
artem-prozorov: done

docker-compose.prometheus.yaml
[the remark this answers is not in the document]
artem-prozorov: Done

docker/prometheus/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM php:7.3-fpm-alpine
```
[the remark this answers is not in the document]
artem-prozorov: I decided to keep this for now, we can refactor later if needed

sdk/Metrics/AbstractMetric.php
```diff
@@ -0,0 +1,42 @@
...
+     * @var string $name
+     */
+    protected $name;
+
+    /**
+     * @var string $description
+     */
+    protected $description;
+
+    public function __construct(string $name, string $description = '')
```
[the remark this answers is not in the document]
artem-prozorov: I'm not sure about it, I haven't found anything on it in the spec. Just did so in order to make things easier. I suggest keep it for now

sdk/Metrics/Exporters/PrometheusExporter.php
artem-prozorov: Discussed via Zoom

api/Metrics/LabelableMetric.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function getLabels(): array;
+
+    /**
+     * Set $labels
+     *
+     * @param  array  $labels
+     * @return  self
+     */
+    public function setLabels(array $labels);
```
morrisonlevi, line 22: I don't like unstructured arrays like this. We should either:
- Make interfaces that explicitly specify all type info (which is a bit tedious)
- Or add structure to the arrays in the doc comments.

I have a preference for the first one, as not everyone will use static analysis tools and those types will all be enforced at runtime, but may accept the latter.

artem-prozorov: Done.

For now I've chosen the second option but added todo for a later refactoring.
I also added this check https://github.com/open-telemetry/opentelemetry-php/blob/56caa169c9f032695c491f15f00d3ce323128247/sdk/Metrics/HasLabels.php#L29 to make sure `setLabels` receives only arrays of strings

api/Metrics/Metric.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Returns metric's description
+     *
+     * @access	public
+     * @return	string
+     */
+    public function getDescription(): string;
+
+    /**
+     * Returns metric's description
```
beniamin, line 26: This comment is wrong. Maybe a copy/paste without the later update.

composer.json
```diff
@@ -8,7 +8,9 @@
         "ext-json": "*",
         "guzzlehttp/guzzle": "^6.5",
         "psr/http-client": "^1.0",
-        "php-http/guzzle6-adapter": "^2.0"
+        "php-http/guzzle6-adapter": "^2.0",
+        "endclothing/prometheus_client_php": "^1.0",
+        "webmozart/assert": "^1.9"
```
beniamin, line 13: I think that we should keep the library as slim as possible and avoid as much as we can to add extra libraries. I understand the fact that the webmozart/assert library makes the code cleaner, but in the end, this is a library that may be used in a lot of projects and we will force all users to integrate the webmozart/assert in their projects.

As I saw below, the asserts made with webmozart are not very complex, and the code will remain elegant even if you switch to plain PHP.

api/Metrics/Metric.php
bobstrecansky: Agreed, this should be `Returns metric's type` or something like that.

artem-prozorov: Fixed

composer.json
artem-prozorov: Assert library removed

PR #157, opened 2020-07-27
Implement an UpDownCounter Instrument for the Metrics API

zsistla:
* Initial UpDownCounter commit.
* Unit tests included.

`UpDownCounter` extends `AbstractMetric` from the `Metrics api` [pull request](https://github.com/open-telemetry/opentelemetry-php/pull/155/files).

PR #158, opened 2020-07-27
Implement a ValueRecorder Instrument for the Metrics API

zsistla:
* Unit tests included.
* Initial ValueRecorder commit.

`ValueRecorder` depends on and extends `AbstractMetric` from the `Metrics api` [pull request](https://github.com/open-telemetry/opentelemetry-php/pull/155/files).

### DIFF
--- a/api/Metrics/ValueRecorder.php
+++ b/api/Metrics/ValueRecorder.php
@@ -7,7 +7,7 @@ namespace OpenTelemetry\Metrics;
 interface ValueRecorder
 {
     /**
-     * record.
+     * records the given value to this ValueRecorder.
      *
      * @access	public
      * @param int|float $value
--- a/api/Metrics/ValueRecorder.php
+++ b/api/Metrics/ValueRecorder.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Metrics;
 
-interface ValueRecorder
+interface ValueRecorder extends Metric
 {
     /**
      * records the given value to this ValueRecorder.
@@ -14,4 +14,36 @@ interface ValueRecorder
      * @return void
      */
     public function record($value) : void;
+
+    /**
+     * Returns the sum of the values
+     *
+     * @access	public
+     * @return	float
+     */
+    public function getSum(): float;
+
+    /**
+     * Returns the min of the values
+     *
+     * @access	public
+     * @return	float
+     */
+    public function getMin(): float;
+
+    /**
+     * Returns the max of the values
+     *
+     * @access	public
+     * @return	float
+     */
+    public function getMax(): float;
+
+    /**
+     * Returns the count of the values
+     *
+     * @access	public
+     * @return	int
+     */
+    public function getCount(): int;
 }
--- a/sdk/Metrics/ValueRecorder.php
+++ b/sdk/Metrics/ValueRecorder.php
@@ -1,0 +1,164 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Metrics;
+
+use InvalidArgumentException;
+use OpenTelemetry\Metrics as API;
+
+/*
+ * Name	: ValueRecorder
+ * Instrument kind : Synchronous
+ * Function(argument) : Record(value)
+ * Default aggregation : MinMaxSumCount
+ * Notes : Per-request, any non-additive measurement
+ *
+ * ValueRecorder is a non-additive synchronous instrument useful for
+ * recording any non-additive number, positive or negative. Values
+ * captured by a Record(value) are treated as individual events
+ * belonging to a distribution that is being summarized. ValueRecorder
+ * should be chosen either when capturing measurements that do not
+ * contribute meaningfully to a sum, or when capturing numbers that
+ * are additive in nature, but where the distribution of individual
+ * increments is considered interesting.
+ *
+ * Example: One of the most common uses for ValueRecorder is to capture latency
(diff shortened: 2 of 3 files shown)


bobstrecansky:
This PR has the same phan issue that https://github.com/open-telemetry/opentelemetry-php/pull/157 has.

zsistla:
Hi Bob,

Thank you for checking out the PR.  

The phan errors are related to this not being a standalone PR but dependent on Artem's PR.  The classes it's not detecting are from his PR.

sdk/Metrics/UpDownCounter.php:23 PhanUndeclaredExtendedClass Class extends undeclared class \OpenTelemetry\Sdk\Metrics\AbstractMetric
36
sdk/Metrics/UpDownCounter.php:23 PhanUndeclaredInterface Class implements undeclared interface \OpenTelemetry\Metrics\LabelableMetric
37
sdk/Metrics/UpDownCounter.php:23 PhanUndeclaredInterface Class implements undeclared interface \OpenTelemetry\Metrics\UpDownCounter
38
sdk/Metrics/UpDownCounter.php:25 PhanUndeclaredTrait Class uses undeclared trait \OpenTelemetry\Sdk\Metrics\HasLabels
39
sdk/Metrics/UpDownCounter.php:39 PhanUndeclaredClassConstant Reference to constant UP_DOWN_COUNTER from undeclared class \OpenTelemetry\Metrics\MetricKind

The errors are such that I think when Artem's PR is merged, the phan checks will pass.

lalex:
Probably wrong file is committed to this PR, `sdk/Metrics/UpDownCounter.php` instead of `sdk/Metrics/ValueRecorder.php`.

zsistla:
> Probably wrong file is committed to this PR, `sdk/Metrics/UpDownCounter.php` instead of `sdk/Metrics/ValueRecorder.php`.

:) Great catch, thanks!!

bobstrecansky:
I think this is ready to go after the merge conflicts are fixed!

sdk/Metrics/ValueRecorder.php
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Updates the ValueRecorder's value with the specified value then returns
+     * the current value count.
+     *
+     * @access	public
+     *
+     * @param float $value, accepts INTs or FLOATs. If value is an int, it it cast as a float.
+     *
+     * @return int $valueCount, returns the current count of values recorded.
```
artem-prozorov, line 136: Looks like there is an inconsistency here because the method `record` doesn't actually return anything

sdk/Metrics/ValueRecorder.php
```diff
@@ -0,0 +1,161 @@
...
+        return API\MetricKind::VALUE_RECORDER;
+    }
+
+    /**
+     * Returns the sum of the values
+     *
+     * @access	public
+     * @return	float
+     */
+    public function getValueSum(): float
```
artem-prozorov, line 75: Should we add all these public functions to the interface?

tests/Sdk/Unit/Metrics/ValueRecorderTest.php
```diff
@@ -0,0 +1,137 @@
...
+        $this->assertEquals(5, $metric->getValueMax());
+        $this->assertEquals(5, $metric->getValueMin());
+        $this->assertEquals(5, $metric->getValueSum());
+        $metric->record(2);
+        $this->assertEquals(2, $metric->getValueCount());
+        $this->assertEquals(5, $metric->getValueMax());
+        $this->assertEquals(2, $metric->getValueMin());
+        $this->assertEquals(7, $metric->getValueSum());
+    }
+    public function testValidNegativeIntRecord()
```
artem-prozorov, line 30: I would suggest to add empty lines to separate tests

tests/Sdk/Unit/Metrics/ValueRecorderTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tests\Sdk\Unit\Metrics;
+
+use InvalidArgumentException;
+use OpenTelemetry\Sdk\Metrics\ValueRecorder;
+use PHPUnit\Framework\TestCase;
+
+class ValueRecorderTest extends TestCase
```
artem-prozorov, line 11: Looks like all tests except the last one follow the same pattern. I would suggest to move data out to a `@dataProvider`. This would let you have only one test with a bunch of test data instead of 8 similar tests

api/Metrics/ValueRecorder.php
```diff
@@ -7,7 +7,7 @@
 interface ValueRecorder
```
artem-prozorov, line 7: I suggest we extend base Metrics interface ( https://github.com/open-telemetry/opentelemetry-php/blob/789819c9c146fca4989e618948183dcf7448ebfb/api/Metrics/Metric.php ) here in this PR

zsistla: Great catch, thank you.

sdk/Metrics/ValueRecorder.php
zsistla: Thank you for reviewing.

I initially left it out as I felt the `API` should mirror the spec and it wasn't covered explicitly in the spec, but looking again, I see:
```
 * Function(argument) : record(value)
 * Default aggregation : MinMaxSumCount
```
Since it calls out the aggregations, I've included functions in the API as there must be a way to access those values.
I didn't include getMean, as that is a nice to have helper function, but isn't strictly needed and isn't called out in the spec as an aggregate either.

PR #161, opened 2020-08-11
Phpunit update

artem-prozorov:
Two major versions of PHPUnit have been released, I think it's tiime to update :)

- Changed CI settings according to shivammathur's recomendations and in order to make PhpUnit 9 run
- Changed test that used some deprecated and removed method


tests/Sdk/Integration/Context/SpanContextTest.php
```diff
@@ -19,7 +19,7 @@ class SpanContextTest extends TestCase
     public function testInvalidSpan(string $traceID, string $spanID, string $errorRegex): void
     {
         $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessageRegExp($errorRegex);
+        $this->expectExceptionMessageMatches($errorRegex);
```
bobstrecansky, line 22: When I left this as the previous value, all of our unit tests pass.  Was this intended to be changed to test PHPUnit, or something else?

For reference, the errors I was getting with `expectExceptionMessageMatches` are as follows:
https://gist.github.com/bobstrecansky/ca3eb2dac6eb37448ac1c9b73733e3a6

artem-prozorov: expectExceptionMessageRegExp has been deprecated and removed from phpunit 9 so I replaced it with a newer one that is acualy the same. Github action fails and that is wierd, there may be some cache. I drafted this PR for now until I find out the reason

bobstrecansky: after I ran a `make update`, my tests passed.  Perhaps it's a cached version of our build image?

bobstrecansky: Rerunning the CI suite to validate.

bobstrecansky: Aha!  The image that we are using uses the `shivammathur/setup-php` github action helper.  We can explicitly call out the version of phpunit that we want to use there; this is most likely the conflict.

https://github.com/shivammathur/setup-php#wrench-tools-support

We would probably just need to define that in our actions workflow:
https://github.com/open-telemetry/opentelemetry-php/blob/master/.github/workflows/php.yml#L27

shivammathur:
@artem-prozorov
Some notes
- As you have `phan` and `phpunit` in your `composer.json`, you do not need to include them in tools input in `setup-php`.
- `pecl` input is not supported on `setup-php` in `v2` and is not required in your workflow, so you can remove it.
- `Cache Composer packages` step should be before `Install dependencies` step. The same step restores cache and adds a step at the end of workflow to save the cache.
- Run `phpunit --migrate-configuration` with `phpunit 9.3` locally to migrate your `phpunit.xml.dist` to new schema.

tests/Sdk/Integration/Context/SpanContextTest.php
artem-prozorov: Is there any particular reason why composer.lock is gitignored?

bobstrecansky: @nenad removed it here: https://github.com/open-telemetry/opentelemetry-php/pull/51

.github/workflows/php.yml
```diff
@@ -24,25 +24,25 @@ jobs:
...
-    - name: Install dependencies
-      if: steps.composer-cache.outputs.cache-hit != 'true'
-      run: composer install --prefer-dist --no-progress --no-suggest
-            
     - name: Cache Composer packages
       id: composer-cache
       uses: actions/cache@v2
       with:
         path: vendor
-        key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
```
artem-prozorov, line 42: Since this step must be run before the composer packages installation, there is no composer.lock at this point, so I changed to composer.json (according to the docs https://github.com/shivammathur/setup-php#cache-composer-dependencies )

.github/workflows/php.yml
```diff
@@ -24,25 +24,25 @@ jobs:
       with:
         php-version: ${{ matrix.php-versions }}
         coverage: xdebug
-        tools: php-cs-fixer, phpunit, phan 
-        pecl: true
+        tools: php-cs-fixer
+        extensions: ast
```
artem-prozorov, line 28: PHP-ACT is required for phan for some reasons

phpunit.xml.dist
```diff
@@ -32,14 +51,5 @@
...
         </testsuite>
     </testsuites>
-    <filter>
-        <whitelist>
-            <directory>contrib</directory>
-            <directory>sdk</directory>
-        </whitelist>
-    </filter>
 
-    <logging>
```
artem-prozorov, line 42: For some reasons PhpUnit 9 generated an error that said that "<log>" element is not expected. The docs says it must be okay. I googled it for a while but havenn't found anything on it. I dont' feel like going deep into this, instead I just moved logging config to the CLI option. Looks like its fine now

bobstrecansky: Sounds good

PR #163, opened 2020-08-14
Storage interface

artem-prozorov:
In order to persist Metrics' data we need to use some external storage. This can be a file-based storage or Redis or Memcached or just a user-provided storage that would implement the Storage interface.
In this PR I'm creating the Storage interface and a base exception for all storage-related exceptions

This PR is based on this branch: https://github.com/open-telemetry/opentelemetry-php/pull/161


api/Metrics/Storage.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Deletes the provided metrics from storage
+     *
+     * @access	public
+     * @throws  \OpenTelemetry\Sdk\Metrics\Exceptions\StorageException
+     * @param	iterable<Metric|string>	$metrics
+     * @return	void
+     */
+    public function deleteMany(iterable $metrics): void
```
bobstrecansky, line 65: You're missing a trailing `;` here.

Issue #113, opened 2020-06-05
Implement a Meter Interface


bobstrecansky:
We need to implement a meter interface, as defined in: 
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#meter-interface

Success criteria: Creating a Meter Interface in the API.

Fahmy-Mohammed:
I've started work on this.

artem-prozorov:
@Fahmy-Mohammed Hi, any news on this issue?

bobstrecansky:
Talked to Amber in our SIG meeting today; Fahmy is on vacation this week but is almost completed with the work on this PR.  Amber said to expect some information next week.

bobstrecansky:
@artem-prozorov made a stub for this [here](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Metrics/Meter.php) and @Fahmy-Mohammed mentioned his PR was almost ready for review.

Fahmy-Mohammed:
Apologies for the delay. PR is ready for a look 👍🏾

bobstrecansky:
Fahmy completed this work in https://github.com/open-telemetry/opentelemetry-php/pull/165, closing this ticket.

PR #165, opened 2020-08-18
Meter interface

Fahmy-Mohammed:
This PR adds a [`Meter Interface`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#meter-interface) that creates the `Counter`,` UpDownCounter`, and `ValueRecoder` [metric instruments](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#the-instruments) (the other metrics instruments will be added once their implementation is complete).

#113

### DIFF
--- a/api/Metrics/Meter.php
+++ b/api/Metrics/Meter.php
@@ -16,15 +16,15 @@ interface Meter
     /**
      * Returns the meter version.
      *
-     * @return string
+     * @return string (optional) - Metric version
      */
     public function getVersion(): ?string;
 
     /**
      * Creates a Counter metric instrument.
      *
-     * @param string $name        - (required) - Counter name
-     * @param string $description - (optional) - Counter description
+     * @param string $name        (required) - Counter name
+     * @param string $description (optional) - Counter description
      *
      * @return Counter
      */
@@ -33,8 +33,8 @@ interface Meter
     /**
      * Creates an UpDownCounter metric instrument.
      *
-     * @param string $name        - (required) - UpDownCounter name
-     * @param string $description - (optional) - UpDownCounter description
+     * @param string $name        (required) - UpDownCounter name
+     * @param string $description (optional) - UpDownCounter description
      *
      * @return UpDownCounter
      */
@@ -43,8 +43,8 @@ interface Meter
     /**
      * Creates a ValueRecorder metric instrument.
      *
-     * @param string $name        - (required) - ValueRecorder name
-     * @param string $description - (optional) - ValueRecorder description
+     * @param string $name        (required) - ValueRecorder name
+     * @param string $description (optional) - ValueRecorder description
      *
      * @return ValueRecorder
      */
--- a/api/Metrics/Meter.php
+++ b/api/Metrics/Meter.php
@@ -6,4 +6,47 @@ namespace OpenTelemetry\Metrics;
 
 interface Meter
 {
+    /**
+     * Returns the meter name.
+     *
+     * @return string
+     */
+    public function getName(): string;
+
+    /**
+     * Returns the meter version.
+     *
+     * @return string
+     */
+    public function getVersion(): ?string;
+
+    /**
+     * Creates a Counter metric instrument.
+     *
+     * @param string $name        - (required) - Counter name
+     * @param string $description - (optional) - Counter description
+     *
+     * @return Counter
+     */
+    public function newCounter(string $name, string $description): Counter;
+
+    /**
+     * Creates an UpDownCounter metric instrument.
+     *
+     * @param string $name        - (required) - UpDownCounter name
+     * @param string $description - (optional) - UpDownCounter description
+     *
+     * @return UpDownCounter
+     */
+    public function newUpDownCounter(string $name, string $description): UpDownCounter;
+
+    /**
+     * Creates a ValueRecorder metric instrument.
+     *
+     * @param string $name        - (required) - ValueRecorder name
+     * @param string $description - (optional) - ValueRecorder description
+     *
+     * @return ValueRecorder
+     */
+    public function newValueRecorder(string $name, string $description): ValueRecorder;
 }
--- a/api/Metrics/Meter.php
+++ b/api/Metrics/Meter.php
@@ -16,9 +16,9 @@ interface Meter
     /**
      * Returns the meter version.
      *
-     * @return string (optional) - Metric version
+     * @return string Metric version
      */
-    public function getVersion(): ?string;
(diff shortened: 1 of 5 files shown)


api/Metrics/Meter.php
```diff
@@ -6,4 +6,47 @@
...
+     * Returns the meter name.
+     *
+     * @return string
+     */
+    public function getName(): string;
+
+    /**
+     * Returns the meter version.
+     *
+     * @return string
```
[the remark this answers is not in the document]
Fahmy-Mohammed: Good call! I've added the "optional" in the comment.

sdk/Metrics/Meter.php
```diff
@@ -8,5 +8,68 @@
...
+
+        return $this->counter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function newUpDownCounter(string $name, string $description = null): API\UpDownCounter
+    {
+        $this->upDownCounter = new UpDownCounter($name, $description);
```
artem-prozorov, line 61: The same as with counter

sdk/Metrics/Meter.php
```diff
@@ -8,5 +8,68 @@
...
+
+        return $this->upDownCounter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function newValueRecorder(string $name, string $description = null): API\ValueRecorder
+    {
+        $this->valueRecorder = new ValueRecorder($name, $description);
```
artem-prozorov, line 71: The same as with counter

tests/Sdk/Unit/Metrics/MeterTest.php
```diff
@@ -0,0 +1,35 @@
...
+use OpenTelemetry\Sdk\Metrics\UpDownCounter;
+use OpenTelemetry\Sdk\Metrics\ValueRecorder;
+use PHPUnit\Framework\TestCase;
+
+class MeterTest extends TestCase
+{
+    public function testMeterInstrumentCreation()
+    {
+        $meter = new Meter('Meter', '0.1');
+        $this->assertInstanceOf(Meter::class, $meter);
```
artem-prozorov, line 18: This assertion is pointless. The $meter is created explicitly with `new`, there is absolutely no possibility that there may be anything other then `Meter` class

tests/Sdk/Unit/Metrics/MeterTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+        $meter_name = $meter->getName();
```
artem-prozorov, line 20: Looks like the codestyle is more towards camelCase other than to snake_case, no?

tests/Sdk/Unit/Metrics/MeterTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->assertSame('Meter', $meter_name);
```
artem-prozorov, line 21: well actually I don't think you need `$meter_name` variable at all. IMHO it's more readable to call `getName()` right in the assertion so you see exaclty what is being compared

tests/Sdk/Unit/Metrics/MeterTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+        $meter_version = $meter->getVersion();
+        $this->assertSame('0.1', $meter_version);
+
+        $counter = $meter->newCounter('Counter', 'A counter');
+        $this->assertInstanceOf(Counter::class, $counter);
```
artem-prozorov, line 27: I would also suggest you test that metric's name and description are passed to the `$counter` correctly
```
$counterName = 'Counter';
$counterDescription = 'A counter';
$counter = $meter->newCounter($counterName, $counterDescription);
$this->assertInstanceOf(Counter::class, $counter);
$this->assertEquals($counterName, $counter->getName());
$this->assertEquals($counterDescription, $counter->getDescription());
```

artem-prozorov: I would also recommend brake this one huge test into 4 different tests: one for checking Meter name and version, the other thee - one for every metric type (test a metric is created and name and description are passed correctly)

tests/Sdk/Unit/Metrics/MeterTest.php
bobstrecansky: Yeah we should probably make this `meterName`

sdk/Metrics/Meter.php
```diff
@@ -8,5 +8,68 @@
...
+    private $name;
+    private $version;
+    private $counter;
+    private $upDownCounter;
+    private $valueRecorder;
+
+    public function __construct(string $name, ?string $version = null)
+    {
+        $this->name = $name;
+        $this->version = $version !== null ? $version : '';
```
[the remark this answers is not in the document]
bobstrecansky: That's slightly more readable maybe?

sdk/Metrics/Meter.php
```diff
@@ -8,5 +8,68 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function newCounter(string $name, string $description = null): API\Counter
+    {
+        $this->counter = new Counter($name, $description);
+
+        return $this->counter;
```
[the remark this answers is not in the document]
Fahmy-Mohammed: Yes, you're right here. Keeping the `Meter` class stateless is the way to go.

tests/Sdk/Unit/Metrics/MeterTest.php
Fahmy-Mohammed: Yes, that's a bit cleaner!

tests/Sdk/Unit/Metrics/MeterTest.php
Fahmy-Mohammed: Good call! I've split those out into their own tests.

sdk/Metrics/Meter.php
```diff
@@ -8,5 +8,56 @@
 
 class Meter implements API\Meter
 {
-    // Temp stub
+    /**
+     * @var string $name
+     * @var string $version
+     */
+    private $name;
+    private $version;
```
artem-prozorov, line 16: No big deal but I prefer to use protected properties instead of private.
The reason behind it is that a developer may need to extend Meter's functinality. In order to do that he will extend this class. The extended class will miss these 2 properties. But actually these properties will be dynamically created in runtime. This MAY lead to some unexpected things

artem-prozorov (approved):

PR #170, opened 2020-08-30
Export SpanStatus

beniamin:
I implemented SpanStatus exporting in ZipkinExporter and fixed some deprecation in unit tests.

Because the StatusCode needs to be exported as Code Name I had to change SpanStatus API and some other implementation in the SDK and convert the code int to string. Here are the specs [Status](https://github.com/open-telemetry/opentelemetry-specification/blob/e0bd41739a92304c0be5d68840f434e281b5f8b2/specification/trace/sdk_exporters/zipkin.md#status).

I am aware that this change may impact the memory footprint and I am open to new solutions.

### DIFF
--- a/api/Trace/SpanStatus.php
+++ b/api/Trace/SpanStatus.php
@@ -6,45 +6,45 @@ namespace OpenTelemetry\Trace;
 
 interface SpanStatus
 {
-    const OK = 0;
-    const CANCELLED = 1;
-    const UNKNOWN = 2;
-    const INVALID_ARGUMENT = 3;
-    const DEADLINE_EXCEEDED = 4;
-    const NOT_FOUND = 5;
-    const ALREADY_EXISTS = 6;
-    const PERMISSION_DENIED = 7;
-    const RESOURCE_EXHAUSTED = 8;
-    const FAILED_PRECONDITION = 9;
-    const ABORTED = 10;
-    const OUT_OF_RANGE = 11;
-    const UNIMPLEMENTED = 12;
-    const INTERNAL = 13;
-    const UNAVAILABLE = 14;
-    const DATA_LOSS = 15;
-    const UNAUTHENTICATED = 16;
+    const OK = 'Ok';
+    const CANCELLED = 'Cancelled';
+    const UNKNOWN = 'Unknown';
+    const INVALID_ARGUMENT = 'InvalidArgument';
+    const DEADLINE_EXCEEDED = 'DeadlineExceeded';
+    const NOT_FOUND = 'NotFound';
+    const ALREADY_EXISTS = 'AlreadyExists';
+    const PERMISSION_DENIED = 'PermissionDenied';
+    const RESOURCE_EXHAUSTED = 'ResourceExhausted';
+    const FAILED_PRECONDITION = 'FailedPrecondition';
+    const ABORTED = 'Aborted';
+    const OUT_OF_RANGE = 'OutOfRange';
+    const UNIMPLEMENTED = 'Unimplemented';
+    const INTERNAL = 'Internal';
+    const UNAVAILABLE = 'Unavailable';
+    const DATA_LOSS = 'DataLoss';
+    const UNAUTHENTICATED = 'Unauthenticated';
 
     const DESCRIPTION = [
-        0 => 'Not an error; returned on success.',
-        1 => 'The operation was cancelled, typically by the caller.',
-        2 => 'Unknown error. For example, this error may be returned when a Status value received from another address space belongs to an error space that is not known in this address space. Also errors raised by APIs that do not return enough error information may be converted to this error.',
-        3 => 'The client specified an invalid argument. Note that this differs from FAILED_PRECONDITION. INVALID_ARGUMENT indicates arguments that are problematic regardless of the state of the system (e.g., a malformed file name).',
-        4 => 'The deadline expired before the operation could complete. For operations that change the state of the system, this error may be returned even if the operation has completed successfully. For example, a successful response from a server could have been delayed long',
-        5 => 'Some requested entity (e.g., file or directory) was not found. Note to server developers: if a request is denied for an entire class of users, such as gradual feature rollout or undocumented allowlist, NOT_FOUND may be used. If a request is denied for some users within a class of users, such as user-based access control, PERMISSION_DENIED must be used.',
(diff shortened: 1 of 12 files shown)


contrib/Zipkin/SpanConverter.php
```diff
@@ -51,13 +53,14 @@ public function convert(Span $span)
...
-            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds
+            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds,
+            'tags' => [
+                self::STATUS_CODE_TAG_KEY => $span->getStatus()->getCanonicalStatusCode(),
+                self::STATUS_DESCRIPTION_TAG_KEY => $span->getStatus()->getStatusDescription()
+            ]
         ];
 
         foreach ($span->getAttributes() as $k => $v) {
-            if (!array_key_exists('tags', $row)) {
```
beniamin, line 58: since the status will be always exported, there is no need for this check.

sdk/Trace/Span.php
```diff
@@ -16,10 +16,8 @@ class Span implements API\Span
     private $startEpochTimestamp;
     private $start;
     private $end;
-    private $statusCode = API\SpanStatus::OK;
```
beniamin, line 19: These attributes are initialized in the constructor, so there is no need to do it here.

contrib/Zipkin/SpanConverter.php
```diff
@@ -51,13 +53,14 @@ public function convert(Span $span)
             ],
             'name' => $span->getSpanName(),
             'timestamp' => (int) ($span->getStartEpochTimestamp() / 1e3), // RealtimeClock in microseconds
-            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds
+            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds,
```
bobstrecansky, line 56: Was this trailing comma added intentionally?

contrib/Zipkin/SpanConverter.php
```diff
@@ -51,13 +53,14 @@ public function convert(Span $span)
...
+            'tags' => [
```
bobstrecansky, line 57: It feels itchy that we have to make two separate function calls for the status code and description when they are so tightly coupled together.

api/Trace/SpanStatus.php
```diff
@@ -6,45 +6,45 @@
...
-    const PERMISSION_DENIED = 7;
-    const RESOURCE_EXHAUSTED = 8;
-    const FAILED_PRECONDITION = 9;
-    const ABORTED = 10;
-    const OUT_OF_RANGE = 11;
-    const UNIMPLEMENTED = 12;
-    const INTERNAL = 13;
-    const UNAVAILABLE = 14;
-    const DATA_LOSS = 15;
-    const UNAUTHENTICATED = 16;
```
bobstrecansky, line 25: If we're concerned about memory footprint, we could always keep these constant numbers and add another array similar to `DESCRIPTION []` but for the status title?  I don't know what the real world implication of changing this from in to string at scale is going to be.

beniamin: I want to make a memory footprint test to measure the impact.

On the other hand, the specs say that one should be able to pass its own Code and Description and that this list is not exhaustive. If we decide to keep the integer codes then we have to allow 2 types of code, int for current defined codes and strings for custom codes.

[StatusCanonicalCodes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#statuscanonicalcode) points to [Standard GRPC Codes](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) which says: 

> The gRPC client and server-side implementations may also generate and return status on their own when errors happen. Only a subset of the pre-defined status codes are generated by the gRPC libraries. This allows applications to be sure that any other code it sees was actually returned by the application (although it is also possible for the server-side to return one of the codes generated by the gRPC libraries).

contrib/Zipkin/SpanConverter.php
beniamin: no. I will remove it

contrib/Zipkin/SpanConverter.php
beniamin: What do you think we can do about this?

api/Trace/SpanStatus.php
beniamin: I just finished my tests and there is no impact on memory footprint.

Test: 
```php
for ($i = 0; $i < 10000000; $i++) {
    $span = $tracer->startAndActivateSpan('span' . $i);
    $status = $statusCodes[array_rand($statusCodes)];

    $span->setSpanStatus(
        constant("OpenTelemetry\Trace\SpanStatus::$status"),
        SpanStatus::DESCRIPTION[constant("OpenTelemetry\Trace\SpanStatus::$status")]
    );

    if ($i % 1000000 == 0) {
        echo PHP_EOL . "Memory usage: ". (memory_get_usage() / 1024 / 1024) . " MB" ;
    }
}
```

status stored as integers:
```
Memory usage: 1.6644973754883 MB
Memory usage: 869.70763397217 MB
Memory usage: 1737.7589035034 MB
Memory usage: 2605.8101730347 MB
Memory usage: 3473.8614425659 MB
```
status stored as string:
```
Memory usage: 1.6651077270508 MB
Memory usage: 869.70824432373 MB
Memory usage: 1737.759513855 MB
Memory usage: 2605.8107833862 MB
Memory usage: 3473.8620529175 MB
```

I think we can stores statuses as strings.

contrib/Zipkin/SpanConverter.php
beniamin: I noticed that on Span interface we have the `getStatus` method, but also `getCanonicalStatusCode` and `getStatusDescription` because Span interface extends SpanStatus interface. This is a little bit odd.

bobstrecansky: Yeah that is a bit odd to me too. Not saying this should dn a blocker, just felt unusual to me.  Easy to change down the road should we choose to do so.

bobstrecansky:
There’s just the one merge conflict. I’m happy to approve and merge once that’s fixed.

bobstrecansky (approved):

PR #171, opened 2020-08-30
remove global span

beniamin:
I removed the need for a global span creation when the Tracer is initialized because of 2 reasons:
- it was confusing and counter-intuitive two have more spans than one request
- Zipkin and Jaeger always reported that is unable to find the parent of the first span because the global span was never exported.

Also, I fixed some issues on BatchExporter:
 - the export in batches was not working properly. The spans were always exported one by one.
 - because there was no last export after the first span processing, the span was exported alone as one batch.
 - fix some issues when computing if enough time has passed because the delay is passed as milliseconds but the clock returns nanoseconds.

Later Edit:
After discussion on Gitter channel we decided that we should use a NoopSpan as a root span, instead of no span at all or a normal Span.

To Do:
- [x] Implement new NooSpan class
- [x] Use NoopSpan as a root span for new traces
- [x] Fix unit tests
- [x] Test the export to Zipkin and Jaeger

### DIFF
--- a/sdk/Trace/SpanOptions.php
+++ b/sdk/Trace/SpanOptions.php
@@ -92,7 +92,7 @@ final class SpanOptions implements API\SpanOptions
     public function toSpan(): API\Span
     {
         $span = $this->tracer->getActiveSpan();
-        $context = $span->getContext()->IsValidContext()
+        $context = $span && $span->getContext()->IsValidContext()
             ? SpanContext::fork($span->getContext()->getTraceId())
             : SpanContext::generate();
 
--- a/sdk/Trace/SpanOptions.php
+++ b/sdk/Trace/SpanOptions.php
@@ -92,7 +92,7 @@ final class SpanOptions implements API\SpanOptions
     public function toSpan(): API\Span
     {
         $span = $this->tracer->getActiveSpan();
-        $context = $span && $span->getContext()->IsValidContext()
+        $context = $span->getContext()->IsValidContext()
             ? SpanContext::fork($span->getContext()->getTraceId())
             : SpanContext::generate();
 
--- a/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -41,7 +41,7 @@ class BatchSpanProcessor implements SpanProcessor
     /**
      * @var int
      */
-    private $lastExportTimestamp = 0;
+    private $lastExportTimestamp;
     /**
      * @var Clock
      */
@@ -107,7 +107,7 @@ class BatchSpanProcessor implements SpanProcessor
      */
     public function forceFlush(): void
     {
-        if (null !== $this->exporter && $this->running) {
+        if (null !== $this->exporter) {
             $this->exporter->export($this->queue);
             $this->queue = [];
             $this->lastExportTimestamp = $this->clock->timestamp();
@@ -126,7 +126,14 @@ class BatchSpanProcessor implements SpanProcessor
 
     protected function enoughTimeHasPassed(): bool
     {
-        $now = $this->clock->timestamp();
+        $now = (int) ($this->clock->timestamp() / 1e6);
+
+        // if lastExport never occurred let it start from now on
+        if (null === $this->lastExportTimestamp) {
+            $this->lastExportTimestamp = $now;
+
+            return false;
+        }
 
         return $this->scheduledDelayMillis < ($now - $this->lastExportTimestamp);
     }
--- a/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/sdk/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -126,7 +126,7 @@ class BatchSpanProcessor implements SpanProcessor
 
     protected function enoughTimeHasPassed(): bool
     {
-        $now = (int) ($this->clock->timestamp() / 1e6);
+        $now = (int) ($this->clock->now() / 1e6);
 
         // if lastExport never occurred let it start from now on
         if (null === $this->lastExportTimestamp) {
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -130,8 +130,6 @@ class Tracer implements API\Tracer
             ? $this->getActiveSpan()->getContext()
             : SpanContext::generate(true);
 
-        $this->setActiveSpan($parent);
-
         $context = SpanContext::fork($parent->getTraceId(), $parent->isSampled());
         $span = $this->generateSpanInstance($name, $context);
 
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -24,6 +24,7 @@ class Tracer implements API\Tracer
     private $resource;
 
     private $rootContext;
+    private $rootSpan;
 
     public function __construct(
         TracerProvider $provider,
@@ -32,13 +33,15 @@ class Tracer implements API\Tracer
     ) {
         $this->provider = $provider;
         $this->resource = $resource;
+        $this->rootSpan = new NoopSpan();
+        $this->setActiveSpan($this->rootSpan);
         $this->rootContext = $context ? $context : SpanContext::generate(true);
     }
 
     /**
      * @return Span
      */
-    public function getActiveSpan(): ?API\Span
+    public function getActiveSpan(): API\Span
     {
         while (count($this->tail) && $this->active->getEnd()) {
             $this->active = array_pop($this->tail);
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -130,6 +130,8 @@ class Tracer implements API\Tracer
             ? $this->getActiveSpan()->getContext()
             : SpanContext::generate(true);
 
+        $this->setActiveSpan($parent);
+
         $context = SpanContext::fork($parent->getTraceId(), $parent->isSampled());
         $span = $this->generateSpanInstance($name, $context);
 
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -13,13 +13,11 @@ class Tracer implements API\Tracer
     private $spans = [];
     private $tail = [];
 
-    /**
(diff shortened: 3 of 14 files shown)


sdk/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -126,7 +126,14 @@ protected function queueReachedLimit(): bool
 
     protected function enoughTimeHasPassed(): bool
     {
-        $now = $this->clock->timestamp();
+        $now = (int) ($this->clock->timestamp() / 1e6);
+
+        // if lastExport never occurred let it start from now on
+        if (null === $this->lastExportTimestamp) {
```
beniamin, line 132: added this to avoid exporting the first span as a separate batch.

sdk/Trace/Tracer.php
```diff
@@ -23,23 +23,22 @@ class Tracer implements API\Tracer
...
         ResourceInfo $resource,
         API\SpanContext $context = null
     ) {
         $this->provider = $provider;
         $this->resource = $resource;
-        $context = $context ?: SpanContext::generate();
-
-        // todo: hold up, why do we automatically make a root Span?
-        $this->active = $this->generateSpanInstance('tracer', $context);
+        $this->rootContext = $context ? $context : SpanContext::generate(true);
```
beniamin, line 35: no need to initialize a new Span, just a new Context if none has been provided.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -126,7 +126,14 @@ protected function queueReachedLimit(): bool
 
     protected function enoughTimeHasPassed(): bool
     {
-        $now = $this->clock->timestamp();
+        $now = (int) ($this->clock->timestamp() / 1e6);
```
beniamin, line 129: delay is in milliseconds and the current must be converted in milliseconds to.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -107,7 +107,7 @@ public function onEnd(API\Span $span): void
      */
     public function forceFlush(): void
     {
-        if (null !== $this->exporter && $this->running) {
```
beniamin, line 110: there is no need to check here if the processor is running because no span will be recorded in onEnd method if running is set to false.

sdk/Trace/SpanOptions.php
```diff
@@ -92,7 +92,7 @@ public function addStart(int $now): API\SpanOptions
     public function toSpan(): API\Span
     {
         $span = $this->tracer->getActiveSpan();
-        $context = $span->getContext()->IsValidContext()
+        $context = $span && $span->getContext()->IsValidContext()
```
bobstrecansky, line 95: Do we need to check for a span and then check to make sure it has a valid context?  Could `$span->getContext()->IsValidContext()` be sufficient here?

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
bobstrecansky: I'm wondering if we should add a helper for this conversion in 
https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/Clock.php

Not really anything important.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
bobstrecansky: I'm used to this kind of comparison check being the other way around:

`if ($this->lastExportTimestamp === null) {`

This is totally fine too; just mentioning it felt strange to me.

tests/Sdk/Unit/Support/HasTraceProvider.php
```diff
@@ -4,8 +4,8 @@
 
 namespace OpenTelemetry\Tests\Sdk\Unit\Support;
 
-use OpenTelemetry\Sdk\Trace\TracerProvider;
 use OpenTelemetry\Sdk\Trace\Tracer;
+use OpenTelemetry\Sdk\Trace\TracerProvider;
```
bobstrecansky, line 8: I'm assuming this was just an editor preference swapping these two locations?

sdk/Trace/SpanOptions.php
beniamin: yes, because I changed getActiveSpan and it may return null, now.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
beniamin: I will think of it, and I will open a different PR for that.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
beniamin: I do this usually to avoid mistakes with attribution of value to the variable instead of comparison. Accidentally one can miss one or two "=" signs and the expression may end to `if ($this->lastExportTimestamp = null) {` and will always return true, without any notice. Switching the operands will throw an error

tests/Sdk/Unit/Support/HasTraceProvider.php
beniamin: Actually I think cs-fixer is to blame.

sdk/Trace/SpanOptions.php
morrisonlevi: Are other tracers using a noop span or their equivalent of null?

beniamin: No, but since we don't create a global span when Tracer is instantiated, then the getActiveSpan may return null. or should I throw an exception instead?

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
morrisonlevi: "time elapsed" is the kind of thing that should use the monotonic clock. The realtime clock can move backwards; the monotonic clock will not (assuming the OS provides a good one, can still happen on some platforms).

sdk/Trace/SpanOptions.php
morrisonlevi: IMO this should return `null` or a no-op span; no exception. We can make a no-op span on demand; no need to store it globally. I'm curious what other tracers do (in Open Tracing it would be a no-op span, FYI).

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
beniamin: Good catch. I will use now() instead of timestamp().

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
bobstrecansky: Fine with me, thanks for the explanation

bobstrecansky: Fine with me, thanks for the explanation

dkarlovi: Yoda conditions are pretty obsolete as a concept with modern static analysis:
https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not

sdk/Trace/Tracer.php
```diff
@@ -23,23 +23,22 @@ class Tracer implements API\Tracer
...
     }
 
     /**
      * @return Span
      */
-    public function getActiveSpan(): API\Span
+    public function getActiveSpan(): ?API\Span
```
morrisonlevi, line 41: Docblocks that don't add any information should be removed, I think. In this case there's actually a mismatch between an API Span and an SDK Span o.O

dkarlovi: Static analysis should have caught that.

BnMcG:
Is there anything needed to get this over the line? I'm finding the fact that Jaeger warns about not being able to find a parent span a little annoying, so happy to put in any work required to get this merged.

beniamin:
@BnMcG I just added a To-Do section on this PR description. You can start from there.

BnMcG:
@beniamin I've drafted https://github.com/beniamin/opentelemetry-php/pull/4 which replaces sets the default active span on the tracer to be the NoopSpan, and generates a context in `startAndActivateSpan` if the active span is a `NoopSpan`. 

This doesn't seem fundamentally that different from what you had before (except now the active span should never be `null`), so I just wanted to check in and see if that's what you had in mind, or if I've missed a trick somewhere, before I fix up the tests, etc.

beniamin:
You are right, the idea was to avoid having a `null` active span. I took a look at the opentelemtry-js SIG and they are doing same thing.

BnMcG:
@beniamin Great - I've raised a PR against your branch with my suggested changes. All the unit tests pass locally, and I've tested exporting to Jaeger and Zipkin with no problems.

BnMcG:
I'll raise another PR to fix the failing checks.

beniamin:
I think it should be rebased with master, before fixing psalm check because psalm is not installed on this branch.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -126,7 +126,14 @@ protected function queueReachedLimit(): bool
 
     protected function enoughTimeHasPassed(): bool
     {
-        $now = $this->clock->timestamp();
+        $now = (int) ($this->clock->now() / 1e6);
```
bobstrecansky, line 129: We could use the [RealtimeClock](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/RealtimeClock.php) Class to handle this, but not a blocker.

bobstrecansky (approved):
Looks good to me.  One question but not a blocker.  I can merge this at EoD if we don't have any more discussion points.

sdk/Trace/SpanProcessor/BatchSpanProcessor.php
morrisonlevi: "Has time passed?" is a duty of a monotonic clock. If it relates to time in the real world "What time is it right now?" and such then it's a concern for a Realtime clock.

PR #182, opened 2020-09-22
Added probability sampling based on TraceId

shovnik:
This pull request hopes to resolve this issue: open-telemetry#86 which requires implementation of the ShouldSample function in the TraceIdRatioBasedSampler as described in these specs: https://github.com/open-telemetry/opentelemetry-specification/blob/e0bd41739a92304c0be5d68840f434e281b5f8b2/specification/trace/sdk.md#shouldsample.

The exact algorithm for acquiring the samples based on traceId wasn't described in the Spec, so I took an approach similar to the one implemented in the python sdk but adapted it to php.
opentelemetry-python sampling:
https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-sdk/src/opentelemetry/sdk/trace/sampling.py

### DIFF
--- a/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
+++ b/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
@@ -53,7 +53,7 @@ class TraceIdRatioBasedSampler implements Sampler
             return new SamplingResult(SamplingResult::RECORD_AND_SAMPLED, $attributes, $links);
         }
         /**
-         * Since php can only store up to 
+         * Since php can only store up to 63 bit positive integers
          */
         $traceIdLimit = (1 << 60) - 1;
         $lowerOrderBytes = hexdec(substr($traceId, strlen($traceId) - 15, 15));
--- a/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
+++ b/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
@@ -29,6 +29,9 @@ class TraceIdRatioBasedSampler implements Sampler
      */
     public function __construct(float $probability)
     {
+        if ($probability < 0.0 or $probability > 0.0) {
+            throw new InvalidArgumentException("probability should be be between 0.0 and 1.0.");
+        }
         $this->probability = $probability;
     }
 
@@ -50,7 +53,12 @@ class TraceIdRatioBasedSampler implements Sampler
         }
 
         // TODO: implement as a function of TraceID when specification is ready
-        $decision = (lcg_value() < $this->probability)
+        /**
+         * For compatibility with 64 bit IDs, the sampler checks the 64 first bits of the trace ID to decide whether to sample
+         */
+        $traceIdLimit = (1 << 64) - 1;
+        $traceIdLowerBytes = ($traceId[0]<<56) + ($traceId[1]<<48) + ($traceId[2]<<40) + ($traceId[3]<<32) + ($traceId[4]<<24) + ($traceId[5]<<16) + ($traceId[6]<<8) + $traceId[7]; 
+        $decision = ($traceIdLowerBytes < $probability * tradeIdLimit)
             ? SamplingResult::RECORD_AND_SAMPLED
             : SamplingResult::NOT_RECORD;
 
--- a/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
+++ b/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Sdk\Trace\Sampler;
 
+use InvalidArgumentException;
 use OpenTelemetry\Sdk\Trace\Sampler;
 use OpenTelemetry\Sdk\Trace\SamplingResult;
 use OpenTelemetry\Trace as API;
--- a/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
+++ b/sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
@@ -29,8 +29,8 @@ class TraceIdRatioBasedSampler implements Sampler
      */
     public function __construct(float $probability)
     {
-        if ($probability < 0.0 or $probability > 0.0) {
-            throw new InvalidArgumentException("probability should be be between 0.0 and 1.0.");
+        if ($probability < 0.0 || $probability > 1.0) {
+            throw new InvalidArgumentException('probability should be be between 0.0 and 1.0.');
         }
         $this->probability = $probability;
     }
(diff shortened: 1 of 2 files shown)


sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
```diff
@@ -45,14 +48,21 @@ public function shouldSample(
...
-            ? SamplingResult::RECORD_AND_SAMPLED
-            : SamplingResult::NOT_RECORD;
+        /**
+         * Since php can only store up to 63 bit positive integers
+         */
+        $traceIdLimit = (1 << 60) - 1;
+        $lowerOrderBytes = hexdec(substr($traceId, strlen($traceId) - 15, 15));
+        $traceIdCondition = $lowerOrderBytes < round($this->probability * $traceIdLimit);
+        $decision = SamplingResult::NOT_RECORD;
+        // TODO: Also sample Spans with remote parent
```
shovnik, line 62: I am not sure how to find out if the current Span has a remote parent without its Span Context, might be related to the other issue requiring implementation of the IsRemote attribute.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
```diff
@@ -45,14 +48,21 @@ public function shouldSample(
...
             return new SamplingResult(SamplingResult::RECORD_AND_SAMPLED, $attributes, $links);
         }
-
-        // TODO: implement as a function of TraceID when specification is ready
-        $decision = (lcg_value() < $this->probability)
-            ? SamplingResult::RECORD_AND_SAMPLED
-            : SamplingResult::NOT_RECORD;
+        /**
+         * Since php can only store up to 63 bit positive integers
+         */
```
shovnik, line 57: I originally wanted to use 64 bits of data from the traceId, but then I realized integers must be signed in php (only 63 bits of positive integers), so I decided to reduce to using 60 bits to only have to extract 15 hex characters instead of 16.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
```diff
@@ -29,6 +30,9 @@ class TraceIdRatioBasedSampler implements Sampler
      */
     public function __construct(float $probability)
     {
+        if ($probability < 0.0 || $probability > 1.0) {
+            throw new InvalidArgumentException('probability should be be between 0.0 and 1.0.');
```
bobstrecansky, line 34: You may want to add tests around this to validate the cases where probability < 0.0 and probability > 1.0.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
bobstrecansky: Yes - I believe we may have to implement the SpanContext `isRemote` attribute for this.  I think a TODO is sufficient here.  Bonus points for creating an issue for this once this PR is merged.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
bobstrecansky: The only issue I can potentially think of with this is we generate a 16 hex character string on the traceID initialization:
https://github.com/open-telemetry/opentelemetry-php/blob/ce036e540f24d48f2f37994118020ff83b8e86b9/sdk/Trace/SpanContext.php#L86

This could potentially have a conflict (albeit a small one) in a large number of traces?

bobstrecansky (commented):
This is a great first PR!  I've added a couple comments to get us started.  @beniamin will want to review this too.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
shovnik: I made sure to only take the last 15 Hex characters of the traceId below to be compatible with any traceId with at least 15 hex characters in case of traceId's of larger size. I am not sure how traceId's with 16 hex characters might be an issue, but I might be misunderstanding your remark.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
shovnik: I added the required tests.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
shovnik: Will do.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
bobstrecansky: I think I'm thinking too much 😅  - this should be fine!

bobstrecansky (approved):

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
morrisonlevi: > I decided to reduce to using 60 bits to only have to extract 15 hex characters instead of 16.

Help me understand why you can't use all 16. Would using the last character throw off the statistics of the sampling or something given the sign bit is always 0?

shovnik: Actually even the traceId limit which would become (1 << 64) -1 would become the value -1 die to the sign bit causing comparison issues. I am not sure how to use more than 63 bits of the integer without having sign issues.

shovnik: I have been trying really hacky ways to utilize the full 64-bits of the trace ID such as story the upper and lower bits separately and doing bit manipulation and carry over to do multiplication and addition, however I realized that since the probability is a float, multiplying the probably by the range limit will inevitably have less than 60-bits of significant digits as floats are 64-bit with many bits dedicated to the exponent. Therefore, changing how I manipulate the traceID will not increase the number of bits of precision I have over 60-bit due to the precision of floats.

bobstrecansky: Do you have any other avenues that you would like to explore or are we happy with this solution now after that reasoning?

shovnik: I can't think of a better solution at the moment, and I believe the current implementation fulfills its purpose. I discussed sampling method in the sampling SIG, and there was talk about hashing the traceID before computing probability threshold in the future but for now sticking to a similar implementation to what was done in other SDKs (python in this case). Sorry, if I am responding slowly, working on a large project I got last week. I am happy with the solution as it is.

bobstrecansky: Ok, I'm also happy with this implementation for now too.  @morrisonlevi - if you're okay with it, I think we can merge this and change later if necessary.

bobstrecansky:
@morrisonlevi - anything else you want to mention before we merge this?  Just wanted to make sure we're all on the same page.

sdk/Trace/Sampler/TraceIdRatioBasedSampler.php
morrisonlevi: I think there is a way to do it, but I won't block the merge on it 👍

Issue #141, opened 2020-07-04
Finish the isRemote function


bobstrecansky:
Currently the `isRemote` function just returns false:
https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/Span.php#L183-L191

More details about `isRemote` can be found in the SpanContext section of the spec:
[SpanContext](https://github.com/open-telemetry/opentelemetry-specification/blob/4daee3af756be75edbd17085f6d382d963027bc8/specification/trace/api.md#spancontext)

Success criteria:
Implement `isRemote` functionality

bobstrecansky:
@michaelorr is working on this.

bobstrecansky:
This function has a signature but still needs to be completed:
https://github.com/open-telemetry/opentelemetry-php/blob/3a83788a4026e55c9eea88d0625c9193f0ad9113/sdk/Trace/Span.php#L202

zsistla:
I'll work on this.

bobstrecansky:
Thanks Amber! Updated the ticket accordingly.

zsistla:
[PR](https://github.com/open-telemetry/opentelemetry-php/pull/194/) is up and current status is `pending review`.

PR #194, opened 2020-09-30
Finish the isRemote function for issue #141

zsistla:
Completed the `isRemote` function in `span.php`.
Additionally, added functionality to `startAndActivateSpanFromContext` and allowed the `isRemote` flag to be indicated according to the spec.

### DIFF
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -79,6 +79,52 @@ class Tracer implements API\Tracer
      * todo: fix ^
      * -> Is there a reason we didn't add this already?
      * @param string $name
+     * @param SpanContext $parent
+     * @param bool $isRemote
+     * @return Span
+     */
+    public function startAndActivateSpanFromContext(string $name, SpanContext $parent, bool $isRemote = false): API\Span
+    {
+        /*
+         * Pass in true if the SpanContext was propagated from a
+         * remote parent. When creating children from remote spans,
+         * their IsRemote flag MUST be set to false.
+         */
+        $context = SpanContext::fork($parent->getTraceId(), $parent->isSampled(), $isRemote);
+        $span = $this->generateSpanInstance($name, $context);
+
+        if ($span->isRecording()) {
+            $this->provider->getSpanProcessor()->onStart($span);
+        }
+
+        $this->setActiveSpan($span);
+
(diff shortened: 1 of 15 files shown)


sdk/Trace/Tracer.php
```diff
@@ -63,6 +63,52 @@ public function setActiveSpan(API\Span $span): void
...
+     * - SpanKind, default to SpanKind.Internal if not specified.
+     * - Attributes - similar API with Span::SetAttributes. These attributes
+     *   will be used to make a sampling decision as noted in sampling
+     *   description. Empty list will be assumed if not specified.
+     * - Start timestamp, default to current time. This argument SHOULD only
+     *   be set when span creation time has already passed. If API is called
+     *   at a moment of a Span logical start, API user MUST not explicitly
+     *   set this argument.
+     * todo: fix ^
+     * -> Is there a reason we didn't add this already?
```
bobstrecansky, line 80: Not one that I'm aware of.  Perhaps we should make a ticket to track that work.

sdk/Trace/Tracer.php
```diff
@@ -63,6 +63,52 @@ public function setActiveSpan(API\Span $span): void
...
+        $this->setActiveSpan($span);
+
+        return $this->active;
+    }
+    /* Span creation MUST NOT set the newly created Span as the currently active
+ * Span by default, but this functionality MAY be offered additionally as a
+ * separate operation.
+ * todo: fix ^
+ * --> what would you like to do for this?  Would it make sense to return $this->tail[] ?
+ * Do we want to offer the setActiveSpan on a newly created span as a separate function?
```
bobstrecansky, line 109: Using $this->tail[] makes sense to me.  I don't think it makes sense to offer `setActiveSpan` on a newly created span as a separate function?

sdk/Trace/Tracer.php
```diff
@@ -95,7 +141,6 @@ public function startAndActivateSpan(string $name): API\Span
 
         return $this->active;
     }
-
```
bobstrecansky, line 98: I'm assuming this was just an autolinter.

bobstrecansky (approved):
Two questions on TODOs but no blockers.  Will give the other approvers until Friday to raise concerns with this PR.

PR #198, opened 2020-10-22
Add psalm static analysis tool to codebase

### LINKED ISSUE #196: Add Psalm and PHPStan to Static Analysis tools to CI/CD tooling
Based on discussions here https://github.com/open-telemetry/opentelemetry-php/pull/185, we are looking forward to adding the following static analysis tools

 [PHPStan](https://github.com/phpstan/phpstan)
 [Psalm](https://github.com/vimeo/psalm)

prondubuisi:
Fixes #196 
hello @dkarlovi  @bobstrecansky this PR adds Psalm

running 
`make psalm` or  `./vendor/bin/psalm`

```
Scanning files...
Analyzing files...



ERROR: LessSpecificImplementedReturnType - sdk/Metrics/HasLabels.php:17:34 - The inherited return type 'array<array-key, string>' for OpenTelemetry\Metrics\LabelableMetric::getLabels is more specific than the implemented return type for OpenTelemetry\Sdk\Metrics\HasLabels::getlabels 'array<array-key, mixed>' (see https://psalm.dev/166)
    public function getLabels(): array


------------------------------
1 errors found
------------------------------
87 other issues found.
You can display them with --show-info=true
------------------------------

Checks took 0.03 seconds and used 5.745MB of memory

```

On the other hand running

`make psalm-info` or ` ./vendor/bin/psalm --show-info=true`

returns more extensive info as seen here https://pastebin.com/4wy2yHV6

### DIFF
--- a/sdk/Metrics/HasLabels.php
+++ b/sdk/Metrics/HasLabels.php
@@ -12,7 +12,7 @@ trait HasLabels
     protected $labels = [];
 
     /**
-     * {@inheritDoc}
+     * @return	array<string>
      */
     public function getLabels(): array
     {
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -50,6 +50,9 @@ jobs:
       env:
         PHAN_DISABLE_XDEBUG_WARN: 1
       run: vendor/bin/phan
+    
+    - name: Run Psalm
+      run: vendor/bin/psalm
 
     - name: Run PHPUnit
       run: vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ test:
 	$(DC_RUN_PHP) php ./vendor/bin/phpunit --colors=always --coverage-text --testdox --coverage-clover coverage.clover
 phan:
 	$(DC_RUN_PHP) env PHAN_DISABLE_XDEBUG_WARN=1 php ./vendor/bin/phan
+psalm:
+	$(DC_RUN_PHP) php ./vendor/bin/psalm
+psalm-info:
+	$(DC_RUN_PHP) php ./vendor/bin/psalm --show-info=true
 trace examples: FORCE
 	docker-compose up -d
 	$(DC_RUN_PHP) php ./examples/AlwaysOnZipkinExample.php
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
         "phpunit/phpunit": "^9.3",
         "composer/xdebug-handler": "^1.3",
         "phan/phan": "^3.0",
-        "friendsofphp/php-cs-fixer": "^2.16"
+        "friendsofphp/php-cs-fixer": "^2.16",
+        "vimeo/psalm": "^4.0"
     }
 }
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="4"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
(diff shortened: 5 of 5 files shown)


dkarlovi:
> returns more extensive info as seen here https://pastebin.com/4wy2yHV6

Why does Pastebin find Psalm's output offensive? :thinking:

dkarlovi:
@prondubuisi thank you for working on this! Would we add this directly to Github actions right away, making it run as part of CI?

prondubuisi:
> > returns more extensive info as seen here https://pastebin.com/4wy2yHV6
> 
> Why does Pastebin find Psalm's output offensive? 🤔

This is the result of handing things off to algorithms , sometimes they get it wrong

dkarlovi (changes requested):
Two changes from my POV:
1. add to Github actions [same way Phan is added](https://github.com/open-telemetry/opentelemetry-php/blob/48581e13ec209bdfa81825767817d9bd75319c8c/.github/workflows/php.yml#L49-L52)
2. fix the issue reported, IIRC it's because traits don't have parents, can't use `@inheritdoc` there, fix is to specify the concrete type in the trait

This should enable the project to start fixing issues by lowering the `errorLevel`.

prondubuisi:
> 1. same way Phan is added

Thanks for the pointers @dkarlovi , updated.

sdk/Metrics/HasLabels.php
```diff
@@ -12,7 +12,8 @@ trait HasLabels
     protected $labels = [];
 
     /**
-     * {@inheritDoc}
+     * @access	public
```
dkarlovi, line 15: Not sure this line is required, probably not.

dkarlovi (approved):
Small gripe, otherwise LGTM! :+1: Thanks once again.

sdk/Metrics/HasLabels.php
prondubuisi: Fixed!

dkarlovi:
```
Psalm was able to infer types for 88.7661% of the codebase
```

Nice, will probably be able to raise this a bit, making the codebase more robust in the process! :rocket:

bobstrecansky:
This looks great!  Thanks for doing this @prondubuisi.  Have you tested this on your local branch?  You should be able to see if CI passes with the Psalm checks enabled.

dkarlovi:
@bobstrecansky you can already see it in the GH Actions output, it works for sure.

prondubuisi:
> This looks great! Thanks for doing this @prondubuisi. Have you tested this on your local branch? You should be able to see if CI passes with the Psalm checks enabled.

Yes tested locally, also Github actions is able to pick the psalm addition. I am curious though is there another way to run CI checks locally except by running the `vendor` or ` make`  commands?

dkarlovi:
@prondubuisi there's some ways to [emulate running Github actions locally](https://github.com/nektos/act), but I haven't tried it.

My solution to this issue is to have [Github actions run my Makefile](https://github.com/dkarlovi/xezilaires-dev/blob/master/.github/workflows/build.yaml#L42).

bobstrecansky:
I've actually used a fork of the repo to test this before.  I'll do this quickly and show you what I mean.

bobstrecansky:
It also appears that adding Psalm roughly doubles our CI time, from ~30s to ~1m.  I think this is okay, just felt it was important to call it out.

bobstrecansky (approved):

bobstrecansky:
@prondubuisi - If you're happy with this now, I'll merge when you're ready for me to.

dkarlovi:
> It also appears that adding Psalm roughly doubles our CI time, from ~30s to ~1m. I think this is okay, just felt it was important to call it out.

Psalm does some in-depth analysis so having the CI run for 30s more is nothing in comparison. It does support caching and diffs so we could likely be able to cache the analysis and reuse it on subsequent runs if the number becomes problematic.

I appreciate your thoroughness about noticing it. :+1:

prondubuisi:
Hello @bobstrecansky good to go!

PR #200, opened 2020-10-23
add phpstan static analysis tool

prondubuisi:
Fixes issues #196 

Hello @dkarlovi @bobstrecansky any input on this?

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,10 @@
+parameters:
+    level: 0
+    paths:
+        -   Context
+        -   api
+        -   contrib
+        -   sdk
+    
+
+
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,10 +1,7 @@
 parameters:
-    level: 0
+    level: 4
     paths:
         -   Context
         -   api
         -   contrib
         -   sdk
-    
-
-
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -53,6 +53,9 @@ jobs:
     
     - name: Run Psalm
       run: vendor/bin/psalm
+    
+    - name: Run Phpstan
+      run: vendor/bin/phpstan analyse
 
     - name: Run PHPUnit
       run: vendor/bin/phpunit --coverage-text --coverage-clover=coverage.clover
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -40,7 +40,7 @@ class Context
      * @param mixed|null $value
      * @param Context|null $parent Reference to the parent object
      */
-    public function __construct(?ContextKey $key=null, $value=null, $parent=null)
+    public final function __construct(?ContextKey $key=null, $value=null, $parent=null)
     {
         $this->key = $key;
         $this->value = $value;
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ psalm:
 	$(DC_RUN_PHP) php ./vendor/bin/psalm
 psalm-info:
 	$(DC_RUN_PHP) php ./vendor/bin/psalm --show-info=true
+phpstan: 
+	$(DC_RUN_PHP) php ./vendor/bin/phpstan analyse
 trace examples: FORCE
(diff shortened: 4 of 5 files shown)


prondubuisi:
Also this is the error by PHPStan at the moment

Note: Using configuration file /gitrepos/open-php/phpstan.neon.
 78/78 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------- 
  Line   Context/Context.php                                    
 ------ ------------------------------------------------------- 
  61     Unsafe usage of new static().                          
         💡 Consider making the class or the constructor final.  
  85     Unsafe usage of new static().                          
         💡 Consider making the class or the constructor final.  
  88     Unsafe usage of new static().                          
         💡 Consider making the class or the constructor final.  
  141    Unsafe usage of new static().                          
         💡 Consider making the class or the constructor final.  
 ------ ------------------------------------------------------- 

                                                                                                                   
 [ERROR] Found 4 errors                                                                                            
                                                                                                                   

Ndubuisis-MacBook-Pro:open-php prondubuisi$

phpstan.neon
```diff
@@ -0,0 +1,10 @@
+parameters:
+    level: 0
+    paths:
+        -   Context
+        -   api
+        -   contrib
+        -   sdk
+
```
bobstrecansky, line 8: Extra whitespace?

phpstan.neon
```diff
@@ -0,0 +1,10 @@
+parameters:
+    level: 0
```
bobstrecansky, line 2: I'm assuming we are going to start at `level 0` here and work our way up?  Is there a big difference in running different levels here?

bobstrecansky (commented):
A couple questions, but I very much like the direction of this PR.  We should fix those psalm errors in a separate PR.

phpstan.neon
dkarlovi: We can start at 4 since my PR brought us up to that IIRC.

bobstrecansky:
@dkarlovi - did you run into that error messaging from psalm when you were making your PR?  I did some research and found some github issues related to:

`Unsafe usage of new static().`

https://github.com/phpstan/phpstan/issues/2963

dkarlovi:
This is PHPStan, we already enabled Psalm.

Basically, new static is unsafe because the extending class can change your constructor (for example, make it have a mandatory param where your doesn't), making your code crash.

Since the error is correct, we can either fix the code or ignore the error in PHPStan config.

bobstrecansky:
Oh my apologies, I did mean PHPStan, typo!
I think it warrants fixing if possible.

prondubuisi:
Hello @bobstrecansky @dkarlovi  Great Convo! @dkarlovi I think we are more interested in the fix. I will be pushing this up and fixing comments so far.  How do you think I can fix the current error @dkarlovi

dkarlovi:
Easiest is to make the constructor final

phpstan.neon
prondubuisi: Fixed

phpstan.neon
prondubuisi: Moved to 4

bobstrecansky (approved):
LGTM.

dkarlovi (approved):
LGTM!

dkarlovi:
@prondubuisi would you be interested in adding mutation testing (in a separate PR)? Since the project already has a test suite with coverage, adding mutation should be pretty straightforward.

@bobstrecansky WDYT?

bobstrecansky:
Sounds like a plan to me!  I say we merge this as is and we can create a separate ticket / PR to track mutation testing.

prondubuisi:
> @prondubuisi would you be interested in adding mutation testing (in a separate PR)? Since the project already has a test suite with coverage, adding mutation should be pretty straightforward.
> 
> @bobstrecansky WDYT?

Sure @dkarlovi  will do.

PR #204, opened 2020-11-02
Implement the sampler into the tracer and trace provider.

zsistla:
Implement the sampler into the tracer and trace provider.

### DIFF
--- a/sdk/Trace/TracerProvider.php
+++ b/sdk/Trace/TracerProvider.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\Sdk\Trace;
 
 use OpenTelemetry\Sdk\Resource\ResourceConstants;
 use OpenTelemetry\Sdk\Resource\ResourceInfo;
+use OpenTelemetry\Sdk\Trace\Sampler\AlwaysOnSampler;
 use OpenTelemetry\Sdk\Trace\SpanProcessor\SpanMultiProcessor;
 use OpenTelemetry\Trace as API;
 
@@ -27,10 +28,20 @@ final class TracerProvider implements API\TracerProvider
      */
     private $resource;
 
-    public function __construct(?ResourceInfo $resource = null)
+    /**
+     * @var Sampler
+     */
+    private $sampler;
+
+    public function __construct(?ResourceInfo $resource = null, ?Sampler $sampler = null)
     {
         $this->spanProcessors = new SpanMultiProcessor();
         $this->resource = $resource ?? ResourceInfo::emptyResource();
+        if (null == $sampler) {
+            $sampler = new AlwaysOnSampler();
+        } else {
+            $this->sampler = $sampler;
+        }
         register_shutdown_function([$this, 'shutdown']);
     }
 
--- a/sdk/Trace/TracerProvider.php
+++ b/sdk/Trace/TracerProvider.php
@@ -37,11 +37,8 @@ final class TracerProvider implements API\TracerProvider
     {
         $this->spanProcessors = new SpanMultiProcessor();
         $this->resource = $resource ?? ResourceInfo::emptyResource();
-        if (null == $sampler) {
-            $this->sampler = new AlwaysOnSampler();
-        } else {
-            $this->sampler = $sampler;
-        }
+        $this->sampler = $sampler ?? new AlwaysOnSampler();
+
         register_shutdown_function([$this, 'shutdown']);
     }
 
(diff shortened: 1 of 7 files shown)


sdk/Trace/TracerProvider.php
```diff
@@ -27,10 +28,17 @@ final class TracerProvider implements API\TracerProvider
...
+    /**
+     * @var Sampler
+     */
+    private $sampler;
+
+    public function __construct(?ResourceInfo $resource = null, ?Sampler $sampler = null)
     {
         $this->spanProcessors = new SpanMultiProcessor();
         $this->resource = $resource ?? ResourceInfo::emptyResource();
+        $this->sampler = $sampler ?? new AlwaysOnSampler();
```
bobstrecansky, line 40: Should this be `AlwaysOnSampler()` if no sampler is provided?  Maybe it should be `AlwaysOffSampler()`?  Don't have strong opinions here.

tests/Sdk/Unit/Trace/TracerProviderTest.php
```diff
@@ -21,4 +25,41 @@ public function gettingSameTracerMultipleTimesShouldReturnSameObject()
...
+
+    /**
+     * @test
+     */
+    public function newTraceProviderDefaultsToAlwaysOnSampler()
+    {
+        $traceProvider = new TracerProvider();
+        $description = $traceProvider->getSampler()->getDescription();
+
+        self::assertSame($description, 'AlwaysOnSampler');
```
bobstrecansky, line 37: Should we potentially test for something else besides the `$description` here?

tests/Sdk/Unit/Trace/TracerProviderTest.php
```diff
@@ -21,4 +25,41 @@ public function gettingSameTracerMultipleTimesShouldReturnSameObject()
...
+        $description = $traceProvider->getSampler()->getDescription();
+
+        self::assertSame($description, 'ParentBased');
+
+        $traceProvider = new TracerProvider(null, new AlwaysOnSampler());
+        $description = $traceProvider->getSampler()->getDescription();
+
+        self::assertSame($description, 'AlwaysOnSampler');
+
+        $traceProvider = new TracerProvider(null, new TraceIdRatioBasedSampler(1.0));
```
bobstrecansky, line 60: Maybe we should have this test be something other than `1.0` - for the assertion.  No strong feelings about this just a thought.

bobstrecansky (commented):
A couple comments; no blockers.

tests/Sdk/Unit/Trace/TracerProviderTest.php
zsistla: I could change it to 0.5 or something like that.  The main thing was to check the samplers were getting used, since I thought the samplers' functionality had already been thoroughly unit tested.

tests/Sdk/Unit/Trace/TracerProviderTest.php
zsistla: I had thought that since the samplers already had unit tests verifying the logic they do regarding samplingResult and decisions, the main point of this would be to check that now we are using it.  I can add more, but it would probably be really similar to what the sampler unit tests do.  Let me know if you want that added.

sdk/Trace/TracerProvider.php
zsistla: It could go either way.  My thinking with this was ease of use.  It's easier to use if spans get recorded by default vs a user having to figure out a sampler to add just to get spans recorded.  I thought more users would prefer to have spans recorded vs not recorded.

What do you think?

tests/Sdk/Unit/Trace/TracerProviderTest.php
bobstrecansky: Was just curious on the reasoning.  This is acceptable to me!

tests/Sdk/Unit/Trace/TracerProviderTest.php
bobstrecansky: Yup; I think something like 0.5 makes sense to me - 1.0 might be easily confused with `AlwaysOn` at some point.

bobstrecansky (approved):
LGTM; will merge tomorrow if no gripes from others.

sdk/Trace/TracerProvider.php
bobstrecansky: Sounds good to me.  Easy enough to change later on if need be.

PR #211, opened 2020-11-05
209 toggle default tags

### LINKED ISSUE #209: How to get rid of op.status_code and op.status_description tag?
Using v0.0.1 of telemetry

image

Wondering what these tags are for and how to get rid of them?
We have 1000s of spans so we don't want to have this tag getting in our way.

Is it possible to disable it by configuration?

Code that adds these tags is in
https://github.com/open-telemetry/opentelemetry-php/blob/v0.0.1/contrib/Zipkin/SpanConverter.php#L57-L60

adhocore:
closes #209 

this is a quick, simple and probably dirty approach to address #209 done in rush.
i have allowed `edits by maintainer` so please feel free to adjust as required.

### DIFF
--- a/contrib/Zipkin/SpanConverter.php
+++ b/contrib/Zipkin/SpanConverter.php
@@ -8,6 +8,8 @@ use OpenTelemetry\Trace\Span;
 
 class SpanConverter
 {
+    const CONFIG_ADD_DEFAULT_TAG = 'add_default_tag';
+
     const STATUS_CODE_TAG_KEY = 'op.status_code';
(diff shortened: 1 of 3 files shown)


bobstrecansky (approved):
This seems reasonable to me.  I don't have strong opinions on this.

bobstrecansky:
@beniamin made a good point in chat - the spec says that the status code is a MUST: https://github.com/open-telemetry/opentelemetry-specification/blob/e0bd41739a92304c0be5d68840f434e281b5f8b2/specification/trace/sdk_exporters/zipkin.md#status

PR #213, opened 2020-11-11
Trace context propagation

Fahmy-Mohammed:
This adds the w3c `traceparent` header propagation. Please see [W3C traceparent documentation](https://www.w3.org/TR/trace-context/#traceparent-header ) for more information. `Tracestate` propagation to follow in another PR.

### DIFF
--- a/api/Trace/PropagationSetter.php
+++ b/api/Trace/PropagationSetter.php
@@ -14,5 +14,5 @@ interface PropagationSetter
      * @param string $value
      * @return void
      */
-    public function set($carrier, string $key, string $value) : void;
+    public function set(&$carrier, string $key, string $value) : void;
 }
--- a/api/Trace/PropagationSetter.php
+++ b/api/Trace/PropagationSetter.php
@@ -9,7 +9,7 @@ interface PropagationSetter
     /**
      * Set the value for a given key on the associated carrier.
      *
-     * @param mixed  &$carrier
+     * @param mixed  $carrier
      * @param string $key
      * @param string $value
      * @return void
--- a/api/Trace/PropagationSetter.php
+++ b/api/Trace/PropagationSetter.php
@@ -9,7 +9,7 @@ interface PropagationSetter
     /**
      * Set the value for a given key on the associated carrier.
      *
-     * @param mixed  $carrier
+     * @param mixed  &$carrier
      * @param string $key
      * @param string $value
      * @return void
--- a/api/Trace/TextMapFormatPropagator.php
+++ b/api/Trace/TextMapFormatPropagator.php
@@ -18,7 +18,7 @@ interface TextMapFormatPropagator
      * the encoded SpanContext using Setter into it's associated carrier.
      *
      * @param SpanContext $context
-     * @param mixed $carrier
+     * @param mixed &$carrier
      * @param PropagationSetter $setter
      * @return void
      */
--- a/api/Trace/TextMapFormatPropagator.php
+++ b/api/Trace/TextMapFormatPropagator.php
@@ -22,7 +22,7 @@ interface TextMapFormatPropagator
      * @param PropagationSetter $setter
      * @return void
      */
-    public function inject(SpanContext $context, $carrier, PropagationSetter $setter) : void;
+    public static function inject(SpanContext $context, &$carrier, PropagationSetter $setter) : void;
 
     /**
      * Retrieves encoded SpanContext using Getter from the associated carrier.
@@ -31,5 +31,5 @@ interface TextMapFormatPropagator
      * @param PropagationGetter $getter
      * @return SpanContext
      */
-    public function extract($carrier, PropagationGetter $getter): SpanContext;
+    public static function extract($carrier, PropagationGetter $getter): SpanContext;
 }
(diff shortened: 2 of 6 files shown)


api/Trace/PropagationSetter.php
```diff
@@ -14,5 +14,5 @@ interface PropagationSetter
      * @param string $value
      * @return void
      */
-    public function set($carrier, string $key, string $value) : void;
+    public function set(&$carrier, string $key, string $value) : void;
```
bobstrecansky, line 17: You may want to update the docbloc for this one:
 
`* @param mixed  $carrier` 

and should probably read 

`* @param mixed  &$carrier`

It might help to make others recognize that this is pass by reference.

api/Trace/TextMapFormatPropagator.php
```diff
@@ -22,7 +22,7 @@ public function fields() : array;
      * @param PropagationSetter $setter
      * @return void
      */
-    public function inject(SpanContext $context, $carrier, PropagationSetter $setter) : void;
+    public static function inject(SpanContext $context, &$carrier, PropagationSetter $setter) : void;
```
bobstrecansky, line 25: same in re: `&$carrier` docblock

sdk/Trace/TraceContext.php
```diff
@@ -0,0 +1,94 @@
...
+namespace OpenTelemetry\Sdk\Trace;
+
+use OpenTelemetry\Trace as API;
+
+final class TraceContext implements API\TextMapFormatPropagator
+{
+    public const TRACEPARENT = 'http_traceparent';
+    public const TRACESTATE = 'http_tracestate';
+
+    // TODO Consolidate these and the validity checks in SpanContext.php
```
bobstrecansky, line 14: This just looks to be a superset of the validity checks in [SpanContext.php](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/SpanContext.php) for reference.

sdk/Trace/TraceContext.php
```diff
@@ -0,0 +1,94 @@
...
+        $setter->set($carrier, self::TRACEPARENT, $traceparent);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function extract($carrier, API\PropagationGetter $getter): API\SpanContext
+    {
+        $traceparent = $getter->get($carrier, self::TRACEPARENT);
+        if (null === $traceparent) {
```
bobstrecansky, line 47: Maybe we could swap out the `yoda_style` in this one?

`if $traceparent === null) {`

I'm actually surprised php-cs-fixer didn't complain about this one.  Verified locally.

sdk/Trace/TraceContext.php
```diff
@@ -0,0 +1,94 @@
...
+            throw new \InvalidArgumentException('Traceparent not present');
+        }
+
+        // Traceparent = {version}-{trace-id}-{parent-id}-{trace-flags}
+        $pieces = explode('-', $traceparent);
+
+        $peicesCount = count($pieces);
+        if ($peicesCount != 4) {
+            throw new \InvalidArgumentException('Unable to extract traceparent. Contains Invalid values');
```
bobstrecansky, line 56: Maybe we could make this exception mention that there are an incorrect number of peices?

bobstrecansky (approved):
A couple small questions, but no blockers.  This PR looks great, thanks @Fahmy-Mohammed !

api/Trace/PropagationSetter.php
Fahmy-Mohammed: Good catch! All fixed 👍🏾

sdk/Trace/TraceContext.php
Fahmy-Mohammed: Good catch. Old `C` habits die hard 🙂

sdk/Trace/TraceContext.php
Fahmy-Mohammed: Yes, you're right. The validity checks here are the same in `SpanContext.php`. I'll add some clean up here in a follow-up PR to consolidate those checks.

api/Trace/PropagationSetter.php
Fahmy-Mohammed: It looks like the `phpstan` tests fail when it is labeled as a reference parameter in the doc block. This [issue](https://github.com/phpstan/phpstan/issues/699) might provide some insight here. I'm not too sure if there's a way around this so it might be better to switch it back?

bobstrecansky: Yeah swapping it back is totally fine with me.  Sorry for the goose chase 😿

Fahmy-Mohammed: Swapped it back 👍🏾

bobstrecansky (approved):

TomHart:
Could you give an example of how to use this? I'm trying to understand what incoming header should be set with what value for it to use that trace ID rather than generating a new one

brettmc:
@TomHart - the distributed-tracing example is what I think you are after. Specifically, https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/distributed-tracing/src/index.php#L46 and L79 (set outgoing w3c trace context headers, and read them back as incoming headers in another request).

PR #214, opened 2020-11-11
Adding Proto folder to opentelemetry-php

riticksingh:
Signed-off-by:Ritick Gautam <riticksinghrajput@gmail.com>


bobstrecansky:
I'm not sure that we want to commit these to the repo - they will need to be updated on a given cadence depending on what's happening with https://github.com/open-telemetry/opentelemetry-proto/tree/master/opentelemetry/proto

It might make more sense to have a script to generate these.  That way our end users have the ability to generate the protobufs based on the changes that are being made in the proto repo.

Thoughts @benagricola @morrisonlevi ?

Proto/Collector/Trace/V1/TraceServiceClient.php
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+//
+namespace Opentelemetry\Proto\Collector\Trace\V1;
+
+/**
+ * Service that can be used to push spans between one Application instrumented with
+ * OpenTelemetry and an collector, or between an collector and a central collector (in this
+ * case spans are sent/received to/from multiple Applications).
+ */
+class TraceServiceClient extends \Grpc\BaseStub {
```
bobstrecansky, line 26: We talked about this in DM; it might make sense to skip phan checks on the proto repo (should we decide we actually want to commit this folder to the repo)

benagricola:
@bobstrecansky I'm guessing you meant to @beniamin 😄 but fwiw I think you've got a point.

I guess the question is how often will changes in the proto repo require corresponding changes in this library? If the answer is close to 1:1 then including the protos as a subtree, submodule or copied manually isn't such a bad idea.

Script generation is definitely more flexible but the tradeoff is introducing API inconsistencies if updating the library without the protos or vice-versa.

bobstrecansky:
@benagricola - yes sorry incorrect tag!  Thank you for your input though! 👍

bobstrecansky:
I'm thinking we should just have a bash script ala `script/proto-gen.sh` that performs the proto generation that you did for this PR.  That makes the generation a bit more modular for other contributors / users.  Thoughts?

riticksingh:
For that also we need to add a folder of opentelemetry-proto so that we can give path of that repo in shell file

riticksingh:
Suggested to remove the check of CI on this shell file

check.sh
```diff
@@ -0,0 +1,13 @@
+echo "Cloning opentelemetry-proto folder"
+svn checkout https://github.com/open-telemetry/opentelemetry-proto/trunk/opentelemetry
```
bobstrecansky, line 2: Is there a reason you chose to use subversion for this checkout rather than git?  I think most people would prefer git in this scenario.

I also think we should validate that the script has the `git` binary available to the user.  If it doesn't, we should message appropriately.  We will probably run this script in a docker container, so we should be sure to choose an image that has git available.

check.sh
```diff
@@ -0,0 +1,13 @@
...
+mkdir proto
+protoc --proto_path=./ --php_out=proto opentelemetry/proto/collector/logs/v1/logs_service.proto
```
bobstrecansky, line 4: We should also verify that the end user has protoc available in the same vein as my first comment.

check.sh
```diff
@@ -0,0 +1,13 @@
+echo "Cloning opentelemetry-proto folder"
+svn checkout https://github.com/open-telemetry/opentelemetry-proto/trunk/opentelemetry
+mkdir proto
```
bobstrecansky, line 3: We will probably want to add `proto/` to `.gitignore` so people don't inadvertently check it in.

check.sh
bobstrecansky: When we use git, we'll also want to reference https://github.com/open-telemetry/opentelemetry-proto rather than https://github.com/open-telemetry/opentelemetry-proto/trunk/opentelemetry

check.sh
```diff
@@ -0,0 +1,13 @@
...
+protoc --proto_path=./ --php_out=proto opentelemetry/proto/collector/logs/v1/logs_service.proto
+protoc --proto_path=./ --php_out=proto opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```
bobstrecansky, line 5: Is there a reason you defined each of these differently rather than using wildcards?

bobstrecansky (changes requested):
This file should also not live in the root directory of the project.  It should be placed in a subfolder (`script/` would be my preference (based on something loosely like [scripts-to-rule-them-all](https://github.com/github/scripts-to-rule-them-all).

riticksingh (commented):
Done

riticksingh (commented):
Done

script/proto_generate.sh
```diff
@@ -0,0 +1,19 @@
+git --version 2>&1 >/dev/null
+GIT_IS_AVAILABLE=$?
```
bobstrecansky, line 2: You don't need to use the previous output here.  You can just do:
`GIT_IS_AVAILABLE=$(git --version 2>&1 >/dev/null)`

script/proto_generate.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ $GIT_IS_AVAILABLE -eq 0 ]
+then
+	protoc --version 3>&1 >/dev/null
```
bobstrecansky, line 5: I think this is supposed to be `2>&1`.

script/proto_generate.sh
```diff
@@ -0,0 +1,19 @@
+git --version 2>&1 >/dev/null
+GIT_IS_AVAILABLE=$?
+if [ $GIT_IS_AVAILABLE -eq 0 ]
+then
```
bobstrecansky, line 4: You don't need this if / then / then logic.  You can just check both dependencies at once.  Something along the lines of:

```
GIT_IS_AVAILABLE=$(git --version 2>&1 >/dev/null)
PROTOC_IS_AVAILABLE=$(protoc --version 2>&1 >/dev/null)

if $GIT_IS_AVAILABLE && $PROTOC_IS_AVAILABLE
then
    PERFORM LOGIC
else
    ERROR MESSAGING FOR BOTH GIT AND PROTOC NEED TO BE INSTALLED
fi


```

bobstrecansky (changes requested):

script/proto_generate.sh
```diff
@@ -0,0 +1,12 @@
+GIT_IS_AVAILABLE=$(git --version 2>&1 >/dev/null)
+PROTOC_IS_AVAILABLE=$(protoc --version 2>&1 >/dev/null)
+if  $GIT_IS_AVAILABLE && $PROTOC_IS_AVAILABLE 
+then
+	cd ..
```
bobstrecansky, line 5: instead of `..`, we could also just use 

`cd $(git rev-parse --show-toplevel)`

To get to the root of our git repo.  This is a pretty handy command to have in your toolset!

script/proto_generate.sh
```diff
@@ -0,0 +1,12 @@
...
+	git clone https://github.com/open-telemetry/opentelemetry-proto
```
bobstrecansky, line 6: This will work the first time, but if you attempt to clone a repo that is already cloned, it won't take any action.  We either need to remove this directory at the end of this script, or make changes accordingly.  What do you think we should do here?

bobstrecansky (approved):
Thanks for fixing these things!  A couple more things I noticed but no real blockers.  What do you think of these recommendations?

script/proto_generate.sh
riticksingh: According to me there should be two possible ways :
1)Delete the folder if there is any existing opentelemetry-proto or proto folder and create new one
2)or to just skip if there is already exist that folder

bobstrecansky: Skipping if the folder exists could be a mistake - an update to the protoc upstream wouldn't translate well in this scenario.

bobstrecansky (approved):

bobstrecansky:
@riticksingh and I talked about this in DM and he has a couple more changes he'd like to make.

bobstrecansky (approved):

PR #215, opened 2020-11-12
Associating Spans with their TracerProvider

zsistla:
* Added functionality to associate a traceprovider resource on a tracer and a tracer resource with a span.

* Added a `defaultResource` method to `ResourceInfo`.

* Added test cases to ‘ResourceTest.php`, `TracerProviderTest.php`, `TracingTest.php`.

### DIFF
--- a/sdk/Trace/Span.php
+++ b/sdk/Trace/Span.php
@@ -21,6 +21,9 @@ class Span implements API\Span
     private $statusCode;
     private $statusDescription;
 
+    /**
+     * @var ResourceInfo
+     */
     private $resource; // An immutable representation of the entity producing telemetry.
 
     private $attributes;
--- a/sdk/Trace/Span.php
+++ b/sdk/Trace/Span.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Sdk\Trace;
 
+use OpenTelemetry\Sdk\Resource\ResourceInfo;
 use OpenTelemetry\Trace as API;
 
 class Span implements API\Span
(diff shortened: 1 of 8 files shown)


sdk/Trace/Span.php
```diff
@@ -62,6 +70,11 @@ public function __construct(
         $this->events = new Events();
     }
 
+    public function getResource(): ResourceInfo
+    {
+        return clone $this->resource;
```
bobstrecansky, line 75: What's the reason behind returning a clone of `$this->resource`?

sdk/Trace/Tracer.php
```diff
@@ -167,7 +168,10 @@ public function getSpans(): array
     {
         return $this->spans;
     }
-
+    public function getResource(): ResourceInfo
+    {
+        return clone $this->resource;
```
bobstrecansky, line 173: same question in re: `clone`

bobstrecansky (approved):
Quick question but not a blocker.  Awesome work @zsistla !

sdk/Trace/Tracer.php
zsistla: Trying to ensure the immutableness of resource.  The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md) says `A Resource is an immutable representation of the entity producing telemetry.` Similar to how we handle the immutableness of `context` in this [getter](https://github.com/open-telemetry/opentelemetry-php/blob/master/sdk/Trace/Span.php#L67).

bobstrecansky: Ahh perfect.  Thanks for the clarification!

Issue #212, opened 2020-11-08
php-cs-fixer not being applied in CI


bobstrecansky:
It was noticed in https://github.com/open-telemetry/opentelemetry-php/pull/210 that php-cs-fixer isn't being applied on PRs.  It appears that there is a fatal in the github action.

https://github.com/open-telemetry/opentelemetry-php/pull/210/checks?check_run_id=1367036086#step:7:4

bobstrecansky:
Wondering if this should be a pre-commit hook rather than a CI check?  Might make it easier for a contributor.

PR #216, opened 2020-11-13
php-cs-fixer bugfix

bobstrecansky:
FIxes https://github.com/open-telemetry/opentelemetry-php/issues/212

This PR fixes the operation of php-cs-fixer in CI.  It was noted that php-cs-fixer wasn't running properly after the swap to Github actions, it was because of a `$TRAVIS_COMMIT_RANGE` variable that was left in the phpcs.sh script.  I determined we don't actually need that file, so I simplified the action to execute the php-cs-fixer directly.  There are 3 commits in this PR:

1) The fix: [bd6b845](https://github.com/open-telemetry/opentelemetry-php/pull/216/commits/bd6b845db22cb91b1458c367f7e0436db1fded54)

2) A purposeful failure: [8f6777b](https://github.com/open-telemetry/opentelemetry-php/pull/216/commits/8f6777b48ad9eb7138f202c8c6827cf53bc2f4a0) (In hindsight, I should have added additional failures here, happy to do so if requested)

3) A purposeful fix of that failure: [2e11a41](https://github.com/open-telemetry/opentelemetry-php/pull/216/commits/2e11a41256875f26dbbe706beaf8c805bbbb98c0) using `make style` locally.

Happy to answer any questions about this PR.

We also discussed potentially doing this as a git pre-commit check as well.  We still could, we would just need to ensure that the vendored binaries are installed (the php-cs-fixer runs from the vendored directory).


bobstrecansky:
Also tagging @nenad as he did the initial implementation.

bobstrecansky:
@morrisonlevi @beniamin for review please.

morrisonlevi (approved):
Looks good as far as I know; basically it checks all files instead of just committed ones?

bobstrecansky:
@morrisonlevi - correct.  There isn't a `TRAVIS_COMMIT_RANGE` equivalent in Github Actions yet, so we have to check them all.  I think this is okay though - this process isn't very expensive and it follows the pattern that `make style` does locally.  Happy to change it if you'd rather something else.

morrisonlevi:
It's fine; was just making sure I understood it correctly.

PR #232, opened 2020-12-17
Semantic Versioning Rationale

bobstrecansky:
Initial Semantic Versioning Rationale document.

### DIFF
--- a/docs/versioning.md
+++ b/docs/versioning.md
@@ -1,0 +1,29 @@
+### Assumptions
+This project uses [semver v2](https://semver.org/), as does the rest of OpenTelemetry.
+
+### Goals
+#### API Stability:
+* Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will function, with any SDK that has the same major version, and equal or greater minor or patch version.
+  * example: libraries that are instrumented with opentelemetry-api-trace:1.0.1 will function with SDK library opentelemetry-sdk-trace:1.11.33.
+* No existing method names or signatures will change with patch versioning.  We will only change signatures in a backwards compatible way with minor versioning.
+
+#### SDK Stability:
+* Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
+* Internal interfaces are allowed to break.
+* No existing method names or signatures will change with patch versioning.  We will only change signatures in a backwards compatible way with minor versioning.
+
+### Methods
+#### Mature signals
+* Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
+* Internal interfaces are allowed to break.
+* API modules for mature (i.e. released) signals will be transitive dependencies of the OpenTelemetry API class.
+* Methods for accessing mature APIs will be added, as appropriate to the OpenTelemetry interface.
+* SDK modules for mature (i.e. released) signals will be transitive dependencies of the SDK.
+* Configuration options for the SDK modules for mature signals will be exposed, as appropriate, on the OpenTelemetry SDK class.
+ 
+#### Immature or experimental signals
+* API namespaces for immature signals will not be transitive dependencies of the API class.
+* API namespaces will be named with an "-experimental" suffix to make it abundantly clear that depending on them is at your own risk.
+* API namespaces for immature signals will be co-versioned along with mature API namespaces.
+* The classes for immature APIs will be used as if they were mature signals. This will enable users to easily transition from immature to mature usage, without having to change imports.
+* SDK classes for immature signals will also be named with an "-experimental" suffix, in parallel to their API modules.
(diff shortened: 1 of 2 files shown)


docs/versioning.md
```diff
@@ -0,0 +1,29 @@
+### Assumptions
+This project uses [semver v2](https://semver.org/), as does the rest of OpenTelemetry.
+
+### Goals
+#### API Stability:
+* Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will function, with any SDK that has the same major version, and equal or greater minor or patch version.
```
dyladan, line 6:
```suggestion
* Once the API for a given signal (spans, logs, metrics, baggage) has been officially released, that API module will function with any SDK that has the same major version and an equal or greater minor version.
```

https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html

patch versions should always be compatible if major/minor are equal

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+  * example: libraries that are instrumented with opentelemetry-api-trace:1.0.1 will function with SDK library opentelemetry-sdk-trace:1.11.33.
+* No existing method names or signatures will change with patch versioning.  We will only change signatures in a backwards compatible way with minor versioning.
```
dyladan, line 8:
```suggestion
* No existing method names or signatures will change with a patch release.
* Method signatures may only change in a backwards compatible way with a minor release.
```

I think this wording is clearer.

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+
+#### SDK Stability:
+* Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
+* Internal interfaces are allowed to break.
```
dyladan, line 12:
```suggestion
* Internal interfaces may break with a minor version change.
```

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+* No existing method names or signatures will change with patch versioning.  We will only change signatures in a backwards compatible way with minor versioning.
```
dyladan, line 13:
```suggestion
* No existing public method names or signatures will change with a patch release.
* Public method signatures may only change in a backwards compatible way with a minor release.
```

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Methods
+#### Mature signals
+* Public portions of the SDK (constructors, configuration, end-user interfaces) must remain backwards compatible.
+* Internal interfaces are allowed to break.
```
dyladan, line 18: Break when? What defines an internal interface?

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+* API modules for mature (i.e. released) signals will be transitive dependencies of the OpenTelemetry API class.
+* Methods for accessing mature APIs will be added, as appropriate to the OpenTelemetry interface.
```
dyladan, line 20:
```suggestion
* Methods for accessing mature APIs will be added, as appropriate, to the OpenTelemetry interface.
```

Is this "OpenTelemetry interface" API or SDK?

docs/versioning.md
```diff
@@ -0,0 +1,29 @@
...
+* SDK modules for mature (i.e. released) signals will be transitive dependencies of the SDK.
+* Configuration options for the SDK modules for mature signals will be exposed, as appropriate, on the OpenTelemetry SDK class.
+ 
+#### Immature or experimental signals
+* API namespaces for immature signals will not be transitive dependencies of the API class.
+* API namespaces will be named with an "-experimental" suffix to make it abundantly clear that depending on them is at your own risk.
+* API namespaces for immature signals will be co-versioned along with mature API namespaces.
```
dyladan, line 27: What does co-versioned mean?

dyladan (commented):
Most of this is nit-picky. Great job!

docs/versioning.md
bobstrecansky: In PHP:

from [interfaces](https://www.php.net/manual/en/language.oop5.interfaces.php):
Interfaces are defined in the same way as a class, but with the interface keyword replacing the class keyword and without any of the methods having their contents defined.

Internal interfaces would be defined to me as a non-client facing part of the API.  Perhaps this bullet point isn't necessary.

docs/versioning.md
bobstrecansky: This should be referring to the API in this case.  I can add a clarifying comment here.

docs/versioning.md
bobstrecansky: I'm actually not sure this is applicable, I think this might be worth removing.

docs/versioning.md
dyladan: I still think the difference between internal and otherwise (public? external?) isn't specifically defined in the document. Is it impossible for a consuming client to reach them? If so, I think it's irrelevant for this doc. If not, then it should be made clear to consumers which interfaces are meant to be internal (and therefore may break with non-major versions). I may be missing some php-specific piece of lore here though.

dyladan (approved):
Looks good to me modulo the one comment I made

beniamin (approved):

Issue #217, opened 2020-11-18
W3C Trace Context Test Service


bobstrecansky:
After finishing the W3C Trace Context work, we'd like to validate the changes we make don't break the W3C spec.  The w3c provides the W3C Distributed Tracing Validation Service, which allows us to test against a service.  The service and definition can be found at the following link:

https://github.com/w3c/trace-context/tree/master/test

Success criteria - create a `Make` task that allows validation with this service. We'll add it into the Github actions workflow after we've determined it's necessary.

@Fahmy-Mohammed mentioned wanting to work on this task.

bobstrecansky:
@Fahmy-Mohammed is working on this.

Fahmy-Mohammed:
I'm working on this

PR #233, opened 2021-01-07
Add Trace Context Tracestate header propagation Part 1

Fahmy-Mohammed:
This adds the first part of `Tracestate` header propagation according to the [W3C Trace Context spec](https://www.w3.org/TR/trace-context/#tracestate-header). The following [operations](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracestate) on the `Tracestate` header are also included:

- Get value for a given key
- Add a new key/value pair
- Update an existing value for a given key
- Delete a key/value pair

**Note**: This does not yet include a few more pieces of work that will come in a follow-up PR. (These were split out of this PR to make it a bit easier to review):

1. [Further validation of tracestate limits (max 512 characters)](https://www.w3.org/TR/trace-context/#tracestate-limits).
2. [Validate new key/value entries and ensure they are placed at the beginning of the header](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field).
3. [Ensure vendors overwrite their previous entry upon reentry to their tracing system](https://www.w3.org/TR/trace-context/#combined-header-value).

After the above work is merged and once #217 is complete (currently in-flight), it will likely expose some edge cases that have yet to be accounted for. Those will be fixed in a follow-up PR.

### DIFF
--- a/api/Trace/SpanContext.php
+++ b/api/Trace/SpanContext.php
@@ -11,7 +11,7 @@ interface SpanContext
     public function getTraceId(): string;
     public function getSpanId(): string;
     public function getTraceFlags(): int;
-    public function getTracestate(): array;
+    public function getTraceState(): ?TraceState;
     public function isValidContext(): bool;
     public function isRemoteContext(): bool;
     public function isSampled(): bool;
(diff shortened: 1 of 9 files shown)


bobstrecansky (approved):
LGTM.  Will let other maintainers take a look and merge tomorrow if no complaints.

PR #234, opened 2021-01-08
Update ParentBased sampler according to specs

lalex:
This updates `PasentBased` sampler according to the current [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/a72807982622260a1a56e106f1f943a5f330894d/specification/trace/sdk.md#parentbased).
`ParentBased` sampler is a composite and it delegates a decision based on the parent `Span` sampled flag.


bobstrecansky (approved):
This looks great to me.  Thank you @lalex !  I'll let the other approvers take a look and merge tomorrow if no one has objections.

PR #248, opened 2021-01-31
TraceProvider lookups tracers only by name, instead of name and version

wRLSS:
This PR addresses https://github.com/open-telemetry/opentelemetry-php/issues/241

PR #245, opened 2021-01-22
Changing master to main in CI

bobstrecansky:
Our default branch has been renamed from `master` to `main`.  This PR makes the same update for our Github Actions for CI so it'll run on the main branch by default.

More discussion [here](https://github.com/open-telemetry/community/issues/402#issuecomment-765562382)

PR #249, opened 2021-02-01
Fixing Style from PR

bobstrecansky:
@wRLSS 's [PR](https://github.com/open-telemetry/opentelemetry-php/pull/248) had a slight style error, but it was not caught in CI because we were waiting for the [github master / main action PR](https://github.com/open-telemetry/opentelemetry-php/pull/245) to be fixed.  This PR fixes up the one styling error that came about so that the CI passes.

### DIFF
--- a/sdk/Trace/TracerProvider.php
+++ b/sdk/Trace/TracerProvider.php
@@ -48,7 +48,6 @@ final class TracerProvider implements API\TracerProvider
 
     public function getTracer(string $name, ?string $version = ''): API\Tracer
     {
-
         if (!$version) {
             $version = 'unknown';
         }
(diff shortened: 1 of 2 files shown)


bobstrecansky:
@wRLSS - I was looking through this, and I noticed phan errored out as well.  the version number is an optional parameter, and that gives you a collision with sprintf.   I'm happy to change this as well if you'd like - should we make a conditional based on the presence of `version` here?  Open to other thoughts as well.

wRLSS:
That's actually weird, I just double-checked this via phpstan and the statement is evaluated correctly. 
https://phpstan.org/r/126dc1c4-e97c-4a6b-9107-1dcf89c148fa

Anyways,  I guess we can change `if` part to null coalesce, a bit simpler
`sprintf("%s@%s", $name, ($version ?? 'unknown'));`

Fahmy-Mohammed (approved):

bobstrecansky:
Updated.  The null coalescing operator made sense to me - it looked less readable at the beginning, but ✅

bobstrecansky:
@morrisonlevi @beniamin for review please.

beniamin (approved):

Issue #219, opened 2020-11-24
User Research Feedback Improvements - Missing Exception


bobstrecansky:
@dkarlovi's [User Research Document](https://docs.google.com/document/d/1n43oTVsnydmbpzvhN-5dgPAvRWKZCiugmmR_wg_8ziQ/edit#) points out that no exception is raised.

In dev mode, he mentioned that not being able to flush spans or not recording any spans should result in an exception being thrown.

This bug report came from [this comment](https://docs.google.com/document/d/1n43oTVsnydmbpzvhN-5dgPAvRWKZCiugmmR_wg_8ziQ/edit#)

bobstrecansky:
throwing exceptions:

1. NoSpansCreatedException
2. InvalidExporterEndpointException

in dev mode would definitely help

zsistla:
I'll work on this.

prondubuisi:
Hello @bobstrecansky @dkarlovi I would love to check this out, but I have some questions

While testing https://github.com/dkarlovi/opentelemetry-php-user-test with the wrong address, I had an exception

```
Ndubuisis-MacBook-Pro:otel-php-test prondubuisi$ bin/console app:hello
InvalidArgumentException {#17
  #message: "Unable to parse provided DSN"
  #code: 0
  #file: "/Users/prondubuisi/gitrepos/otel-php-test/vendor/open-telemetry/opentelemetry/contrib/Zipkin/Exporter.php"
  #line: 54
  trace: {
    /Users/prondubuisi/gitrepos/otel-php-test/vendor/open-telemetry/opentelemetry/contrib/Zipkin/Exporter.php:54 { …}
    /Users/prondubuisi/gitrepos/otel-php-test/bin/console:59 {
      ›     'My App name CLI',
      ›     'http://localhost:941277/api/v2/spans'
      › );
      arguments: {
        $name: "My App name CLI"
        $endpointUrl: "http://localhost:941277/api/v2/spans"
      }
    }
  }
}
2021-01-31T23:29:12+01:00 [critical] Uncaught Exception: Unable to parse provided DSN


```

I am wondering if we need to change this exception to `InvalidExporterEndpointException`

Also is there a way to look into docker or elsewhere within a codebase to suggest an appropriate endpoint from the SDK?
Lastly is there a way to tell if the SDK is being used in Dev Mode? or should we allow developers using the SDK to determine if they want to turn errors off or on using indicators particular to their implementation?

[...]

PR #250, opened 2021-02-03
NoSpansCreatedException

zsistla:
See issue #219 

When calling tracer functions that deal with spans, `NoSpansCreatedException` is now thrown if no spans have been created.  Previously, `NoopSpans` were being silently returned, leaving the user no way of knowing whether a span was explicitly created or not.

Regarding adding `InvalidExporterEndpointException`, it seems like the exporters should handle their own endpoint validations, like the exception generated here for an invalid endpoint:
https://github.com/open-telemetry/opentelemetry-php/blob/main/contrib/Zipkin/Exporter.php#L63

It seems like it would be an exporter case by case basis depending what they need for their endpoints.

Added test cases.
Removed one test case (testActiveRootSpanParentIsNoopSpanIfNoParentProvided
) because it didn’t actually show what the code(https://github.com/open-telemetry/opentelemetry-php/blob/main/sdk/Trace/Tracer.php#L59) was doing where if no parent was passed in, the code sends in noop, but then a boilerplate context is generated.

### DIFF
--- a/sdk/Trace/Exceptions/NoSpansCreatedException.php
+++ b/sdk/Trace/Exceptions/NoSpansCreatedException.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Trace\Exceptions;
+
+use Exception;
+
+class NoSpansCreatedException extends Exception
+{
+}
--- a/sdk/Trace/SpanOptions.php
+++ b/sdk/Trace/SpanOptions.php
@@ -12,7 +12,7 @@ final class SpanOptions implements API\SpanOptions
     /**
      * @var string
      */
-    private $name;
+    private $name = null;
     private $parent = null;
     private $attributes = null;
     private $links = null;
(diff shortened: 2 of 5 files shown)


morrisonlevi:
I don't fully understand what this is trying to solve. Can you explain more why a NoopSpan is problematic for end-users?

Also, throwing exceptions in telemetry code can be undesirable, because if left unhandled it will crash (or at least prematurely end). As much as possible, we don't want issues in telemetry code to affect the main app.

zsistla:
Hi @morrisonlevi ,
Thanks for feedback and good to see you pop in again :).

This was in response to #219 , but PHP Otel doesn't have a `dev mode` so the implementation went straight in with no dev mode toggle.

After addressing the concerns in issue #219 , I tend to agree with you.  I put the PR up to get discussion on it, but the silent Noop functionality seemed a lot cleaner.

Without a `dev mode` to filter the exception, throwing the exception seems to make it harder for the end user.  
I'd vote that unless we get implement/activate `dev mode`, the exception is less usable and we should table this PR and issue #219 for when we can get that mode.

@bobstrecansky what are you thoughts?

bobstrecansky:
My opinion is that a dev mode would be a very nice addition to this project.  I think we should probably open an issue about that and discuss how we want to execute that long term.

I do think the functionality in this PR would benefit more from a dev mode environment; a noop span does offer a lot of functionality.

Perhaps our next step is to discuss dev mode for this project and weave this into that process?

bobstrecansky:
We are planning on working on some core functionality before we start to implement a dev mode.  Closing for now.

PR #252, opened 2021-02-16
Add IdGenerator interface

lalex:
Add SDK IdGenerator interface and default RandomIdGenerator implementation.

Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/03a6f0fd9d1d2f5e08858b102ba620ebf3e4a6fa/specification/trace/sdk.md#id-generators

### DIFF
--- a/sdk/Trace/IdGenerator.php
+++ b/sdk/Trace/IdGenerator.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Trace;
+
+interface IdGenerator
+{
+    public function generateTraceId(): string;
+
+    public function generateSpanId(): string;
+}
--- a/sdk/Trace/RandomIdGenerator.php
+++ b/sdk/Trace/RandomIdGenerator.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Trace;
+
+class RandomIdGenerator implements IdGenerator
+{
+    private const TRACE_ID_HEX_LENGTH = 32;
+    private const SPAN_ID_HEX_LENGTH = 16;
+
+    public function generateTraceId(): string
+    {
+        return $this->randomHex(self::TRACE_ID_HEX_LENGTH);
+    }
+
+    public function generateSpanId(): string
+    {
+        return $this->randomHex(self::SPAN_ID_HEX_LENGTH);
+    }
+
+    private function randomHex(int $hexLength): string
+    {
+        try {
+            return bin2hex(random_bytes(intdiv($hexLength, 2)));
+        } catch (\Throwable $e) {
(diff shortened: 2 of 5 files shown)


sdk/Trace/IdGenerator.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Trace;
+
+interface IdGenerator
+{
+    public function generateTraceId(): string;
```
bobstrecansky, line 9: In the spec, these are labeled `generateSpanIdBytes` and `generateTraceIdBytes`  - should we follow suit with the spec here?

sdk/Trace/RandomIdGenerator.php
```diff
@@ -0,0 +1,35 @@
...
+        try {
+            return bin2hex(random_bytes(intdiv($hexLength, 2)));
+        } catch (\Throwable $e) {
+            return $this->fallbackAlgorithm($hexLength);
+        }
+    }
+
+    private function fallbackAlgorithm(int $hexLength): string
+    {
+        return substr(str_shuffle(str_repeat('0123456789abcdef', $hexLength)), 1, $hexLength);
```
bobstrecansky, line 33: I think this is fine for now, but we may want to consider another option here in the future.  I think it'd be a very strange use case where `bin2hex` doesn't return a correct value.

bobstrecansky (commented):
A couple small questions and a couple psalm errors, but I think this PR is off to a great start!

sdk/Trace/IdGenerator.php
lalex: In the spec they show an example from Java implementation, where they also have a dedicated classes `TraceId` and `SpanId` for this. I don't think we have `bytes` here in PHP yet, for now Trace ID and Span ID are plain strings. In such case more suitable name is `generateTraceIdString` which sounds too obvious.

sdk/Trace/RandomIdGenerator.php
lalex: The fallback method guards a rare case when `random_bytes()` may throw an error (https://www.php.net/manual/en/function.random-bytes.php).
The implementations of this method was copied from the [existing code](https://github.com/open-telemetry/opentelemetry-php/blob/96d0feb2d23dc9c42aef9587f4260a9e174a05bb/sdk/Trace/SpanContext.php#L186).

lalex:
It seems the psalm errors are not related to the changes from the PR. I think it's better to fix it in separate PR.

sdk/Trace/RandomIdGenerator.php
bobstrecansky: Good enough for me!

bobstrecansky:
The psalm errors seem to be related to https://github.com/vimeo/psalm/issues/5203

I'll add a PR to address this.

bobstrecansky (approved):

PR #253, opened 2021-02-17
Psalm Fix

bobstrecansky:
@lalex's [PR](https://github.com/open-telemetry/opentelemetry-php/pull/252) this week introduced a Psalm release that caused a test failure.

This fixes the template annotation that (I think) needs to happen to follow Psalm's directives.

Related github issue: https://github.com/vimeo/psalm/issues/5203

PR #255, opened 2021-02-19
Feature/add record exception

prondubuisi:
Add Implementation for recordException
Add test to check if recordException event is registered on noopspan

Are there any examples that I should try this event out with?
Are there any tests I could add for this event?

### DIFF
--- a/api/Trace/Span.php
+++ b/api/Trace/Span.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Trace;
 
+use Exception;
+
 interface Span extends SpanStatus, SpanKind
 {
     public function getSpanName(): string;
@@ -57,6 +59,13 @@ interface Span extends SpanStatus, SpanKind
      * @return Span Must return $this to allow setting multiple links at once in a chain.
      */
     public function addLink(SpanContext $context, ?Attributes $attributes = null): Span;
+
+    /**
+     *
+     * @param Exception $exception
+     * @return Span Must return $this to allow setting multiple attributes at once in a chain.
+     */
+    public function recordException(Exception $exception): Span;
 
     /**
      * Calling this method is highly discouraged; the name should be set on creation and left alone.
--- a/examples/AlwaysOnJaegerExample.php
+++ b/examples/AlwaysOnJaegerExample.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 require __DIR__ . '/../vendor/autoload.php';
 
+use Exception as Exception;
 use OpenTelemetry\Contrib\Jaeger\Exporter as JaegerExporter;
 use OpenTelemetry\Sdk\Trace\Attributes;
 use OpenTelemetry\Sdk\Trace\Clock;
@@ -55,6 +56,9 @@ if (SamplingResult::RECORD_AND_SAMPLED === $samplingResult->getDecision()) {
         $span->addEvent('generated_session', $timestamp, new Attributes([
             'id' => md5((string) microtime(true)),
         ]));
+
+        $exception  = new Exception('Record event test exception');
+        $span->recordException($exception);
 
         $tracer->endActiveSpan();
     }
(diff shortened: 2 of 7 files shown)


bobstrecansky:
@prondubuisi - those CI errors are unrelated to your PR; I've got a fix in for them in https://github.com/open-telemetry/opentelemetry-php/pull/253

prondubuisi:
Okay @bobstrecansky. I will wait for the fix and next steps after that. Thanks.

api/Trace/Span.php
```diff
@@ -58,6 +58,14 @@ public function addEvent(string $name, int $timestamp, ?Attributes $attributes =
      */
     public function addLink(SpanContext $context, ?Attributes $attributes = null): Span;
 
+    /**
+     * @param string $name
+     * @param int $timestamp
+     * @param Attributes|null $attributes
+     * @return Span Must return $this to allow setting multiple attributes at once in a chain.
+     */
+    public function recordException(string $name, int $timestamp, ?Attributes $attributes = null): Span;
```
lalex, line 67: According to [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) the Span should accept `Exception` object (`Throwable` interface) instead of just its' name. The name of the event [will be](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md#recording-an-exception) `"exception"`.

api/Trace/Events.php
```diff
@@ -15,6 +15,15 @@ interface Events extends \IteratorAggregate, \Countable
...
     public function addEvent(string $name, ?Attributes $attributes = null, ?int $timestamp = null): Events;
 
+    /**
+ * Adding an exception event should not invalidate nor change any existing iterators.
+     * @param string $name
+     * @param Attributes|null $attributes
+     * @param int|null $timestamp
+     * @return Events Must return $this to allow setting multiple attributes at once in a chain.
+     */
+    public function recordException(string $name, ?Attributes $attributes = null, ?int $timestamp = null): Events;
```
lalex, line 25: For the `Events` collection should not be difference between a regular event and exception (see comment for the `Span` interface). So the method `Events::recordException()` may be redundant.

prondubuisi: Makes sense @lalex , without an exception object, recordException is just another alias for addEvent, I will look into the links and update the PR, thanks.

prondubuisi:
Hello @bobstrecansky I can't tell why I am having CI/CD errors for this, Any help here? thanks

lalex:
> Hello @bobstrecansky I can't tell why I am having CI/CD errors for this, Any help here? thanks

It seems the operation `1 / 0` doesn't throw an `Exception` but triggers `E_WARNING` instead in PHP 7.*.

Also phpstan forbids to write erroneous code even for testing purposes. You may ask it to ignore that line with the phpdoc [annotation](https://phpstan.org/user-guide/ignoring-errors).

prondubuisi:
Super thanks @lalex for the insights, I figured the warning part, after a lot of head-scratching and Stackoverflow.
Let me work on getting my exception tests to pass.

prondubuisi:
Fixed the PHPStan error, Is there a way I can replicate the test errors locally? running `make test` works well for me

lalex:
@prondubuisi the problem is that the `DivisionByZeroError` is actually a subclass of `Error`, not of `Exception`. And both [implement](https://www.php.net/manual/en/class.throwable.php) `Throwable` interface.

prondubuisi:
Thanks, @lalex lots of takeaways from this. The links were very helpful.

bobstrecansky (approved):

examples/AlwaysOnJaegerExample.php
```diff
@@ -56,6 +56,12 @@
             'id' => md5((string) microtime(true)),
         ]));
 
+        try {
```
bobstrecansky, line 59: Did you mean to leave these in the example? Same question for the other example.

Issue #238, opened 2021-01-14
Fix default span status


jodeev:
Default span status should be `Unset`, but the current implementation uses `Ok`. 

Also the statusCodes list has been shrunk to only 3 statuses as described in the [SetStatus](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/api.md#set-status) section and the current behaviour has been changed.

**Notes**
Spans are not meant to be used to propagate information within a process. To prevent misuse, implementations SHOULD NOT provide access to a Span's attributes besides its SpanContext.

Vendors may implement the Span interface to effect vendor-specific logic. However, alternative implementations MUST NOT allow callers to create Spans directly. All Spans MUST be created via a Tracer.

With the exception of the function to retrieve the Span's SpanContext and recording status, Span Status may NOT be changed after the Span is finished.

zsistla:
I'll work on this.

zsistla:
Please see PR: https://github.com/open-telemetry/opentelemetry-php/pull/258

PR #258, opened 2021-02-24
Modified SpanStatus to conform with the feature-freeze specs.

zsistla:
1) Updated SpanStatus to adhere to the latest [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/api.md#set-status).

a) Made `UNSET` the statusCode default.
b) Removed significant amount of status codes that were removed from the [spec](https://github.com/open-telemetry/opentelemetry-specification/commit/30eb18a30dff36422de9a2a851010a87e66e4d9d#diff-1c55a70901ca0a001050f25ba8881a9b841777e6d8eaab500faf20230765ff76).
c) Only allowed `ERROR` codes to set descriptions.

The key points in the spec to notice were:
```
The Span interface MUST provide:
	•	An API to set the Status. This SHOULD be called SetStatus. This API takes the StatusCode, and an optional Description, either as individual parameters or as an immutable object encapsulating them, whichever is most appropriate for the language. Description MUST be IGNORED for StatusCode Ok & Unset values.
```
Additionally, opted for the following behavior:
```
If invalid status codes were given, the default behavior is to set status code/description to the default (UNSET).
```

2) Some refactoring in `Span.php` to actually utilize `SpanStatus` (similar to what the NoopSpan was doing) vs setting private variables code and description in the span itself.

3) Updated Tests to account for the changes.

### DIFF
--- a/api/Trace/SpanStatus.php
+++ b/api/Trace/SpanStatus.php
@@ -6,42 +6,14 @@ namespace OpenTelemetry\Trace;
 
 interface SpanStatus
 {
+    const UNSET = 'Unset';
     const OK = 'Ok';
-    const CANCELLED = 'Cancelled';
-    const UNKNOWN = 'Unknown';
-    const INVALID_ARGUMENT = 'InvalidArgument';
-    const DEADLINE_EXCEEDED = 'DeadlineExceeded';
-    const NOT_FOUND = 'NotFound';
-    const ALREADY_EXISTS = 'AlreadyExists';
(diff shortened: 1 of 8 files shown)


zsistla:
Fixes Issue #238 .

bobstrecansky (approved):

bobstrecansky (approved):

bobstrecansky:
Fixed the merge conflict with a living PR, merging!

Issue #259, opened 2021-02-24
Change SpanContext to Baggage


bobstrecansky:
We should use the standard implementation of [Baggage](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md)

@Fahmy-Mohammed mentioned he would work on this in our meeting today.

Fahmy-Mohammed:
Yes, I can work on this

jimshowalter:
@Fahmy-Mohammed Do you have a sense for how long it will take to make PHP support baggage, and do you want help?

Fahmy-Mohammed:
I'll work on this over the weekend 👍🏾

PR #264, opened 2021-02-28
Add ContextValueTrait and Scope to store in Context

lalex:
This PR proposes an idea of `trait ContextValueTrait` and `class Scope` to interact with the current `Context`.

Usage example:
```php
class Span {
  use ContextValueTrait;
}

$span = $tracer->startSpan();
$scope = Span::setCurrent($span);
try {
  $activeSpan = Span::getCurrent(); // returns $span
} finally {
  $scope->close();
}
$activeSpan = Span::getCurrent(); // returns NoopSpan
$span->end();
```

This also can be used in implementation of [Baggage API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md).

### DIFF
--- a/Context/ContextValueTrait.php
+++ b/Context/ContextValueTrait.php
@@ -9,6 +9,7 @@ trait ContextValueTrait
     /**
      * @param Context $context
      * @return mixed|null
+     * @phan-suppress PhanAbstractStaticMethodCallInTrait
      */
     public static function extract(Context $context)
     {
@@ -23,6 +24,7 @@ trait ContextValueTrait
      * @param mixed $value
      * @param Context $context
      * @return Context
+     * @phan-suppress PhanAbstractStaticMethodCallInTrait
      */
     public static function insert($value, Context $context): Context
     {
@@ -32,6 +34,7 @@ trait ContextValueTrait
     /**
      * @param mixed $value
      * @return Scope
+     * @phan-suppress PhanAbstractStaticMethodCallInTrait
      */
     public static function setCurrent($value): Scope
     {
@@ -42,6 +45,7 @@ trait ContextValueTrait
 
     /**
      * @return mixed|null
+     * @phan-suppress PhanAbstractStaticMethodCallInTrait
      */
     public static function getCurrent()
     {
--- a/Context/ContextValueTrait.php
+++ b/Context/ContextValueTrait.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Context;
+
+trait ContextValueTrait
+{
+    /**
+     * @param mixed $value
+     * @param Context|null $context
+     * @return Scope
+     */
+    public static function setCurrent($value, ?Context $context = null): Scope
+    {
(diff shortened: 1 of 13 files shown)


bobstrecansky:
For reference, this pattern was discussed as it was used in the OpenCensus PHP library: 
https://github.com/census-instrumentation/opencensus-php#usage

cc:\ @Fahmy-Mohammed, as I know he was planning on [instrumenting the baggage API from SpanContext](https://github.com/open-telemetry/opentelemetry-php/issues/259)

@lalex I think this is a very good start.  I think we can use this to start and eventually add the `closure / callable` method as well later.

lalex:
@bobstrecansky yeap, it slightly differs from OpenCensus and follows the specifications for [context interaction](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction). Not only `Tracer` needs this functionality. I think `Propagator`s may find it useful too.

beniamin:
I try to understand the logic behind the PR and I have some issues connecting the pieces together. 

Can explain a little bit what is the connection between `Context`, `SpanContext`, `ContextValueTrait` and `Span`. I have some issues to get why a Span has the ContextValue traits? 

Shouldn't these traits to be on the Context? I mean why would a Span should know how to extract or inject a new key in the Context? Shouldn't be this the responsibility of the Context itself? 

Maybe if you can point me to some specification that describes the interaction between Spans and Contexts.

lalex:
@beniamin sure, let's start from the simple one. `SpanContext` is nothing more than serializable identifier of the Span that holds TraceId, SpanId, TraceFlags, and TraceState. It can be sent over the wire between the application boundaries to keep a relation between spans in the trace.

`Context` is the in-process container that helps to propagate the currently active span between execution scopes. The span could be injected into the Context and then extracted somewhere else in the code. Context could be passed as an argument to subsequent calls or global current context could be used instead.
The `Context` doesn't care about types of stored objects.

The instance of `ContextKey` class is used as an identifier to get/set an object in the Context. The [identity operator](https://github.com/open-telemetry/opentelemetry-php/blob/07cdd5ad971dc33e8845ef6d30a24109b976d3d1/Context/Context.php#L105) is used to compare context keys. So the only instance of the `ContextKey` should be used to referrer object of the same type, spans in our case.

`ContextValueTrait` is a try to make some kind of ContextStorableInterface with the default implementation of getters/setters and the requirement to override in extending class the method that provides the `ContextKey` instance.

I'm not sure that the trait should be used in the `Span` class directly. But definitely it should not be a part of the Context namespace because it's it has implementation details which is part of SDK. May be `SpanContextManager` class could be created for this.

For reference you may check [Context Interaction](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction) and [Context](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/context.md) sections in specification.

bobstrecansky (approved):
@beniamin - is this clear to you now?  I think this is ready for merge.

bobstrecansky:
Spoke with @beniamin , this is g2g.

bobstrecansky:
Thank you @lalex !

PR #275, opened 2021-03-10
Basic example to send zipkin style data to a New Relic backend.

zsistla:
1) Added a basic zipkin to New Relic exporter to the `contrib` directory and also associated tests.
2) Can easily run from `make trace`
3) It requires an environment variable (`NEW_RELIC_INSERT_KEY`) be set to successfully run.
4) If the env var isn't set, it just skips the example.


bobstrecansky (approved):
This looks awesome, thanks @zsistla !

Fahmy-Mohammed (approved):
Looks great! Thanks for adding this!

beniamin (approved):
Nice work

PR #276, opened 2021-03-10
Experimental exporter, span converter, example, and tests.

zsistla:
1) Added an experimental New Relic exporter, span converter, example, and tests.
2) Can run from `make trace`
3) It requires an environment variable (`NEW_RELIC_INSERT_KEY`) be set to successfully run.
4) If the env var isn't set, it just skips the example.

### DIFF
--- a/contrib/Newrelic/Exporter.php
+++ b/contrib/Newrelic/Exporter.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Contrib\Newrelic;
 
-use GuzzleHttp\HandlerStack;
-use GuzzleHttp\Middleware;
 use GuzzleHttp\Psr7\Request;
 use Http\Adapter\Guzzle7\Client;
 use InvalidArgumentException;
@@ -26,6 +24,9 @@ use Psr\Http\Client\RequestExceptionInterface;
  */
 class Exporter implements Trace\Exporter
 {
+    private const DATA_FORMAT = 'newrelic';
+    private const DATA_FORMAT_VERSION_DEFAULT = '1';
+
     /**
      * @var string
      */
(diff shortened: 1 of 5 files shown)


contrib/Newrelic/Exporter.php
```diff
@@ -0,0 +1,159 @@
...
+
+        if (empty($spans)) {
+            return Trace\Exporter::SUCCESS;
+        }
+
+        $convertedSpans = [];
+        foreach ($spans as $span) {
+            array_push($convertedSpans, $this->spanConverter->convert($span));
+        }
+        $commonAttributes = ['attributes' => [ 'service.name' => $this->name,
```
beniamin, line 106: I think this 2 arrays($commonAttributes and $payload) needs some formatting. Maybe you can run `make style` to fix them for you.

contrib/Newrelic/Exporter.php
```diff
@@ -0,0 +1,159 @@
...
+            $response = $this->client->sendRequest($request);
+        } catch (RequestExceptionInterface $e) {
+            return Trace\Exporter::FAILED_NOT_RETRYABLE;
+        } catch (NetworkExceptionInterface | ClientExceptionInterface $e) {
+            return Trace\Exporter::FAILED_RETRYABLE;
+        }
+
+        $statusCode = $response->getStatusCode();
+        $reason = $response->getReasonPhrase();
+        echo "\nsendRequest response = " . $statusCode . "\n";
```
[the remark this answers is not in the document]
zsistla: Thanks for commenting @beniamin! These comments apply to the #275 as well, so I will apply them there  as well :)

That's correct,  I hesitated leaving them in, but initially left them in since it was experimental.  I'll comment out the echos, so when we have logger facilities, we can remember to get them back in.

contrib/Newrelic/Exporter.php
zsistla: Thanks for commenting @beniamin! These comments apply to the #275 as well, so I will apply them there  as well :)

`make style` didn't change the formatting, so I'll format it by hand.

bobstrecansky:
This PR looks good to go after you apply @benagricola 's requested fixes!

bobstrecansky (approved):

PR #282, opened 2021-03-13
Update SpanProcessor parameters onStart

### LINKED ISSUE #278: Update SpanProcessor parameters OnStart + OnEnd
Conform to interfaces requirements in spec for SpanProcessor, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#interface-definition

**[OnStart](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart)**
* Parameters should be:
  * `span` - a read/write span object for the started 'span'. It SHOULD be possible to keep a reference to this span object and updates to the span SHOULD be reflected in it. For example, this is useful for creating a SpanProcessor that periodically evaluates/prints information about all active span from a background thread.
  * `parentContext` - the parent `Context` of the span that the SDK determined (the explicitly passed `Context`, the current `Context` or an empty `Context` if that was explicitly requested).

**[OnEnd](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onendspan)**
* Parameters:
  * `span` - a readable span object for the ended span. Note: Even if the passed Span may be technically writable, since it's already ended at this point, modifying it is not allowed.

lalex:
Pass the parent Context to the SpanProcessor on Span start.

Fixes #278

### DIFF
--- a/tests/Sdk/Integration/SpanProcessorTest.php
+++ b/tests/Sdk/Integration/SpanProcessorTest.php
@@ -18,7 +18,7 @@ class SpanProcessorTest extends TestCase
     public function parentContextShouldBePassedToSpanProcessor()
     {
         $parentContext = new Context();
-        
+
         $spanProcessor = $this->createMock(SpanProcessor::class);
         $spanProcessor
             ->expects($this->once())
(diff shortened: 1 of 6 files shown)


bobstrecansky (approved):
LGTM, thanks @lalex !

Fahmy-Mohammed (approved):

tests/Sdk/Integration/SpanProcessorTest.php
```diff
@@ -0,0 +1,34 @@
...
+use PHPUnit\Framework\TestCase;
+
+class SpanProcessorTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function parentContextShouldBePassedToSpanProcessor()
+    {
+        $parentContext = new Context();
```
zsistla, line 20: This looks great, thanks for putting this PR up!  Would it make sense to have a test case where parentContext is null and/or not passed in?

lalex: Yes, I've added a test case when `null` is passed as a parentContext argument on Span creation. The current Context is used in this case and passed to the SpanProcessor.

PR #286, opened 2021-03-16
Fix memory leak in Exporter

lalex:
Remove an unused history middleware from the default HTTP client in Exporter which causes a memory leak.


bobstrecansky (approved):

Fahmy-Mohammed (approved):

zsistla (approved):
Great find, thanks!

PR #289, opened 2021-03-18
Update sampling behavior

### LINKED ISSUE #268: Update sampling behavior to conform to spec
These gaps have been identified in our sampling:
* Default sampler should be ParentBasedSampler as described in [SDK Span Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/sdk.md#sdk-span-creation).
* The flag combination `SampledFlag == true` and `IsRecording == false` could cause gaps in the distributed trace, and because of this, our OpenTelemetry API MUST NOT allow this combination.
* SamplingResult
   * Sampling Decision values has been changed to: DROP, RECORD_ONLY, RECORD_AND_SAMPLE
must contain the Tracestate
   * Does not contain Links --> there may be an error in specs???

lalex:
* Change shouldSample() method signature to accept the Context
* Default sampler is ParentBased(root=AlwaysOn)
* Avoid to start sampled and non-recording span, return NooSpan
* Change names of SamplingResult constants: DROP, RECORD_ONLY, RECORD_AND_SAMPLE
* Return an updated TraceState in SamplingResult (parent TraceState by default)
* Merge Attributes from the SamplingResult with the Attributes provided in the startSpan method

Fixes #268

### DIFF
--- a/examples/AlwaysOffTraceExample.php
+++ b/examples/AlwaysOffTraceExample.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 require __DIR__ . '/../vendor/autoload.php';
 
+use OpenTelemetry\Context\Context;
 use OpenTelemetry\Sdk\Trace\Attributes;
 use OpenTelemetry\Sdk\Trace\Clock;
 use OpenTelemetry\Sdk\Trace\Sampler\AlwaysOffSampler;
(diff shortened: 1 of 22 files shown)


bobstrecansky (approved):
Awesome work as always @lalex !

bobstrecansky:
@benagricola @zsistla @Fahmy-Mohammed @morrisonlevi for review please!

Fahmy-Mohammed (approved):
It seems that the `tracer` test coverage could use a couple more test cases but this isn't critical. Everything looks great otherwise!

lalex:
@Fahmy-Mohammed right, a couple of tests are needed. I've added one more test to cover the case of overridden `TraceState`. Since `Attributes` are not adopted enough in `Span` creation I left this case for the future.

PR #304, opened 2021-04-06
Updated components of automated examples.

zsistla:
* Added grpc capabilities to the PHP container dockerfile.
* Added an otel collector (exports to `logging`) to `docker-compose.yaml` to exercise the OTLPGrpc Exporter.
* Edited Makefile to call the OTLPGrpc example.
* Added a file needed to configure the otel collector to send traces directly to logging.

This will automatically spin up an `otel/opentelemetry-collector-contrib` otel [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib) and functional backends to easily demonstrate the newest functionality of the OTLP Grpc Exporter once #272  is merged. This streamlines the process by adding it to the examples in the Makefile.  If users want to experiment with their own credentials/otel backends, they just need to modify `files/collector/otel-collector-config.yml` with the specific information and that will be automatically spun up the next time they run 'make trace'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ trace examples: FORCE
         # The following examples do not use the DC_RUN_PHP global because they need environment variables.
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnNewrelicExample.php
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnZipkinToNewrelicExample.php
+i	docker-compose run -e OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317 --rm php php ./examples/AlwaysOnOTLPGrpcExample.php
+
 
 metrics-prometheus-example:
 	@docker-compose -f docker-compose.prometheus.yaml up -d web
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ trace examples: FORCE
         # The following examples do not use the DC_RUN_PHP global because they need environment variables.
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnNewrelicExample.php
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnZipkinToNewrelicExample.php
-i	docker-compose run -e OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317 --rm php php ./examples/AlwaysOnOTLPGrpcExample.php
+	docker-compose run -e OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317 --rm php php ./examples/AlwaysOnOTLPGrpcExample.php
 
 
 metrics-prometheus-example:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -22,6 +22,8 @@ trace examples: FORCE
         # The following examples do not use the DC_RUN_PHP global because they need environment variables.
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnNewrelicExample.php
 	docker-compose run -e NEW_RELIC_ENDPOINT -e NEW_RELIC_INSERT_KEY --rm php php ./examples/AlwaysOnZipkinToNewrelicExample.php
+i	docker-compose run -e OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317 --rm php php ./examples/AlwaysOnOTLPGrpcExample.php
```
bobstrecansky, line 25: Is this `i` errant?

docker-compose.yaml
```diff
@@ -6,6 +6,8 @@ services:
       dockerfile: docker/Dockerfile
     volumes:
     - ./:/usr/src/myapp
+    depends_on:
+      - otel-collector
```
bobstrecansky, line 10: I recognize this might be a slight bit of a bikeshedding discussion, but should our default depend on the collector?  I almost think it would make more sense to have a separate docker-compose file for something that depends on the connector.  I'm not strongly opinionated on this.

bobstrecansky (commented):
2 quick questions, but I'm ready to merge this when you are!

docker-compose.yaml
zsistla: Happy to change this.  I initially thought to use this `docker-compose` because even though the collector isn't configured for it now, it could be updated to send to the zipkin and jaeger services as well, but there is a good argument for not having the php service be dependent for other cases.  I created the new yaml which also diversifies our examples so overall a good thing :)

files/collector/otel-collector-config.yml
```diff
@@ -0,0 +1,31 @@
...
+      grpc:
+  zipkin:
+
+exporters:
+  zipkin:
+    endpoint: "http://zipkin:9411/api/v2/spans"
++  newrelic:
++    apikey: 'ADD_A_LICENSE_KEY_HERE'
++    timeout: 30s
++    spans_host_override: "trace-api.newrelic.com"
```
SeanHood, line 13: Should these `+`'s be here?

bobstrecansky: Appears to be a copy from a git diff.  I've done this one too many times too.

bobstrecansky: I think this is just from a `git diff` - agreed they should be removed.

Viewing the raw file makes things slightly more clear:
https://github.com/open-telemetry/opentelemetry-php/blob/d24984cfb72e4c02016fef3637842948344a2b84/files/collector/otel-collector-config.yml#L10-L13

bobstrecansky (approved):
Minus the `git diff` change in https://github.com/open-telemetry/opentelemetry-php/blob/d24984cfb72e4c02016fef3637842948344a2b84/files/collector/otel-collector-config.yml#L10-L13 I think we are good to go here!

Thanks for your work on this @zsistla !

files/collector/otel-collector-config.yml
zsistla: :) thanks for catching this!!!!

docker-compose-collector.yaml
```diff
@@ -0,0 +1,34 @@
...
+    volumes:
+      - ./files/collector/otel-collector-config.yml:/etc/otel-collector-config.yml
+    ports:
+      - "1888:1888"   # pprof extension
+      - "8888:8888"   # Prometheus metrics exposed by the collector
+      - "8889:8889"   # Prometheus exporter metrics
+      - "13133:13133" # health_check extension
+      - "9411"   # Zipkin receiver
+      - "4317:4317"        # OTLP gRPC receiver
+      - "55680:55679" # zpages extension
```
Fahmy-Mohammed, line 34: The comments might need a bit of formatting here but that is a tiny nit that you can totally ignore!

Fahmy-Mohammed (approved):
Looks great!

PR #305, opened 2021-04-09
Update automated collector example to be cache independent and add children spans to example.

zsistla:
After doing a docker system prune, I realized the previous solution had been dependent on cached info.

Here's an independent version.

* Made independent of docker cache... :)    
* Ignore docker-compose orphans.
* Made `make collector` section in Makefile since the collector related tests were previously changed to have their own docker-compose they needed a docker-compose up.
 * updated the example file to account for an error in the shouldSample function call, changed `RECORD_AND_SAMPLED` to `RECORD_AND_SAMPLE`, and added a children to a parent root span.

There are two different examples of how to make children spans.  Here's output of the 2nd method:
image


zsistla:
Added database/internal/external/error attributes.
Mix of SpanKind::KIND_CLIENT and SpanKind::KIND_SERVER

check out the cool data:

image

PR #311, opened 2021-04-22
Feature/make guides accessible

### LINKED ISSUE #310: Make User guides more accessible
Images for the user guides are quite large, also there is no mention of the user guides in the readme file.

prondubuisi:
Fixes #310 
Reduce image size for user guides
Mention user guides in readme


bobstrecansky (approved):
Thanks @prondubuisi !

Issue #321, opened 2021-05-04
TypeError from calling TraceContext::extract(getallheaders(), ...)


raythang:
### Describe your environment
- php (fpm-fcgi) 7.3.21
- nginx 1.15.6

### Steps to reproduce
```php
<?php
use OpenTelemetry\Sdk\Trace\PropagationMap;
use OpenTelemetry\Sdk\Trace\TraceContext;

$carrier = getallheaders();
$map = new PropagationMap();

$context = TraceContext::extract($carrier, $map);  // <--- this call induces the problem
```

### What is the expected behavior?
No TypeError is raised, and the context is successfully extracted.

### What is the actual behavior?
The `extract` call above induces the following:
```
Uncaught TypeError: strtolower() expects parameter 1 to be string,
int given in /vendor/open-telemetry/opentelemetry/sdk/Trace/PropagationMap.php:28
```

### Additional context

`getallheaders` returns an associative array of request headers (name => value)
https://www.php.net/manual/en/function.getallheaders.php 

However, it appears that `getallheaders` may return *integer* entries for headers.  Consider:
```
[rthang@rthang.dev current]$ cat echo-getallheaders.php
<?php

echo var_export(getallheaders(), true);
echo PHP_EOL;

[rthang@rthang.dev current]$ curl localhost/echo-getallheaders.php -H '1: numeric-headers-for-the-loss'
array (
  1 => 'numeric-headers-for-the-loss',
  'Accept' => '*/*',
  'Host' => 'localhost',
  'User-Agent' => 'curl/7.29.0',
  'Content-Type' => 'application/x-www-form-urlencoded',
  'X-Forwarded-For' => '127.0.0.1',
  'Content-Length' => '',
)
```

This conflicts with assumption by `TraceContext::extract` that the `$carrier`'s keys are all strings, triggering the TypeError from `strtolower` that we see today.  https://github.com/open-telemetry/opentelemetry-php/blob/4c0da3dec054d6fd0d4b570ef1e5f11fe2d7fd88/sdk/Trace/PropagationMap.php#L28

Fahmy-Mohammed:
Hi @raythang,

Thanks for highlighting this error. I will work on a fix for this today.

raythang:
Thanks @Fahmy-Mohammed lmk if you need more detail.

Fahmy-Mohammed:
#329 should fix this issue. Please us know if there are any other issues related to this.

PR #329, opened 2021-05-07
Trace context key extraction fix

Fahmy-Mohammed:
There are instances where a `$carrier` containing both string and integer header keys was generating an error when it was extracting a TraceContext header. This is now fixed by ensuring the key is a string before converting it to lowercase. See #321  for more information.

```
Uncaught TypeError: strtolower() expects parameter 1 to be string,
int given in /vendor/open-telemetry/opentelemetry/sdk/Trace/PropagationMap.php:28
```

### DIFF
--- a/tests/Sdk/Unit/Trace/PropagationMapTest.php
+++ b/tests/Sdk/Unit/Trace/PropagationMapTest.php
@@ -56,7 +56,7 @@ class PropagationMapTest extends TestCase
 
         $this->assertNull($map->get([], 'a'));
         $this->assertNull($map->get($carrier, 'b'));
-         
+
         $this->expectException(\InvalidArgumentException::class);
         $value = $map->get('invalid carrier', 'a');
     }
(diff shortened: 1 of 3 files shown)


tests/Sdk/Unit/Trace/PropagationMapTest.php
```diff
@@ -125,4 +125,20 @@ public function testGetArrayValuesFromCarrier()
...
+     * @test
+     */
+    public function testGetNumericalKeyFromCarrier()
+    {
+        // Carrier contains an array as one of the values
+        $carrier = [
+            1 => ['alpha'],
+            'b' => 'bravo',
+        ];
+        var_dump($carrier);
```
bobstrecansky, line 139: May want to remove these `var_dump` instances.

Fahmy-Mohammed: Yeah just saw that. It's already out 👍🏾

PR #331, opened 2021-05-08
Add working example and tests for improvements

prondubuisi:
This is a sample pull request for PSR Improvements.

## Composer requirements
nyholm/psr7

This pulls in https://github.com/php-http/message-factory, which is an interface package for Requests and Streams, then provides an implementation for these interfaces.

## Code Changes
There is a code change with the ZipkinExporter which now has four requirements, clientName, url, client, clientInterface, streamInterface.

This code change affects the tests and export examples dependent on the `ZipkinExporter`.

This PR makes Changes to both the `JaegerExporterTest` as well as the `AlwaysOnJaegerExample.php` to meet changes with the `ZipkinExporter`

To run the JaegerExporter test `php ./vendor/bin/phpunit  --filter failsIfNotRunning  tests/Contrib/Unit/JaegerExporterTest.php`

# Questions
* How would these improvements affect the ease of using the package? As one needs to be experienced with clients as well as factory implementations to be able to meet the exporter requirements

* Are there any alternatives to using `nyholm/psr` for the Factory implementation, Is there any way we can use Guzzle?

cc @dkarlovi  @bobstrecansky


composer.json
```diff
@@ -11,7 +11,8 @@
         "php-http/guzzle7-adapter": "^0.1.1",
         "promphp/prometheus_client_php": "^2.2.1",
         "grpc/grpc": "^1.30",
-        "google/protobuf": "^v3.3.0"
+        "google/protobuf": "^v3.3.0",
+        "nyholm/psr7": "^1.4"
```
dkarlovi, line 15: This is incorrect. The correct dependency here is [`psr/http-factory-implementation`](https://packagist.org/providers/psr/http-factory-implementation) (or it might be [`psr/http-message-implementation`](https://packagist.org/providers/psr/http-message-implementation) :thinking: )

You can add this package to `require-dev` to make the tests run.

dkarlovi (changes requested):
Small change, otherwise LGTM.

composer.json
```diff
@@ -11,7 +11,8 @@
         "php-http/guzzle7-adapter": "^0.1.1",
```
dkarlovi, line 11: The Guzzle, PSR dependencies here can be removed and replaced with [`psr/http-client-implementation`](https://packagist.org/providers/psr/http-client-implementation).

Can add Guzzle into `require-dev` for the tests.

dkarlovi (changes requested):
Spoke too soon, still a piece missing.

dkarlovi:
I want to clarify this change: with the changes proposed to `composer.json`, the package will expect to find *some* implementation of the HTP client and HTTP factories.

If you're already using Symfony, you're likely to have the Symfony HTTP client installed, maybe even the Nyholm PSR7 package too In that case you'll just install this package and it will find those dependencies satisfied.

```
# works, it already has the deps satisfied
composer require open-telemetry/opentelemetry
```

If you don't have anything, you'll need to specify the desired HTTP layer with the package, so you'll do something like

```
# doesn't work
composer require open-telemetry/opentelemetry

# works
composer require open-telemetry/opentelemetry symfony/http-client nyholm/psr7

# works
composer require open-telemetry/opentelemetry guzzlehttp/guzzle nyholm/psr7

# works
composer require open-telemetry/opentelemetry guzzlehttp/guzzle guzzlehttp/psr7

# works
composer require open-telemetry/opentelemetry symfony/http-client guzzlehttp/psr7
```

Think of the `-implementation` packages like "interfaces" (which is a bit confusing, because the actual PHP interfaces are in the base package, without the `-implementation`), but the point is you're asking for some (any!) implementation of those interfaces, you don't care which.

prondubuisi:
Thanks for the reviews @dkarlovi, I will be checking out the different implementations and will get back with more questions D:

prondubuisi:
Some Combinations above do not work, as they miss these classes
```
use Psr\Http\Message\RequestFactoryInterface;
use Psr\Http\Message\StreamFactoryInterface;
```
Also,  there is no factory implementation for them that doesn't require passing something when using the new Constructor()

Except I am missing something with the usage 

```
composer require open-telemetry/opentelemetry symfony/http-client guzzlehttp/psr7

composer require open-telemetry/opentelemetry guzzlehttp/guzzle guzzlehttp/psr7(I think Guzzle comes with this out of the box)
```

Specifically, the guzzlehttp/psr7 namespace has no InterfaceFactory or InterfaceFactory implementation

An example or snippet using any of the above combinations will be very helpful. Thanks

dkarlovi:
@prondubuisi see this:
https://github.com/guzzle/psr7/blob/master/src/HttpFactory.php

dkarlovi:
New Guzzle comes with [PSR 18](https://www.php-fig.org/psr/psr-18/) (HTTP client) implementation, but you still need [PSR 7](https://www.php-fig.org/psr/psr-7/) (HTTP messages), you generate them by using [PSR 17](https://www.php-fig.org/psr/psr-17/) (HTTP factories, generate PSR 7 objects).

This is why you `require` two separate things:
1. `psr/http-factory-implementation` (creates the HTTP request objects)
2.  `psr/http-client-implementation` (sends the HTTP requests, creates HTTP response objects)

composer.json
prondubuisi: So I remember now

> https://github.com/guzzle/psr7/blob/master/src/HttpFactory.php

Hello @dkarlovi , thanks for this, it turns out that each time I run `composer require guzzlehttp/psr7` , it installs version `1.7-dev`

I will be getting something new so I can use the Factory

prondubuisi:
> @prondubuisi see this:
> https://github.com/guzzle/psr7/blob/master/src/HttpFactory.php

Hello @dkarlovi , thanks for this, it turns out that each time I run `composer require guzzlehttp/psr7` , it installs version `1.7-dev`

I will be getting something new so I can use the Factory

prondubuisi:
> 2\. psr/http-client-implementation

prondubuisi:
> New Guzzle comes with [PSR 18](https://www.php-fig.org/psr/psr-18/) (HTTP client) implementation, but you still need [PSR 7](https://www.php-fig.org/psr/psr-7/) (HTTP messages), you generate them by using [PSR 17](https://www.php-fig.org/psr/psr-17/) (HTTP factories, generate PSR 7 objects).
> 
> This is why you `require` two separate things:
> 
> 1. `psr/http-factory-implementation` (creates the HTTP request objects)
> 2. `psr/http-client-implementation` (sends the HTTP requests, creates HTTP response objects)

This is still confusing, in terms of how I can use just these two packages `guzzlehttp/guzzle` `guzzlehttp/psr7` to have a working example. I will dig around later in the day and see what I come up with. Thanks.

dkarlovi:
@prondubuisi I've tried you branch and have changed the Composer file to look like this:

```json
{
    "name": "open-telemetry/opentelemetry",
    "description": "OpenTelemetry makes robust, portable telemetry a built-in feature of cloud-native software.",
    "type": "library",
    "license": "Apache-2.0",
    "require": {
        "php": "^7.3 || ^8.0",
        "ext-json": "*",
        "promphp/prometheus_client_php": "^2.2.1",
        "grpc/grpc": "^1.30",
        "google/protobuf": "^v3.3.0",
        "psr/http-client-implementation": "^1.0",
        "psr/http-factory-implementation": "^1.0"
    },
    "authors": [
        {
            "name": "Bob Strecansky",
            "email": "bob.strecansky@gmail.com"
        },
        {
            "name": "Dmitry Krokhin",
            "email": "nekufa@gmail.com"
        },
        {
            "name": "Levi Morrison",
            "email": "levim@php.net"
        }
    ],
    "autoload": {
        "psr-4": {
            "OpenTelemetry\\Context\\": "Context",
            "OpenTelemetry\\": "api",
            "OpenTelemetry\\Sdk\\": "sdk",
            "OpenTelemetry\\Contrib\\": "contrib",
            "Opentelemetry\\Proto\\Collector\\Trace\\V1\\": "proto/Opentelemetry/Proto/Collector/Trace/V1",
            "Opentelemetry\\Proto\\Trace\\V1\\":"proto/Opentelemetry/Proto/Trace/V1",
            "Opentelemetry\\Proto\\Common\\V1\\":"proto/Opentelemetry/Proto/Common/V1",
            "Opentelemetry\\Proto\\Resource\\V1\\":"proto/Opentelemetry/Proto/Resource/V1",
            "GPBMetadata\\Opentelemetry\\Proto\\Collector\\Trace\\V1\\": "proto/GPBMetadata/Opentelemetry/Proto/Collector/Trace/V1",
            "GPBMetadata\\Opentelemetry\\Proto\\Trace\\V1\\":"proto/GPBMetadata/Opentelemetry/Proto/Trace/V1",
            "GPBMetadata\\Opentelemetry\\Proto\\Common\\V1\\":"proto/GPBMetadata/Opentelemetry/Proto/Common/V1",
            "GPBMetadata\\Opentelemetry\\Proto\\Resource\\V1\\":"proto/GPBMetadata/Opentelemetry/Proto/Resource/V1"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "OpenTelemetry\\Tests\\": "tests/"
        }
    },
    "require-dev": {
        "phpunit/phpunit": "^9.3",
        "composer/xdebug-handler": "^1.3",
        "phan/phan": "^3.0",
        "friendsofphp/php-cs-fixer": "^2.18",
        "vimeo/psalm": "^4.0",
        "phpstan/phpstan": "^0.12.50",
        "phpstan/phpstan-phpunit": "^0.12.16",
        "psalm/plugin-phpunit": "^0.13.0",
        "guzzlehttp/guzzle": "^7.3",
        "guzzlehttp/psr7": "^2.0@RC",
        "symfony/http-client": "^5.2",
        "nyholm/psr7": "^1.4"
    }
}
```

(notice I've added two separate HTTP clients and two separate PSR7 factories)

Using with Guzzle client and Nyholm PSR7
```php
$exporter = new JaegerExporter(
    'alwaysOnJaegerExample',
    'http://jaeger:9412/api/v2/spans',
    new \GuzzleHttp\Client(),
    new \Nyholm\Psr7\Factory\Psr17Factory(),
    new \Nyholm\Psr7\Factory\Psr17Factory()
);
```

Using with Guzzle client and Guzzle PSR7:
```php
$exporter = new JaegerExporter(
    'alwaysOnJaegerExample',
    'http://jaeger:9412/api/v2/spans',
    new \GuzzleHttp\Client(),
    new \GuzzleHttp\Psr7\HttpFactory(),
    new \GuzzleHttp\Psr7\HttpFactory()
);
```

prondubuisi:
> They are not real packages, check the links to packagist above to see. Use the provided composer.json to test.

Yeah, I will be removing them from the composer.json. I tested your examples above and it works perfectly.  Thanks. Working on updating my PR  later in the day to reflect the changes.

prondubuisi:
Hello @dkarlovi, I spoke with @bobstrecansky , and we think having backward compatibility  with the package is desirable, In line with that I made changes to `composer.json` to always require 
```
        "guzzlehttp/guzzle": "^7.3",
        "guzzlehttp/psr7": "^2.0@RC"
```
We also use the Request/Stream factory interface to make for the PSR flexibility. I think that in doing so we can have the best of both worlds. What do you think?

dkarlovi:
> we think having backward compatibility with the package is desirable

I'm missing the whole conversation context there: BC with which package?

> I think that in doing so we can have the best of both worlds. What do you think?

I obviously disagree because my proposal was to remove the direct Guzzle dependency to begin with. But if you and @bobstrecansky think there's a reason for it, go for it. :+1:

prondubuisi:
Hello @dkarlovi, thanks for your feedback. I think I was unclear with my comment above. Currently, we have a lot of persons utilizing the open-telemetry PHP package. With most persons using the exporters too! If we make the exporter constructor require four arguments instead of two, we would introduce breaking changes.

On the other hand, PSR improvements will improve the adoption of the project, especially by devs who don't want behind the scene magic, and already have Application-level configurations for their responses and streams. 

So we thought it would be great to have the best of the two, that is nonbreaking changes, as well as PSR flexibility. 

So like before, if users don't specify a client, we use Guzzle by default. 

But if users specify a Client as well as Response and Stream factories, we go ahead and use those, since our exporters use factory interfaces.

dkarlovi:
@prondubuisi the issue I see here is remove one of the major advantages of this change: not relying on specific version of the dependency, the dependency itself needs to keep BC with the generic interface and can change internally all it likes.

if you explicitly request `"guzzlehttp/guzzle": "^7.3"`, that means I cannot use this library together with a package which requires Guzzle 6 (even though I could, by decorating it myself), you're back to square one and haven't really moved by a significant margin.

This is one of the major headaches I see with using explicit versions of as major packages a HTTP layer is and Composer only allowing you one copy of a package at a time. The change proposed removes that headache and allows you to get started.

I agree this creates a UX issue, I've mentioned some approaches to alleviate them in the issue, [I'm discussing it with the author of `nyholm/psr7` in another PR](https://github.com/thephpleague/oauth2-client/pull/885/files/b52ecef677bf7339b55fe242fd8ec2c25579eabe#r631355582).

Again, if you feel still explicitly requesting Guzzle is a good middle ground because of BC (which is surprising to me because the package had only one release, 0.0.1, can hardly expect BC guarantees at this point).

prondubuisi:
Hello @bobstrecansky, I think @dkarlovi has great points here, finding a middle ground brings us back to Square 1. What do you think on next steps for this?

prondubuisi:
Hello @dkarlovi @bobstrecansky, I think this is ready for review.

prondubuisi:
I think we have issues with our CI/CD @bobstrecansky

bobstrecansky:
Rerunning the jobs now to see if it was Github Actions being inconsistent.

bobstrecansky:
@dkarlovi - I believe @prondubuisi is waiting for a review here.  Would you be able to review this?

tests/Contrib/Unit/ZipkinToNewrelicExporterTest.php
```diff
@@ -4,6 +4,8 @@
 
 namespace OpenTelemetry\Tests\Contrib\Unit;
 
+use GuzzleHttp\Client;
```
dkarlovi, line 7: I'd make at least one test suite use the alternate client and/or factory implementation since they're installed.

tests/Contrib/Unit/ZipkinToNewrelicExporterTest.php
```diff
@@ -105,7 +107,7 @@ public function shouldThrowExceptionIfInvalidDsnIsPassed(string $invalidDsn)
     {
         $this->expectException(InvalidArgumentException::class);
 
-        new Exporter('test.zipkinToNR', $invalidDsn, '');
+        new Exporter('test.zipkinToNR', $invalidDsn, '', new Client(), new HttpFactory(), new HttpFactory());
```
dkarlovi, line 110: This is a shift in UX, we might make a subsequent PR where we fix it by doing some heuristic.

dkarlovi (approved):
LGTM, the comments can be fixed in subsequent PRs!

Great work @prondubuisi!

tests/Contrib/Unit/ZipkinToNewrelicExporterTest.php
prondubuisi: Looks good @dkarlovi, that would help users see the usage with other factory implementation. Would do that in an incoming example

tests/Contrib/Unit/ZipkinToNewrelicExporterTest.php
prondubuisi: How do you mean @dkarlovi, how can we do this?

dkarlovi: Something like:

```php
    public function __construct(
        $name,
        string $endpointUrl,
        string $licenseKey,
        ClientInterface $client = null,
        RequestFactoryInterface $requestFactory = null,
        StreamFactoryInterface $streamFactory = null,
        SpanConverter $spanConverter = null,
```

And then

```php
if ($client === null) {
    if (class_exists(\Http\Discovery\HttpClientDiscovery::class)) {
        $client = \Http\Discovery\HttpClientDiscovery::find();
    } elseif (class_exists(\GuzzleHttp\Client)) {
        $client = new \GuzzleHttp\Client(); // this allows for some BC
    } else {
        throw new LogicException('No HTTP client found or provided');
    }
}
```

[A similar example](https://php-http.readthedocs.io/en/latest/discovery.html#psr-17-factory-discovery) would be there for factories. I'd have this in a trait which all the exporters could then use to simplify implementation.

prondubuisi: This looks like something interesting to check out, Looking forward to the next iteration. Thanks @dkarlovi

zsistla (approved):

bobstrecansky (approved):

prondubuisi:
Hello, @bobstrecansky looks like the CI checks are failing? I have tried running Psalm locally, but I can't replicate the CI errors.

bobstrecansky:
I see the same.  I can dig into this tomorrow with you.

prondubuisi:
> I see the same. I can dig into this tomorrow with you.

This works, thanks.

PR #332, opened 2021-05-11
Change SpanContext to Baggage

bobstrecansky:
Closes https://github.com/open-telemetry/opentelemetry-php/issues/259

Replacing uses of `SpanContext` to `Baggage` as per the spec.

### DIFF
--- a/api/Trace/Span.php
+++ b/api/Trace/Span.php
@@ -9,8 +9,8 @@ use Exception;
 interface Span extends SpanStatus, SpanKind
 {
     public function getSpanName(): string;
-    public function getContext(): SpanContext;
-    public function getParent(): ?SpanContext;
+    public function getContext(): Baggage;
+    public function getParent(): ?Baggage;
 
     /**
      * Returns Epoch timestamp value (RealtimeClock) when the Span was created
@@ -54,11 +54,11 @@ interface Span extends SpanStatus, SpanKind
     public function addEvent(string $name, int $timestamp, ?Attributes $attributes = null): Span;
 
     /**
-     * @param SpanContext $context
+     * @param Baggage $context
      * @param Attributes|null $attributes
      * @return Span Must return $this to allow setting multiple links at once in a chain.
      */
-    public function addLink(SpanContext $context, ?Attributes $attributes = null): Span;
+    public function addLink(Baggage $context, ?Attributes $attributes = null): Span;
 
     /**
      *
--- a/api/Trace/SpanOptions.php
+++ b/api/Trace/SpanOptions.php
@@ -14,7 +14,7 @@ interface SpanOptions
     public function setSpanName(string $name): SpanOptions;
     /** should default to INTERNAL if not called */
     public function setSpanKind(int $spanKind): SpanOptions;
-    public function setParentContext(SpanContext $span): SpanOptions;
+    public function setParentContext(Baggage $span): SpanOptions;
     public function setParentSpan(Span $span): SpanOptions;
     public function addAttributes(Attributes $attributes): SpanOptions;
     public function addLinks(Links $links): SpanOptions;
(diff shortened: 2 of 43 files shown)


api/Trace/Span.php
```diff
@@ -9,8 +9,8 @@
 interface Span extends SpanStatus, SpanKind
 {
     public function getSpanName(): string;
-    public function getContext(): SpanContext;
-    public function getParent(): ?SpanContext;
+    public function getContext(): Baggage;
```
zsistla, line 12: Should this function name change too?

sdk/Trace/NoopSpan.php
```diff
@@ -13,7 +13,7 @@ class NoopSpan implements API\Span
 {
     use ContextValueTrait;
 
-    /** @var API\SpanContext */
+    /** @var API\Baggage */
     private $context;
```
zsistla, line 17: Just a nit, wondering if this should also be changed to `baggage` for clarity down the road (for those looking later who don't know the history of the change).

api/Trace/SpanOptions.php
```diff
@@ -14,7 +14,7 @@ interface SpanOptions
     public function setSpanName(string $name): SpanOptions;
     /** should default to INTERNAL if not called */
     public function setSpanKind(int $spanKind): SpanOptions;
-    public function setParentContext(SpanContext $span): SpanOptions;
+    public function setParentContext(Baggage $span): SpanOptions;
```
zsistla, line 17: Should it be `setParentBaggage`?

zsistla:
Thanks for tackling this, @bobstrecansky!  It seemed like a lot of references to track down.

api/Trace/Span.php
bobstrecansky: I think we could change it to `getBaggage`.

sdk/Trace/NoopSpan.php
bobstrecansky: I think this is prudent.

api/Trace/SpanOptions.php
bobstrecansky: Sounds good to me!

bobstrecansky:
I think I'll have to look more in depth into how to make these updates appropriately; they aren't just find and replace.  I'll take care of that this week.

sdk/Trace/NoopSpan.php
bobstrecansky: Thinking about this statement more, there is a difference between `Context` and SpanContext`.  Perhaps it's smart to just change `SpanContext`?

api/Trace/Span.php
bobstrecansky: Thinking about this statement more, there is a difference between `Context` and SpanContext`.  Perhaps it's smart to just change `SpanContext`?

bobstrecansky:
@zsistla @Fahmy-Mohammed @beniamin @morrisonlevi I think this is ready for merge now after some 👍🏼

Fahmy-Mohammed (approved):
Looks like there are some coverage issues, but it looks great otherwise 👍🏾

api/Trace/Span.php
zsistla: Good call.  Still says getContext in the spec :)

sdk/Trace/NoopSpan.php
zsistla: I'm okay with this staying as is given how our implementation implements them both;   although I still think changing the internal variables might add more clarity, but this is just a nit and not merge blocking.  The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md) allows for more distinct entities, but php baggage and context seem a lot more tightly coupled at this point.

zsistla:
Not sure what is going on with the coverage numbers since nothing fundamentally changed except the name.

bobstrecansky:
Yeah, I feel the same.  I'm considering removing those code coverage checks.  Merging now.

PR #336, opened 2021-05-19
Use IdGenerator to generate trace ids instead of randomHex

### LINKED ISSUE #281: TracerProvider IdGenerator parameter is not used for trace id/span id generation
**Describe your environment** 

PHP 7.3
Commit of dev-main version: https://github.com/open-telemetry/opentelemetry-php/commit/bb0bbafec1d48509171aa12be68182075ef8e459

**Steps to reproduce**
Construct TracerProvider with custom generator

`
$tracer = (new TracerProvider(null, null, new TracingIdGenerator()))
        ->addSpanProcessor(new BatchSpanProcessor($exporter, Clock::get()))
        ->getTracer('io.opentelemetry.contrib.php', 'dev-main');
`

I would expect the traceId of a new span to use the generator specified.
If the new generator uses 8 bytes hex instead of 16 i would like to see it (per example to make it W3C compatible with tracing formats of trace and span ids).
`
$clientSpan->getContext()->getTraceId();
`

16 byte hex ids returned

**Additional context**
I was investigating on how to make context propagation using  TraceContext class of the same project after using the w3c trace format.

`
 $parentContext = TraceContext::extract($carrier, new PropagationMap());
`

`
Fatal error: Uncaught InvalidArgumentException: SpanID must be exactly 8 bytes (16 chars) and at least one non-zero byte, got 0c365901539b66fcb52c24a5329de59f 
`

deki:
fixes #281

### DIFF
--- a/sdk/Trace/Tracer.php
+++ b/sdk/Trace/Tracer.php
@@ -118,7 +118,7 @@ class Tracer implements API\Tracer
         $parentContextIsNoopSpan = !$parentContext->isValid();
 
         if ($parentContextIsNoopSpan) {
-            $parentContext = $this->importedContext ?? Baggage::generate(true);
+            $parentContext = $this->importedContext ?? Baggage::fork($this->provider->getIdGenerator()->generateTraceId(), true);
         }
 
         /*


bobstrecansky (approved):

Fahmy-Mohammed (approved):
Thanks for fixing this 👍🏾

Issue #363, opened 2021-06-16
Tracer - remove tracer-scoped "active" span API in favor of global, context-scoped spans


Nevay:
**Is your feature request related to a problem?**
Some of the current tracer API is centered around tracer-scoped active spans. These methods seem to predate the addition of the global context propagation, the global context should be used instead (as already done for `::startSpan()` in #264).

**Describe the solution you'd like**
Tracer-scoped spans should be replaced with usage of global, context-scoped spans:
- `::finishSpan()` & `::deactivateActiveSpan()` are neither used nor implemented
- `::getActiveSpan()` can be replaced by `Span::getCurrent()`
- `::endActiveSpan()` can be replaced by `Span::getCurrent()->end()`
- `::setActiveSpan()` can be replaced by `Span::setCurrent()`
- `::startAndActivateSpan()` can be replaced by `::startSpan()` followed by `Span::setCurrent()`

bobstrecansky:
@Nevay - I think this can be closed?

PR #364, opened 2021-06-16
Remove not implemented Tracer::finishSpan() and ::deactivateActiveSpan()

Nevay:
These two methods are neither used, nor implemented, nor required by spec.
#363


bobstrecansky (approved):
Thanks for reading through the spec and noticing this.

PR #369, opened 2021-06-23
ensure insecure variable is converted to bool

### LINKED ISSUE #368: insecure env variable not interpreted correctlt
**Steps to reproduce**
I used the OtlpGrpc\Exporter with `OTEL_EXPORTER_OTLP_INSECURE` env variable set to `false`

The grpc connection is established using a **secure** connection.

The grpc connection is established using an **insecure** connection

**Additional context**
The code that translates the env variable is not translating the string into a boolean.

codeboten:
Added tests and refactored the code in the OtlpGrpc Exporter to ensure the environment variable for insecure was correctly used. I also fixed another bug I found, which was the incorrect $certificateFile variable was used in the case where `$this->certificateFile !== ''`.

Fixes #368


bobstrecansky (approved):
LGTM.

cc\  @SeanHood

codeboten:
> LGTM, sans the one PHPUnit test failure.
> 
> cc\ @SeanHood

Ended up resorting to reflection, as different versions of the grpc extension return `null` or not `null` for ChannelCredentials.

bobstrecansky (approved):

PR #373, opened 2021-06-28
Change Random Hex generator to public function

ohamuy:
This simply makes a random hex generator function from private to public. This was checked to be alright with maintainer @bobstrecansky


bobstrecansky (approved):
LGTM.  For context - this was part of the move to the `opentelemetry-php-contrib` repo.

PR #374, opened 2021-07-02
Span context trace context update

### LINKED ISSUE #371: Change TraceContext Propagator Error Handling
Currently the traceContext propagator `extract()` function throws exceptions which is not up to specifications. If a value cannot be parsed from the carrier, the implementation must not throw an exception and must not store a new value in the Context, in order to preserve any previously existing valid value.

This is the file the issue is in: 
https://github.com/open-telemetry/opentelemetry-php/blob/60618f1c8c5d058709993922a9eccfa8df4c99e4/sdk/Trace/TraceContext.php

The following test file should also be changed to conform to changes: 
https://github.com/open-telemetry/opentelemetry-php/blob/60618f1c8c5d058709993922a9eccfa8df4c99e4/tests/Sdk/Unit/Trace/TraceContextTest.php

The `TaceContext` should not throw any exceptions even if logs do not exist yet. Comments can be placed at every current exception to preserve the desired message but an empty or invalid trace context const should be returned with is consistent with implementation in other repositories. 

Anything that calls this function and receives an invalid `SpanContext` should also not replace the span in the active Context

The following constant would suffice to replace unparsable or invalid values which are consistent with js repository implementation:

```
const INVALID_SPANID = '0000000000000000';
const INVALID_TRACEID = '00000000000000000000000000000000';
```

For example the following could be returned if the http header happens to be missing (i.e. null):
`return SpanContext::restore(API\SpanContext::INVALID_TRACE, API\SpanContext::INVALID_SPAN, false);`
This is equvalent to the `getInvalid()` method in the `SpanContext` class definition.
 
These are currently defined in the `SpanContext` API file. This file can be imported and the constants then used:
https://github.com/open-telemetry/opentelemetry-php/blob/60618f1c8c5d058709993922a9eccfa8df4c99e4/sdk/Trace/SpanContext.php

This is the minimum necessary to remove the exceptions as something must be returned in place of invalid values but all valid values must be preserved. 

@bobstrecansky  this could be assigned to @Fahmy-Mohammed or @ohamuy, I would be happy to do it as I am doing similar work right now in the AWS X-Ray propagator

ohamuy:
Fixes #371 

This addresses all problems brought up in the above issue.

Updates the `TraceContext` file to conform to specifications. `SpanContext` file was given additional helper functions that are useful for future necessary Id validation.

Tests were updated accordingly. All tests pass.

@Fahmy-Mohammed @bobstrecansky

### DIFF
--- a/sdk/Trace/SpanContext.php
+++ b/sdk/Trace/SpanContext.php
@@ -182,14 +182,16 @@ final class SpanContext implements API\SpanContext
     /**
      * @return bool Returns a value that indicates whether a trace id is valid
      */
-    public static function isValidTraceId($traceId): bool {
+    public static function isValidTraceId($traceId): bool
+    {
         return ctype_xdigit($traceId) && strlen($traceId) === self::TRACE_LENGTH && $traceId !== self::INVALID_TRACE && $traceId === strtolower($traceId);
     }
 
     /**
      * @return bool Returns a value that indicates whether a span id is valid
      */
-    public static function isValidSpanId($spanId): bool {
+    public static function isValidSpanId($spanId): bool
+    {
         return ctype_xdigit($spanId) && strlen($spanId) === self::SPAN_LENGTH && $spanId !== self::INVALID_SPAN && $spanId === strtolower($spanId);
     }
 
@@ -197,7 +199,8 @@ final class SpanContext implements API\SpanContext
      * @return bool Returns a value that indicates whether trace flag is valid
      * TraceFlags must be exactly 1 bytes (1 char) representing a bit field
      */
-    public static function isValidTraceFlag($traceFlag): bool {
+    public static function isValidTraceFlag($traceFlag): bool
+    {
         return ctype_xdigit($traceFlag) && strlen($traceFlag) === self::TRACE_FLAG_LENGTH;
     }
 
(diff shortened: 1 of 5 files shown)


ohamuy:
I believe Psalm is failing not due to my code changes but to the GRCP Exporter and other parts in the main repo. This was tested on the main repo without my changes. Please lmk if I'm wrong!

ohamuy:
Validation of removing `preg_match` and replacing with the added fucntions in the SpanContext file was done to make sure it was more optimal. Testing the two against each other in PHP Bench the following resulted:
```

+------------------+-----------------------------+-----+------+-----+-----------+----------+--------+
| benchmark        | subject                     | set | revs | its | mem_peak  | mode     | rstdev |
+------------------+-----------------------------+-----+------+-----+-----------+----------+--------+
| IdValidatorBench | benchPregmatchValidator     | 0   | 1000 | 5   | 639.952kb | 15.363μs | ±2.56% |
| IdValidatorBench | benchFunctionComboValidator | 0   | 1000 | 5   | 639.968kb | 4.381μs  | ±0.67% |
+------------------+-----------------------------+-----+------+-----+-----------+----------+--------+
```

The new function is about 3x faster than `preg_match`

bobstrecansky:
You'll need to run a `make style` on your code to get it to pass the style test.

Once that's done, I think this looks like a fine PR.

bobstrecansky (approved):

PR #384, opened 2021-07-23
Fix merging of attributes provided within SamplingResult in Tracer

bwoebi:
This fixes processing of attributes provided in the SamplingResult. Previously they were either broken (Attribute instance as value) or not present at all.

I also added `null, $sampler` to the TracerProvider args in the examples, so that one is not confused why e.g. attributes are not applied when trying another sampler than the AlwaysOnSampler.

### DIFF
--- a/docs/laravel-integration.md
+++ b/docs/laravel-integration.md
@@ -147,11 +147,11 @@ Next, we create a trace then add processors for each trace(One for Jaeger and an
 ```php
 if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
 
-    $jaegerTracer = (new TracerProvider())
+    $jaegerTracer = (new TracerProvider(null, $sampler))
         ->addSpanProcessor(new BatchSpanProcessor($jaegerExporter, Clock::get()))
         ->getTracer('io.opentelemetry.contrib.php');
 
-    $zipkinTracer = (new TracerProvider())
+    $zipkinTracer = (new TracerProvider(null, $sampler))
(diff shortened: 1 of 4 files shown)


bobstrecansky:
@bwoebi - looks like you had a style error.  You should be able to fix this using `make style`.

bwoebi:
@bobstrecansky Done :-)

ericmustin:
👋 hello friends, just chiming in, I work over on opentelemetry-ruby and also am colleagues with @bwoebi . Not much PHP background but generally I think this change lgtm and gets the SDK closer to spec.

@bobstrecansky would you be able to  review when you have a free moment? We're keen to get this change added to the repo and have this available in an upcoming release.

Also, as a followup task,  (i don't think it should block this PR), this SDK may also be missing the sampler ability to modify tracestate, as per: 

- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampler
> A Tracestate that will be associated with the Span through the new SpanContext. If the sampler returns an empty Tracestate here, the Tracestate will be cleared, so samplers SHOULD normally return the passed-in Tracestate if they do not intend to change it.

- https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#traces .  


I haven't investigated this too closely but just putting it on the radar

bobstrecansky (approved):
@ericmustin - thanks for your note!  I'm happy with this change and am approving it.  I believe you're right with getting closer to the spec.

PR #392, opened 2021-08-05
Reintroduce InstrumentationLibrary

### LINKED ISSUE #228: Adding InstrumentationLibrary to SDK
Adding instrumentation feature to opentelemetry-php to sdk mentioned here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#additional-span-interfaces

While working for OTLP/GRPC , it requires Instrumentation library in sdk for dealing with proto files

Something like this : [opentelemetry-go reference ](https://github.com/open-telemetry/opentelemetry-go/blob/66db2d845ba72bae8fcc1b3d43965470cbe8f5c3/sdk/instrumentation/library.go)

lalex:
Fixes #228.

An `InstrumentationLibrary` is associated with the `Tracer` on every `getTracer()` call and then with its' `Span`s.

This PR adds some "todo" with the `Span::setInstrumentationLibrary()` method. It should be a part of `Span` creation method, but the change of the `Span`'s constructor signature causes huge refactoring and adds a complexity for the code review at the current PR.

### DIFF
--- a/sdk/InstrumentationLibrary.php
+++ b/sdk/InstrumentationLibrary.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk;
+
+/**
+ * Represents the instrumentation library information associated with the Tracer or Meter
+ */
+class InstrumentationLibrary
+{
+    private $name;
+
+    private $version;
+
+    public function __construct(string $name, ?string $version = null)
+    {
+        $this->name = $name;
+        $this->version = $version;
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function getVersion(): ?string
+    {
+        return $this->version;
+    }
+}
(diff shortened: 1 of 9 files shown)


bobstrecansky (approved):

bobstrecansky:
Thanks @lalex !

xinfeingxia85:
@lalex 
What is the reason for removing this code? 

```
$resource = ResourceInfo::create(  
        new Attributes(
            [
                ResourceConstants::SERVICE_NAME => $name,
                ResourceConstants::SERVICE_VERSION => $version,
                ResourceConstants::SERVICE_INSTANCE_ID => uniqid($name . $version),
            ]
        )
    );
```
This causes the previous code to be incompatible and cannot get the service name.
Thinks!

lalex:
@xinfeingxia85
This code was not correct because arguments `$name` and `$version` of the `TracertProvider::getTracer($name, $version)` method are not a name and version of the [Service](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service). These arguments should be used to create an [InstrumentationLibrary](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracer-creation). To get a service name (among other Resource attributes) use `Span::getResource()` method.

xinfeingxia85:
@lalex 
OK, thank you!

Issue #389, opened 2021-08-02
Change the OTLP HTTP port


bobstrecansky:
https://github.com/open-telemetry/opentelemetry-specification/pull/1839

We need to change the default OTLP HTTP port from 4317 to 4318

bobstrecansky:
This has been completed with #398 .

PR #398, opened 2021-08-13
Change OTLP/HTTP port (#389)

0xRenegade:


bobstrecansky (approved):

SeanHood:
This PR changed the default port for OTLP/GRPC as well as OTLP/HTTP. I think only the latter was intended to be changed.

To add confusion to the mix, https://github.com/open-telemetry/opentelemetry-specification/pull/1847 the new OTLP/HTTP port change in the spec had been reverted as it may be possible after all.

bobstrecansky:
Yeah, just saw that - we can revert this PR for now if that's cool with you @SeanHood .

SeanHood:
I think that might be best for now

bobstrecansky:
If you'll do me a huge favor and open the revert for this PR, I'll approve and merge it.

PR #406, opened 2021-09-05
Implement links

Blacksmoke16:
First pass at supporting links.

* Add links to both otlp gprc and proto exporters
  * Also includes `trace_state` information as it was needed for links
* Replace `TraceState#build` with `TraceState#__toString`
  * This felt like a more idiomatic implementation because `null` gets casted to an empty string so you can just do `(string) $context->getTraceState()` versus needing to make sure it's not `null` before calling `build`. Similarly both `null` and `''` are falsely so conditional logic would treat them both the same.
* Update `TraceContext` to not include the `tracestate` header of the value would be empty
  * The spec says empty headers should avoid being sent

Some outstanding questions

* Should there be a `links` argument on the constructor of `Span`?
  * Currently the tracer is just calling `setLinks` if it's not `null`
* It seems `OpenTelemetry\Contrib\Otlp\SpanConverter` isn't up to date as it's using `annotations` versus `attributes`? From what I know it should be the same as the protobuf implementation but just JSON encoded
  * Can save this for a follow up PR if that's not expected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ vendor
 .idea/
 coverage.clover
 tests/coverage
-.php_cs.cache
+.php-cs-fixer.cache
 
 # W3C Test Service build artifacts
 tests/TraceContext/W3CTestService/test_app
--- a/api/Trace/TraceState.php
+++ b/api/Trace/TraceState.php
@@ -55,9 +55,7 @@ interface TraceState
     public function getListMemberCount(): int;
 
     /**
-     * Build the TraceState header
-     *
-     * @return string|null
+     * Returns a string representation of this TraceSate
      */
-    public function build(): ?string;
+    public function __toString(): string;
 }
--- a/contrib/OtlpGrpc/SpanConverter.php
+++ b/contrib/OtlpGrpc/SpanConverter.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Contrib\OtlpGrpc;
 
+use function array_key_exists;
+use function hex2bin;
 use Opentelemetry\Proto;
 use Opentelemetry\Proto\Common\V1\AnyValue;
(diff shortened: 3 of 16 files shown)


bobstrecansky:
@Blacksmoke16 this looks great at first pass!  There's a small psalm error now; I believe that's our only limiter!

Blacksmoke16:
@bobstrecansky Oh hm, not sure what that error is about as it seems unrelated to my changes but I'll take a look!

.gitignore
```diff
@@ -5,7 +5,7 @@ vendor
 .idea/
 coverage.clover
 tests/coverage
-.php_cs.cache
+.php-cs-fixer.cache
```
Blacksmoke16, line 8: This was still using the old named after it was renamed as part of 1a1e0bd3d29f2b91fd7c436bc79cec8fb9941a97.

bobstrecansky (approved):

bobstrecansky:
@Blacksmoke16 - I'm sorry, I realized I merged this without giving you any answers to your open questions!  Inline below:
 
 
 
```
Should there be a links argument on the constructor of Span?
Currently the tracer is just calling setLinks if it's not null
```

I think we should set a (null default value) argument for the span constructor.  I read this as part of the spec:
 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-operations

Looking at other languages, it seems they follow this convention too.  Go example:
https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L59

```
It seems OpenTelemetry\Contrib\Otlp\SpanConverter isn't up to date as it's using annotations versus attributes? From what I know it should be the same as the protobuf implementation but just JSON encoded
Can save this for a follow up PR if that's not expected
```

That'd be a wonderful update.  Feel free to tackle this work or file an issue!

bobstrecansky:
The way that I'm reading the referenced part of the specification, I would agree with you!

Issue #227, opened 2020-11-24
  User Research Feedback Improvements - Use SetSpanStatus to change the span to an error manually, when catching the exception. 


bobstrecansky:
We'd like to be able to use SetSpanStatus to change the span to an error manually while catching an exception.

This ticket is the result of [this comment](https://docs.google.com/document/d/1n43oTVsnydmbpzvhN-5dgPAvRWKZCiugmmR_wg_8ziQ/edit?disco=AAAAHTU6d5k)

prondubuisi:
Hello @bobstrecansky  @dkarlovi, Does the recordException event solve this problem? I think wrapping  a code block in a try/catch i.e

```
try{
//something
}(Exception $exception){
     $span.recordException($exception);
}
```

Would automatically add the error event to the span

bobstrecansky:
You would need to have some sort of functionality that causes an exception for it to be thrown.  Something akin to:

https://www.w3schools.com/php/php_exception.asp

prondubuisi:
Thanks for the response @bobstrecansky, I am a little confused. Maybe I am missing what the problem here is, as well as the proposed solution. From @dkarlovi  comment, he says " Unclear how, the span code is not equal to the exception code, there's no way to auto-transform the exception (which is an object of type \Throwable) into proper span information." He also continues "Even though I've set the status from the exception, the span doesn't reflect any sort of exception."

prondubuisi:
Maybe a flow description of what is happening and what is expected to happen will help me understand this issue better.

bobstrecansky:
I'll defer to @dkarlovi as he was the one who was proposing the solution.

prondubuisi:
Hello @bobstrecansky I looked at this [comment](https://docs.google.com/document/d/1n43oTVsnydmbpzvhN-5dgPAvRWKZCiugmmR_wg_8ziQ/edit?disco=AAAAHTU6d5k) again, and the commit for the [exception test](https://github.com/dkarlovi/opentelemetry-php-user-test/commit/cda54e62eba88eae95d354b3dfa10105d69b95e4) , notice the similarity with https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/AlwaysOnJaegerExample.php#L59, I think the only issue now is confirming we see the exception properties with Jaeger and Zipkin,  which I don't. Once we can confirm that, I think this should be solved.

dkarlovi:
[...]

PR #407, opened 2021-09-08
Use Clock to get timestamp for events

Blacksmoke16:
Took me a while to figure out why my recorded exception wasn't showing up in the backend, but it turns out the event's timestamp wasn't in the duration of the span so it wasn't rendered as such.

This PR updates `renderException` and `Events#addEvent` to use `Clock#timestamp` instead of `time()`.

This probably fixes: https://github.com/open-telemetry/opentelemetry-php/issues/227#issuecomment-795151796.


bobstrecansky (approved):
LGTM; thanks @Blacksmoke16 !

PR #408, opened 2021-09-08
Update link API to be spec compliant

Blacksmoke16:
* Remove `addLink` from `Span`
* Remove `sampler` constructor arg from `Span`
  * Wasn't used and other langs don't really pass that in either as far as I can tell
* Rename `addLinks` to `addLink` in `SpanOptions`
  * Accepts an `API\Link` instance now
* Update `Links#addLink` to accept an `API\Link` instance
  * This is mainly so that the explicit link instance doesn't need to be re-created
* Allow settings attributes and links on `Span` creation
  * Forwards arguments from tracer to the `Span` constructor
  * Span start time is still missing from the API, but that can be handled later


sdk/Trace/Sampler/AlwaysOnSampler.php
```diff
@@ -39,7 +39,7 @@ public function shouldSample(
 
         return new SamplingResult(
             SamplingResult::RECORD_AND_SAMPLE,
-            null,
+            $attributes,
```
Blacksmoke16, line 42: This is so that the attributes passed to the samper are propagated to the created span. Not sure why this was `null` in the first place.

sdk/Trace/Span.php
```diff
@@ -51,27 +51,25 @@ public function __construct(
...
         $this->spanProcessor = $spanProcessor;
-        $moment = Clock::get()->moment();
-        $this->startEpochTimestamp = $moment[0];
-        $this->start = $moment[1];
+        [$this->startEpochTimestamp, $this->start] = Clock::get()->moment();
         $this->spanStatus = new SpanStatus();
 
         // todo: set these to null until needed
-        $this->attributes = new Attributes();
+        $this->attributes = $attributes ?? new Attributes();
```
Blacksmoke16, line 70: Spec says for attributes:

> An empty collection will be assumed if not specified.

api/Trace/Links.php
```diff
@@ -11,9 +11,7 @@ public function getIterator(): LinksIterator;
 
     /**
      * Adding a link should not invalidate nor change any existing iterators.
-     * @param SpanContext $context
-     * @param Attributes|null $attributes
```
bobstrecansky, line 15: Should we add `@param Link` as a string here?

sdk/Trace/Sampler/AlwaysOnSampler.php
bobstrecansky: Good question.  I'm not sure why either.

sdk/Trace/Tracer.php
```diff
@@ -123,16 +127,17 @@ public function setActiveSpan(API\Span $span): void
     }
 
     /**
-     * @param string $name
```
bobstrecansky, line 126: Is there a reason you removed these?

sdk/Trace/Tracer.php
```diff
@@ -255,21 +283,27 @@ public function getTracerProvider(): TracerProvider
         return $this->provider;
     }
 
+    public function startSpanWithOptions(string $name): API\SpanOptions
+    {
+        return new SpanOptions($this, $name, $this->provider->getSpanProcessor());
+    }
+
     /**
-     * @param string $name
```
bobstrecansky, line 259: same thing here in re: removal

bobstrecansky (approved):
Made a couple comments about removing docstrings (that aren't necessarily blockers), but other than that this looks great!

sdk/Trace/Tracer.php
Blacksmoke16: To me they're pointless given the related arguments already have a PHP type hint.

api/Trace/Links.php
Blacksmoke16: Would there be any benefit in doing that given it already has a `Link` type hint?

Issue #266, opened 2021-03-04
Implement no-op span behavior per spec


jodeev:
Match the no-op span behaviour to the one described in the spec section: [Wrapping a SpanContext in a Span](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/api.md#wrapping-a-spancontext-in-a-span)

<p>&nbsp;</p>

**Notes**
Currently [`NoopSpan` is used](https://github.com/open-telemetry/opentelemetry-php/blob/main/sdk/Trace/NoopSpan.php).

1. Confirm if our use of `NoopSpan` is intended to have the same behavior as `NonRecordingSpan`
2. If so, then we should rename it to `NonRecordingSpan` per the spec, and adapt the behavior as required by the spec.

Issue #240, opened 2021-01-14
Implement Context Interaction


jodeev:
Implement [Context Interaction](https://github.com/open-telemetry/opentelemetry-specification/blob/6589f46926dbbd11178a9f93889bf83a50ad2dd8/specification/trace/api.md#context-interaction) as per the specification.

bobstrecansky:
@beniamin - are you still planning on working on this?

Blacksmoke16:
Think this is resolved via #410.

Issue #239, opened 2021-01-14
Determining the Parent Span from a Context - should conform to latest OTel spec


jodeev:
It seems that the linking between the Context/SpanContext/Span is not implemented properly. 

The OpenTelemetry spec says that the SpanContext should be encapsulated inside the Span, but the Tracer is able to create a new Span with a specific SpanContext, created outside of the Span. 

For example TraceContext::extract returns a SpanContext and Tracer::startAndActivateSpanFromContext receives SpanContext as parameter.

We should follow the spec for the implementation.

alexperez52:
Hello, 
I'd like to take on this issue

PR #410, opened 2021-09-12
Context propagation refactor

Blacksmoke16:
Partially/fully resolves:

* https://github.com/open-telemetry/opentelemetry-php/issues/239
* https://github.com/open-telemetry/opentelemetry-php/issues/266
* https://github.com/open-telemetry/opentelemetry-php/issues/240

Are still some things that need done to be 100% compliant. I'll go thru the changes again and get a more detailed list, but the main two are:

* Moving most of the new logic into the API side of things
  * Given our current API implementation is only interfaces I figured it would be better to save this for later
* Refactor `Tracer` to change how spans can be created and activated in one method
  * Possibly via `Tracer#inSpan` that takes a callable?
  * Is slightly painful since you need to return both the span and the scope

TODO:

- [x] Document what still needs to be updated to be spec compliant
- [x] Add some doc comments where applicable to new types/methods
- [x] Better unit test coverage of new logic
- [ ] TBD

Future PR work:

* Fully implement https://github.com/open-telemetry/opentelemetry-php/issues/266
  * Do a pass on Span API to ensure the methods on the interface match spec
  * Maybe save implementing it in the API side for later?
* Decouple `Baggage` from `Context`
  * Make it its own class that implements the `Baggage` and `ImplicitContextKeyed` interfaces
* Refactor `Tracer`
  * Should no longer handle keeping track of active spans/making them active
    * https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/api.md#tracer
* Move the implementation of some things from the SDK to API side of things
  * Mainly noop span and context storage logic
    * Probably via abstract classes in the case of PHP?
* Support propagation of `Baggage`
  * Create a new `TextMapPropagator`
  * Create a new `CompositeTextMapPropagator` type to wrap Trace and Baggage implementations
    * https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/context/api-propagators.md#composite-propagator
  * Move types to the API layer and expose global propagator types
    * https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/context/api-propagators.md#global-propagators


Context/Context.php
```diff
@@ -4,9 +4,6 @@
 
 namespace OpenTelemetry\Context;
 
-/**
- * @template TContext of Context
```
Blacksmoke16, line 8: Going to look into if I can bring this back, but psalm was complaining when trying to pass a context instance to `TraceContextPropagator#inject`.

Context/Context.php
```diff
@@ -64,6 +69,22 @@ public function set(ContextKey $key, $value)
         return new static($key, $value, $this);
     }
 
+    /**
+     * @todo: Implement this on the API side
+     */
+    public function with(ImplicitContextKeyed $value): Context
```
Blacksmoke16, line 75: Given `Context` is an immutable collection I went with `with`, but given PHP doesn't have overloads I'm open to other names. I considered `withSpan` but this can also be used with baggage :/

Additionally would make it make sense to update `setValue` to `withValue` as I think that's a more common name for immutable containers.

Context/Context.php
```diff
@@ -94,19 +115,16 @@ public static function setValue(ContextKey $key, $value, $parent=null)
...
      */
     public function get(ContextKey $key)
     {
         if ($this->key === $key) {
             return $this->value;
         }
+
         if (null === $this->parent) {
-            throw new ContextValueNotFoundException();
+            return null;
```
Blacksmoke16, line 127: This relates to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md. Allows returning an invalid span in the case of there not being an active span. Checking for `null` seemed better than catching an exception.

Context/ImplicitContextKeyed.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Context;
+
+/**
+ * Represents a value that can be sored within {@see Context}.
+ * Allows storing themselves without exposing a {@see ContextKey}.
+ */
+interface ImplicitContextKeyed
```
Blacksmoke16, line 11: This mirrors the Java implementation and allows for `Baggage` to follow the same patterns as `Span`. Java apparently allows defining default method implementations on interfaces, so there is some minor duplication, but could introduce a trait similar to what `ContextValueTrait` was doing if it gets out of hand.

Makefile
```diff
@@ -38,7 +38,7 @@ proto:
 bash:
 	$(DC_RUN_PHP) bash
 style:
-	$(DC_RUN_PHP) php ./vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php --stop-on-violation --using-cache=no -vvv
+	$(DC_RUN_PHP) php ./vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php --using-cache=no -vvv
```
Blacksmoke16, line 41: Unrelated, but this was annoying me as you needed to run `make style` multiple times just to fix everything.

api/Trace/Span.php
```diff
@@ -55,5 +56,7 @@ public function setSpanStatus(string $code, ?string $description = null): Span;
      */
     public function end(int $timestamp = null): Span;
 
+    public function getSpanContext(): SpanContext;
```
Blacksmoke16, line 59: This is defined no the Java Span API, and is needed due to where active spans and such are stored now. E.g. within `TraceContextPropagator::inject`. I did also notice there is `Span#getContext()` that returns a clone of the span context, possibly could just replace that method with this one.

api/Trace/TextMapPropagator.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Trace;
+
+use OpenTelemetry\Context\Context;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#textmap-propagator
+ */
+interface TextMapPropagator
```
Blacksmoke16, line 12: This used to be `TextMapFormatPropagator`. Renamed and update interface based on latest spec.

api/Trace/TextMapPropagator.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Injects specific values from the provided {@see Context} into the provided carrier
+     * via an {@see PropagationSetter}.
+     *
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#textmap-inject
+     *
+     * @param mixed $carrier
+     */
+    public static function inject(&$carrier, Context $context = null, PropagationSetter $setter = null): void;
```
Blacksmoke16, line 31: I took some inspiration from the Ruby implementation where it made the context and setter optional to fall back on current context and the default setter (`PropagationMap` in our case).

Just makes it a bit easier to use, but happy to not do that if we'd rather be more explicit.

sdk/Baggage.php
```diff
@@ -4,23 +4,20 @@
...
 namespace OpenTelemetry\Sdk;
 
+use Generator;
 use OpenTelemetry\Context\Context;
 use OpenTelemetry\Context\ContextKey;
 
-/**
- * @extends Context<Baggage>
- */
 class Baggage extends Context
```
Blacksmoke16, line 11: Ultimately I think it would make more sense to better follow Java's implementation and make `Baggage` _not_ extend `Context`, but instead implement `ImplicitContextKeyed` itself.

That'll be for another PR tho.

sdk/Trace/NoopSpan.php
```diff
@@ -4,17 +4,21 @@
...
 use Throwable;
 
-class NoopSpan implements API\Span, ReadableSpan
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span
+ * @todo: Implement this on the API side.
+ * @todo: Can we just use {@see https://www.php.net/manual/en/language.oop5.anonymous.php}?
+ * @todo: If not rename this to `NonRecordingSpan`.
+ */
+class NoopSpan implements ReadWriteSpan
```
Blacksmoke16, line 20: Still some work to be done here. 

Ref: https://github.com/open-telemetry/opentelemetry-php/issues/266

sdk/Trace/NoopSpan.php
```diff
@@ -53,6 +56,7 @@ public function __construct(API\SpanContext $spanContext = null)
         }
         $this->attributes = new Attributes();
         $this->events = new Events();
+        $this->links = new Links();
```
Blacksmoke16, line 59: Missed this in #408 and #406.

sdk/Trace/Span.php
```diff
@@ -47,6 +47,44 @@ class Span implements API\Span, ReadableSpan
...
+    }
+
+    public static function getInvalid(): NoopSpan
+    {
+        if (null === self::$invalidSpan) {
+            self::$invalidSpan = new NoopSpan();
+        }
+
+        return self::$invalidSpan;
+    }
```
Blacksmoke16, line 77: Kinda annoying needing to have a static var just for this, but is needed so it returns the same instance given we cant just define it as a PHP const. Maybe could via `define`, but at that point would be easier to have a dedicated type and just have this method delegate to that.

sdk/Trace/TraceContextPropagator.php
```diff
@@ -16,40 +22,55 @@
  * traceparent header and relevant parts of the tracestate header containing
  * their proprietary information.
  */
-final class TraceContext implements API\TextMapFormatPropagator
+final class TraceContextPropagator implements API\TextMapPropagator
```
Blacksmoke16, line 25: Renamed this to make it more clear what it is, and updated implementation to better follow the new interface. Used https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java as a reference.

sdk/Trace/TraceContextPropagator.php
```diff
@@ -16,40 +22,55 @@
...
 
-    /**
-     * {@inheritdoc}
-     */
-    public static function inject(API\SpanContext $context, &$carrier, API\PropagationSetter $setter): void
+    /** {@inheritdoc} */
+    public static function inject(&$carrier, Context $context = null, PropagationSetter $setter = null): void
     {
+        $context = $context ?? Context::getCurrent();
+        $setter = $setter ?? new PropagationMap();
```
Blacksmoke16, line 41: Need to read the spec a bit more, but might be better to have these be singletons and expose them via class methods.

sdk/Trace/Tracer.php
```diff
@@ -124,6 +124,10 @@ public function setActiveSpan(API\Span $span): void
         $this->tail[] = $this->active;
 
         $this->active = $span;
+
+        // FIXME: This should either be called manually or as part of a dedicated tracer operation,
+        // probably some sort of `inSpan` method?
+        $span->makeCurrent();
```
Blacksmoke16, line 130: This relates to the 2nd major thing that needs to be done. Essentially this should either be called manually by the user, or we have an `inSpan` method that makes a yielded span active for the duration of a callback. Otherwise it might be hard to keep a dedicated method that creates a span and makes it the current one since it would need to return _both_ the span itself and a `Scope` to reset the active span.

sdk/Trace/Span.php
```diff
@@ -334,12 +372,19 @@ public function isStatusOk(): bool
...
+    public function storeInContext(Context $context): Context
+    {
+        return $context->set(SpanContextKey::instance(), $this);
+    }
+
+    /**
+     * @todo: Implement this on the API side
      */
-    protected static function getContextKey(): ContextKey
+    public function makeCurrent(): Scope
```
Blacksmoke16, line 385: This is what Java called it, but `makeActive` would might make more sense. Open to other suggestions.

Context/Context.php
bobstrecansky: With seems adequate to me for now.  I agree with you that `withSpan` may get confusing alongside baggage.

`withValue` also makes sense to me for immutable containers.  

Naming these attributes is difficult.  What synonym would be applicable with `Value` in this particular instance 🤔

Context/ImplicitContextKeyed.php
bobstrecansky: Sounds like a plan to me.  Following the JAVA spec is perfectly acceptable here!

Makefile
bobstrecansky: No problem.  Happy this helps with the developer experience!

api/Trace/Span.php
bobstrecansky: Yeah replacing `Span#getContext()` with this sounds a bit cleaner to me.

api/Trace/TextMapPropagator.php
bobstrecansky: Agreed that this makes things a bit easier to use; I'm all for it.

sdk/Baggage.php
bobstrecansky: Sounds good to me.  Maybe worth making an issue about it?

sdk/Trace/TraceContextPropagator.php
bobstrecansky: I think you're correct here - I can't think of cases where we'd have more than one setter.

sdk/Trace/Span.php
bobstrecansky: I'd have to think back, but I don't think there is a way to make a span current but not active.  `makeActive` feels like the sane choice to me here too.

Blacksmoke16: I also thought of `->activate()` if we think that's better 🤷. That seems to be what C# uses.

api/Trace/Span.php
Blacksmoke16: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-context

Looks like we should do the opposite, my as well use the name the spec suggests.

sdk/Baggage.php
Blacksmoke16: Was planning on tackling Baggage after getting context/propagation up to date. But I can create a more generic Baggage issue if you want it for tracking?

sdk/Trace/TextMapGetterSetter.php
```diff
@@ -0,0 +1,116 @@
...
+use function sprintf;
+use function strtolower;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/context/api-propagators.md#textmap-propagator Getter and Setter.
+ *
+ * Default implementation of {@see API\PropagationGetter} and {@see API\PropagationSetter}.
+ * This type is used if no custom getter/setter is provided to {@see API\TextMapPropagator::inject()} or {@see API\TextMapPropagator::extract()}.
+ */
+class TextMapGetterSetter implements API\PropagationGetter, API\PropagationSetter
```
Blacksmoke16, line 25: This used to be called `PropagationMap` but I renamed it to be a bit more clear on what it actually is.

sdk/Trace/TextMapGetterSetter.php
```diff
@@ -0,0 +1,116 @@
...
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
+    /** {@inheritdoc} */
+    public function keys($carrier): array
+    {
+        if (is_array($carrier)) {
```
Blacksmoke16, line 45: This method is required by the API but was missing an implementation. Opted to only support arrays even tho the actual getter supports `ArrayAccess` as that interface doesn't expose a way to get all the keys so oh well.

sdk/Trace/TextMapGetterSetter.php
Blacksmoke16: Although maybe `ArrayGetterSetter` would be even better...

sdk/Trace/Tracer.php
bobstrecansky: The Java pattern seems reasonable to me.  Leaving those details to the user might make things more flexible in the long run.  We can always update it later if we get complaints / don't love the implementation details.

sdk/Trace/Span.php
bobstrecansky: `->activate()` makes sense to me - that feels explicit enough.

sdk/Trace/TextMapGetterSetter.php
bobstrecansky: There aren't going to be times when we are GetterSetter'ing anything besides an array, correct?

sdk/Trace/TextMapGetterSetter.php
bobstrecansky: Yup; I like that plan.

sdk/Trace/TextMapGetterSetter.php
Blacksmoke16: I don't think we can have something for _all_ use cases. The majority of the time the carrier is most likely going to be an array, so having that be the default is fine. If that is not the case for someone, then they would need to define their own getter/setter type. I.e. there is a dedicated object or something.

Might make sense to make the carrier the 2nd argument as that might be more common than using a custom context?

sdk/Trace/TextMapGetterSetter.php
Blacksmoke16: Option 2: Define another interface that extends `ArrayAccess` to expose the `keys`, then only accept that or array. This way we can still call it `ArrayGetterSetter` but support both built in arrays and array-like objects. Thoughts?

Context/Context.php
Blacksmoke16: What about `withContextValue`? I.e. _a value that is meant to be stored in context._ This means the API would be:

* `set` => `with`
* `setValue` => `withValue`
* `with` => `withContextValue`

bobstrecansky: Sounds good to me!  Thanks for the clear, concise explanation.

sdk/Trace/TextMapGetterSetter.php
bobstrecansky: Agreed that a carrier will certainly be more common than using a custom context.  Sounds like a plan to me.

sdk/Trace/TextMapGetterSetter.php
Blacksmoke16: What about `ArrayAccessGetterSetter` then it's clear it could be used for anything that can be accessed as an array?

bobstrecansky: Sounds like a plan!

bobstrecansky (approved):

Issue #327, opened 2021-05-05
OTLP/GRPC exporter implementation -  SpanConverter functions should return an empty object


bobstrecansky:
A number of the SpanConverter functions should really return an empty object for the respective type, rather than a partially constructed object.

PR #412, opened 2021-09-13
Return empty objects when no spans are passed in OTLP/GRPC exporter implementation

Grunet:
Fixes the instances of [issue #327 ](https://github.com/open-telemetry/opentelemetry-php/issues/327) that I was able to notice.

I don't really have a sense for how this fits in overall yet, so I'm not sure what else I need to take into consideration.

Also I couldn't figure out the "idiomatic PHP" way to tell if an iterable was empty or not, hence this solution.

### DIFF
--- a/contrib/OtlpGrpc/SpanConverter.php
+++ b/contrib/OtlpGrpc/SpanConverter.php
@@ -176,11 +176,12 @@ class SpanConverter
 
     public function as_otlp_resource_span(iterable $spans): ResourceSpans
     {
-        // TODO: Should return an empty ResourceSpans when $spans is empty
-        // At the minute it returns an semi populated ResourceSpan
+        $isSpansEmpty = true; //Waiting for the loop to prove otherwise
 
         $ils = $convertedSpans = [];
         foreach ($spans as $span) {
+            $isSpansEmpty = false;
+
             /** @var \OpenTelemetry\Sdk\InstrumentationLibrary $il */
             $il = $span->getInstrumentationLibrary();
             $ilKey = sprintf('%s@%s', $il->getName(), $il->getVersion()??'');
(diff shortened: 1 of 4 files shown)


Grunet:
Currently blocked on the CLA signing. See [this Slack message](https://cloud-native.slack.com/archives/C01NFPCV44V/p1631498316087400?thread_ts=1631244100.052100&cid=C01NFPCV44V) for more details.

bobstrecansky:
@Grunet - you have a style error in your PR.  You should be able to run `make style` locally and it should let you know what you need to fix or automatically fix it.

Grunet:
> 
> 
> @Grunet - you have a style error in your PR. You should be able to run `make style` locally and it should let you know what you need to fix or automatically fix it.

Ah yup, think it should be fixed now

WillsonHG:
@Grunet  Please try to sign your ICLA again. You'll have to wait about 5 seconds for the ICLA option to turn green, it seems to be loading a bit slowly.

Grunet:
> 
> 
> Still seeing a `braces` violation here: https://github.com/open-telemetry/opentelemetry-php/pull/412/checks?check_run_id=3598306857#step:8:10
> 
> I think you'll just need to clean that up and we'll be able to see the next step in the CI process!

Ah apologies about that! I think it should be good now (mental note made to run "make style" as many times as needed until it gets to 100%)

Grunet:
> 
> 
> @Grunet Please try to sign your ICLA again. You'll have to wait about 5 seconds for the ICLA option to turn green, it seems to be loading a bit slowly.

@WillsonHG I was able to get through the whole process this time. Wasn't sure if I was supposed to use my real name or not so it might have both that and my auto-filled username. Not sure if that's a problem or how I would fix it

Grunet:
FWIW I just realized the code changes here may fail if ```$spans``` is an "exhausted generator" like I mentioned in [this issue](https://github.com/open-telemetry/opentelemetry-php/issues/415), since I think it would throw an exception of ```Exception: Cannot traverse an already closed generator``` when hitting the foreach loop (similar to [this SO post](https://stackoverflow.com/questions/43593670/php-iterating-twice-a-generic-iterable))

I'm not quite sure what the appropriate handling for that situation is (wrap the loop with a try/catch?) or if it really needs to be handled rn

bobstrecansky:
@Grunet - we can always come back to the handling situation later if it arises!  I'm happy with this PR as is.

PR #417, opened 2021-09-19
Add AttributeLimits, SpanLimits and dropped event/links counting

lalex:
This PR adds support for [Span Limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits) and common [Attribute Limits](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits).

The `Span` class counts attributes, events and links dropped due to limits. These counters are exported via OTLP.

Fixed: Numeric attribute key is interpreted as an integer value which breaks string-key type checks on reading from Attributes.

### DIFF
--- a/sdk/Internal/StringUtil.php
+++ b/sdk/Internal/StringUtil.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Internal;
+
+class StringUtil
+{
+    public static function substr(string $value, int $offset, int $length): string
+    {
+        if (function_exists('mb_substr')) {
+            return \mb_substr($value, $offset, $length);
+        }
+
+        return \substr($value, $offset, $length);
+    }
+
+    public static function strlen(string $value): int
+    {
+        if (function_exists('mb_strlen')) {
+            return \mb_strlen($value);
+        }
+
+        return \strlen($value);
+    }
+}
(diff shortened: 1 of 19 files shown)


sdk/Internal/StringUtil.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Sdk\Internal;
+
+class StringUtil
+{
+    public static function substr(string $value, int $offset, int $length): string
+    {
+        if (function_exists('mb_substr')) {
```
bobstrecansky, line 11: Can you think of a particular place where `mb_substr` and `mb_strlen` wouldn't exist?  I can't, but I'm probably a bit jaded here.

bobstrecansky (approved):
One question but it's definitely not a blocker.  Going to merge this!

sdk/Internal/StringUtil.php
lalex: `mbstring` is a non-default extension (https://www.php.net/manual/en/mbstring.installation.php) while Docker image has it enabled by default.

This check in code could be replaced by the additional requirement in `composer.json`. But I'm not sure which is better. When the user doesn't have `mbstring` installed that means the user code doesn't care about Unicode at all. The library would not help.

bobstrecansky: Thank you for the context.  I think this is sufficient for now; we can most certainly include that in `composer.json` if we ever decided we wanted to.  I like the direction you took.

PR #420, opened 2021-09-20
Publish OpenTelemetry PHP Base Docker Image

bobstrecansky:
Currently, a developer for this project uses a docker base image that includes the gRPC extension.  That takes approximately 15-20 minutes to build, causing developer friction.  This PR creates a github action that builds a new opentelemetry-php base docker image on push.  This will be used as the default standard base Docker image in subsequent PRs.

### DIFF
--- a/.github/workflows/publish-otel-php-base-docker-image
+++ b/.github/workflows/publish-otel-php-base-docker-image
@@ -1,0 +1,30 @@
+name: publish-otel-php-base-docker-image
+on:
+  push:
+    branches: [ 'main' ]
+  workflow_dispatch:
+jobs:
+  push_to_registry:
+    name: OpenTelemetry PHP base docker image creation
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
> This PR creates a github action that builds a new opentelemetry-php base docker image on push.

This feels a bit overkill given it's only going to change when we change the version of PHP or one of the extensions.

https://github.com/open-telemetry/opentelemetry-php/blob/main/docker-compose.yaml#L3 should probably also be updated to use this image.

zsistla:
This is a great idea!  I like the official docker image being put up to at least give a baseline for future, repeatable builds as well as protect against images beyond the php otel repo's control that we use but that might change or be eoled in the future.

zsistla (approved):

bobstrecansky:
@Blacksmoke16 - I thought about potentially limiting it to those extension changes, but I thought this was a good first step.
I'll change the docker-compose config, as well as a couple other places it gets used in a subsequent PR.

PR #422, opened 2021-09-21
Span Refactor

### LINKED ISSUE #239: Determining the Parent Span from a Context - should conform to latest OTel spec

### LINKED ISSUE #266: Implement no-op span behavior per spec

### LINKED ISSUE #379: Circular and Vestigial Functions in SpanContext
*Is your feature request related to a problem?*

The `SpanContext` is creating trace and span ids when that is a job that should be done by the `IdGenerator`. There are a few functions that should not exist within the `SpanContext` class: 

* `generate()`
* `generateSampled()`
* `fork()`
* `restore()`
* *` randomHex()`

These functions generate a `SpanContext` with the built in `randomHex()` function when instead a new `SpanContext` should only be created through the constructor. This is consistent with the specifications and other implementations of `SpanContext` in the other language repos. 
 
*Describe the solution you'd like*

The four functions above should be removed and all span context creation should occur through the constructor of `SpanContext`. An `IdGenerator` should be used to generate any trace or span ids necessary for a given `SpanContext`.

It may be useful to have a function to create a new instance of `SpanContext` derived from an existing one, for example to set the sample flag. 

@alolita

Blacksmoke16:
Still some work to be done to ensure everything is in an actual working state, but wanted to get this created to start a discussion around the approaches and such.

Sorry for the large PR, but it's kinda core and touches a lot of stuff :/

What this PR does:

* Drops support for PHP 7.3
  * This was needed to allow using the proper return type hint on the SDK Span methods
    * Ref: https://www.php.net/manual/en/language.oop5.variance.php
* Refactor the Span API to better follow the spec/Java implementation. Fixes #239 
  * Removes some extraneous methods, renames some, and updates some signatures
  * Changes the `Span` constructor and how you create a span
  * Introduces a dedicated `SpanData` interface that represents an immutable snapshot of a span mainly for use as part of span exporters
    * Ref: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/sdk.md#additional-span-interfaces
    * > Note: Typically this will be implemented with a new interface or (immutable) value type. In some languages SpanProcessors may have a different readable span type than exporters (e.g. a SpanData type might contain an immutable snapshot and a ReadableSpan interface might read information directly from the same underlying data structure that the Span interface manipulates).
* Refactor the `Tracer` implementation to leverage a `SpanBuilder` versus relying on multiple methods with variable amount of arguments
  * Renames `NoopSpan` to `NonRecordingSpan` and makes it _only_ implement `Span`, not `ReadWriteSpan`. Fixes #266
    * Ref: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/api.md#wrapping-a-spancontext-in-a-span
* `TracerProvider` and `Tracer` now leverage a `TracerSharedState` in order to share the instances of idGenerator, sampler, etc.
* Various other minor/API changes to help with having more proper typing
* Update/simplify `Clock` interface, rename SDK `Clock` to `SystemClock` and introduce `TestClock`
  * Add ability to set a `Clock` instance to use by default, e.g. for testing
* Consoldiate `SpanContext` constructor methods into 2, one for remote and one for non remote contexts
  * Fixes #379

What still needs to be done:

* Finish updating tests
* Ensure everything still functions
  * Will need to update all the examples, probably will do one just to help test, but others can prob be done in a follow up PR
* Merge in #417 and fix conflicts
  * Pretty much have `todo`s where they were needed so shouldn't be too bad
* Address any PR feedback


sdk/Trace/Span.php
```diff
@@ -59,383 +81,374 @@ public static function fromContext(Context $context): API\Span
...
+     * @readonly
+     *
+     * @todo: Java just has this as list<API\Event>, could we just do that?
      */
-    public function setSpanStatus(string $code, ?string $description = null): API\Span
-    {
-        if ($this->isRecording()) {
-            $this->spanStatus->setStatus($code, $description);
-        }
+    private API\Events $events;
```
Blacksmoke16, line 176: I'm skeptical if we really need these `API\Events` and `API\Links` objects versus just using arrays, or `ArrayObject`.

sdk/Trace/Span.php
```diff
@@ -59,383 +81,374 @@ public static function fromContext(Context $context): API\Span
...
 
-    public function setStart(int $start): Span
-    {
-        $this->start = $start;
+    /** @readonly */
+    private int $kind;
 
-        return $this;
-    }
+    // TODO: Store a clock instance on the Span?
```
Blacksmoke16, line 191: Java seems to store a `Clock` instance within the object, not sure there's any reason to do that in our case given you can just `Clock::get()`. In Java's case I think it's special in that the clock is anchored to when the span is created, versus just reading system time info every time.

sdk/Trace/TracerProvider.php
```diff
@@ -4,60 +4,54 @@
...
-        $this->idGenerator = $idGenerator ?? new RandomIdGenerator();
+    /** @var array<string, API\Tracer> */
+    private $tracers;
+
+    /** @readonly */
+    private TracerSharedState $tracerSharedState;
+
+    /** @param list<SpanProcessor>|SpanProcessor $spanProcessors */
+    public function __construct(
+        $spanProcessors = [],
```
Blacksmoke16, line 25: I made it so you can pass 1 (or an array of) span processors to the constructor, versus needing to call `addSpanProcessor` a bunch of times. This is mainly needed so that `TracerSharedState` could be created with the full set of processors.

sdk/Trace/Test/SpanData.php
```diff
@@ -0,0 +1,275 @@
...
+
+namespace OpenTelemetry\Sdk\Trace\Test;
+
+use OpenTelemetry\Sdk\InstrumentationLibrary;
+use OpenTelemetry\Sdk\Resource\ResourceInfo;
+use OpenTelemetry\Sdk\Trace as SDK;
+use OpenTelemetry\Sdk\Trace\StatusData;
+use OpenTelemetry\Trace as API;
+
+class SpanData implements SDK\SpanData
```
Blacksmoke16, line 13: This is a mutable version of `SpanData`, mainly useful in tests as a mini builder so you don't _have_ to provide _every_ span constructor argument.

Blacksmoke16: Should probably do some composer something or other so this doesn't get shipped when installing the package. I.e. should be a development type only.

bobstrecansky (commented):
A couple initial comments but I don't think I have any comments that are blockers currently.

sdk/Trace/Span.php
bobstrecansky: You think they might just be vestegial?

sdk/Trace/Span.php
bobstrecansky: Yeah I think that our implementation is good!

sdk/Trace/Test/SpanData.php
bobstrecansky: You might need to move it into its own separate file and ignore it if that's what you'd like to do.  I'm sure there is a way to do it for a particular class but I don't know that answer off the top of my head.

contrib/ZipkinToNewrelic/SpanConverter.php
```diff
@@ -42,22 +44,26 @@ private function sanitiseTagValue($value)
...
+            'parentId' => $spanParent->isValid() ? $spanParent->getSpanId() : null,
             'localEndpoint' => [
                 'serviceName' => $this->serviceName,
             ],
-            'name' => $span->getSpanName(),
-            'timestamp' => (int) ($span->getStartEpochTimestamp() / 1e3), // RealtimeClock in microseconds
-            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds
+            'name' => $span->getName(),
+            'timestamp' => $startTimestamp,
+            'duration' => max(1, $endTimestamp - $startTimestamp), // Diff in microseconds
```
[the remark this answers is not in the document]
Blacksmoke16: Was more so for `contrib/Newrelic/SpanConverter.php:44` needing to be a `float`.

sdk/Trace/SpanBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+            return $this;
+        }
+
+        $this->totalNumberOfLinksAdded++;
+
+        if (null === $this->links) {
+            $this->links = new Links();
+        }
+
+        // TODO: Skip links over the limit.
```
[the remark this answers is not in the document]
Blacksmoke16: Java does:

```java
    // don't bother doing anything with any links beyond the max.
    if (links.size() == spanLimits.getMaxNumberOfLinks()) {
      return;
    }
```

so something like that.

sdk/Trace/Tracer.php
```diff
@@ -4,334 +4,48 @@
...
-        int $spanKind = API\SpanKind::KIND_INTERNAL,
-        ?API\Attributes $attributes = null,
-        ?API\Links $links = null
-    ): API\Span {
-        $parentContextIsNoopSpan = !$parentContext->isValid();
-
-        if ($parentContextIsNoopSpan) {
-            $parentContext = $this->importedContext ?? SpanContext::fork($this->provider->getIdGenerator()->generateTraceId(), true);
+        if ($this->tracerSharedState->hasShutdown()) {
+            // TODO: Return a noop tracer
```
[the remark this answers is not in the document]
Blacksmoke16: I'm pretty sure we just don't have one yet.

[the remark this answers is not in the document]
bobstrecansky: I'm surprised but not shocked 👍🏼

sdk/Trace/SpanBuilder.php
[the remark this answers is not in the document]
bobstrecansky: Yup; just wanted to make sure I wasn't missing anything.

contrib/Zipkin/SpanConverter.php
```diff
@@ -42,22 +44,26 @@ private function sanitiseTagValue($value)
...
+            'parentId' => $spanParent->isValid() ? $spanParent->getSpanId() : null,
             'localEndpoint' => [
                 'serviceName' => $this->serviceName,
             ],
-            'name' => $span->getSpanName(),
-            'timestamp' => (int) ($span->getStartEpochTimestamp() / 1e3), // RealtimeClock in microseconds
-            'duration' => (int) (($span->getEnd() - $span->getStart()) / 1e3), // Diff in microseconds
+            'name' => $span->getName(),
+            'timestamp' => $startTimestamp,
+            'duration' => max(1, $endTimestamp - $startTimestamp),
```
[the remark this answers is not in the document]
Blacksmoke16: I may have got a bit overzealous on using `max`, but for Zipkin it is correct. Basically we just want the minimum value to be `1`, which without `max` could be `0`.

Ref: https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml#L467 and https://github.com/open-telemetry/opentelemetry-java/pull/1694

sdk/Trace/Span.php
Blacksmoke16: Or at least unnecessary. It's just trying to represent something that can be traversed, counted, and have things added to it. I.e. essentially a https://www.php.net/manual/en/class.arrayobject.php. Not really providing any meaningful abstraction/benefit.

api/Trace/Span.php
```diff
@@ -4,62 +4,97 @@
...
 
     /**
-     * @param string $name
-     * @param int $timestamp
-     * @param Attributes|null $attributes
-     * @return Span Must return $this to allow setting multiple attributes at once in a chain.
+     * Returns the current {@see Span} from the current {@see Context},
+     * falling back on {@see Span::getEmpty()} if there is no span in the current context.
+     *
+     * @todo Implement this in the API layer
```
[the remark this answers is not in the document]
Blacksmoke16: Some of the stuff that has this `todo` says like:

> This functionality SHOULD be fully implemented in the API when possible.

or 

> This functionality MUST be fully implemented in the API, and SHOULD NOT be overridable.

Ref: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/api.md#context-interaction and https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/api.md#wrapping-a-spancontext-in-a-span

However at the moment the actual implementation is in the SDK. Moving all that to the API side of things will be another PR. Java has the ability to define default implementations of things on its interfaces. Given we can't do that, the next best thing would probably be an `AbstractSpan` or something that defines these implementations as `final`, then the SDK `Span` inherits from it. Will also help with DRYing some stuff up abit as well.

sdk/Trace/ImmutableSpan.php
```diff
@@ -0,0 +1,154 @@
...
+    private int $endEpochNanos;
+    private bool $hasEnded;
+
+    /**
+     * @internal
+     * @psalm-internal OpenTelemetry\Sdk
+     *
+     * @param non-empty-string $name
+     */
+    public function __construct(
```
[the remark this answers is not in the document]
Blacksmoke16: Is the same constructor Java has. With the introduction of `Test\SpanData` and the `SpanBuilder` being publicly exposed, probably not as big of a deal as before. It's not something you should need to interact with often/ever in real code.

sdk/Trace/Span.php
lalex: Moreover an additional layer of abstraction gets `psalm` crazy, it can't infer a correct type of `$event` in `foreach ($events as $event)` loop.

sdk/Trace/SpanBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    private string $spanName;
+
+    /** @readonly */
+    private InstrumentationLibrary $instrumentationLibrary;
+
+    /** @readonly */
+    private TracerSharedState $tracerSharedState;
+
+    // TODO: Add SpanLimits to constructor
```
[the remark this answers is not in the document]
lalex: Span limits have arrived in #417

api/Trace/Span.php
```diff
@@ -4,62 +4,97 @@
...
+     * @param non-empty-string $name
+     */
+    public function updateName(string $name): Span;
+
+    /**
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/trace/api.md#set-status
+     *
+     * @param StatusCode::STATUS_* $code
+     *
+     * @todo: Make this return SDK\StatusData
```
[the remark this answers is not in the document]
Blacksmoke16: :thinking: Actually I don't remember what this was for...

sdk/InstrumentationLibrary.php
```diff
@@ -9,9 +9,19 @@
  */
 class InstrumentationLibrary
 {
-    private $name;
+    private static ?self $empty = null;
 
-    private $version;
+    public static function getEmpty(): InstrumentationLibrary
```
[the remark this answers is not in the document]
Blacksmoke16: Java exposes one as well, which after grepping thru the code only seems to be used in tests. We could mark it as internal, or move it to some `Test` name space? Is there any reason to _not_ expose it? That would deff be the simpler solution.

sdk/Trace/ImmutableSpan.php
```diff
@@ -0,0 +1,154 @@
...
+        Span $span,
+        string $name,
+        API\Links $links,
+        API\Events $events,
+        API\Attributes $attributes,
+        int $totalAttributeCount,
+        int $totalRecordedEvents,
```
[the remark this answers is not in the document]
Blacksmoke16: > Is it possible to have `count($events) !== $totalRecordedEvents`?

Yea, in the case more events were added than the allowed limit. `$totalRecordedEvents` would be incremented but they wouldn't be actually added to the list of events. Using this method it's possible to calculate the amount of dropped events

But thanks for that link, maybe we could change the interface a bit to expose the dropped count based on `total - events.size` instead of just the total amount. That would probably be a bit more spec compliant.

sdk/Trace/ImmutableSpan.php
```diff
@@ -0,0 +1,154 @@
...
+namespace OpenTelemetry\Sdk\Trace;
+
+use OpenTelemetry\Sdk\InstrumentationLibrary;
+use OpenTelemetry\Sdk\Resource\ResourceInfo;
+use OpenTelemetry\Trace as API;
+
+/**
+ * @psalm-immutable
+ */
+final class ImmutableSpan implements SpanData
```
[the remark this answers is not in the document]
Blacksmoke16: > Maybe this SpanData interface should not substitute an existing ReadbleSpan interface too much?

The other major point is that it's immutable.  If you pass the actual span typed as `ReadableSpan`, then it's possible to mutate the original one by reference, which isn't great. Not sure I fully understand your question/concern.

composer.json
```diff
@@ -50,17 +53,18 @@
...
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.3",
         "composer/xdebug-handler": "^2.0",
-        "phan/phan": "^5.0",
         "friendsofphp/php-cs-fixer": "^3.0",
-        "vimeo/psalm": "^4.0",
+        "guzzlehttp/guzzle": "^7.3",
+        "guzzlehttp/psr7": "^2.0@RC",
+        "mockery/mockery": "^1.4",
```
Blacksmoke16, line 60: I installed mockery in order to test some of the interactions of `SpanBuilder`. IMO it's a much better mocking library than the one that comes with `PHPUnit`.

sdk/Trace/ImmutableSpan.php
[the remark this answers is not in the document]
lalex: Sounds reasonable. But I have the only concern about a performance overhead that may be caused by copying data, which may be insignificant.

Do we need such strong protection for immutability? Readable span is supposed to be used inside SDK and contributed processors and exporters. The code is covered by the static analysis tools like psalm. And for the end-user the API exposes a writable Span interface.

For me, the `SpanData` holds a `Span`'s state enriched by something from the "context" like Tracer's state or something else which is not a part of the Span but it is useful or necessary for the Exporter or Processor.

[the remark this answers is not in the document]
bobstrecansky: We do need strong protection for immutability; that's  one of the main things all of the maintainers harped on.  There will most certainly be many third party exporters (they live in the `contrib/` repo now, but will eventually all land in https://github.com/open-telemetry/opentelemetry-php-contrib/

Does that answer your question @Blacksmoke16 ?

sdk/InstrumentationLibrary.php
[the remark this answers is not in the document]
bobstrecansky: I don't have any gripes with using a `Test` namespace or marking it as internal.

api/Trace/Events.php
```diff
@@ -15,6 +15,5 @@ interface Events extends \IteratorAggregate, \Countable
      */
     public function addEvent(string $name, ?Attributes $attributes = null, ?int $timestamp = null): Events;
 
-    public function count(): int;
```
Blacksmoke16, line 18: This method is already added since the interface extends `Countable` and was causing issues with duplicate code suggestions and psalm `pure` methods i been leveraging.

contrib/OtlpGrpc/SpanConverter.php
```diff
@@ -81,25 +81,23 @@ public function as_otlp_span_kind($kind): int
...
+            'start_time_unix_nano' => $span->getStartEpochNanos(),
+            'end_time_unix_nano' => $span->getEndEpochNanos(),
+            'kind' => $this->as_otlp_span_kind($span->getKind()),
             'trace_state' => (string) $span->getContext()->getTraceState(),
-            'dropped_attributes_count' => $span->getAttributes()->getDroppedAttributesCount(),
-            'dropped_events_count' => $span->getDroppedEventsCount(),
-            'dropped_links_count' => $span->getDroppedLinksCount(),
+            'dropped_attributes_count' => $span->getTotalDroppedAttributes(),
+            'dropped_events_count' => $span->getTotalDroppedEvents(),
+            'dropped_links_count' => $span->getTotalDroppedLinks(),
```
Blacksmoke16, line 100: Technically these should not be set if they are `0`. But this can be handled in another PR.

phpstan.neon.dist
```diff
@@ -11,6 +12,7 @@ parameters:
         - vendor
         - proto
         - tests/TraceContext/W3CTestService
+        - examples # TODO: Remove this once examples are updated
```
Blacksmoke16, line 15: I excluded examples directory temporarily from phpstan and psalm just to cut down on the scope of an already huge PR. Can do a pass on the examples in a follow up PR.

psalm.xml.dist
```diff
@@ -12,9 +12,11 @@
             <directory name="vendor"/>
             <directory name="proto"/>
             <directory name="tests/TraceContext/W3CTestService"/>
+            <directory name="examples"/> <!--TODO: Remove this once examples are updated-->
         </ignoreFiles>
     </projectFiles>
     <plugins>
         <pluginClass class="Psalm\PhpUnitPlugin\Plugin"/>
+        <pluginClass class="Psalm\MockeryPlugin\Plugin"/>
```
Blacksmoke16, line 20: Allows Psalm to properly handle `Mockery::mock` types and such. Also had to add one for phpstan.

tests/Sdk/Integration/SpanLimitsTest.php
```diff
@@ -1,101 +0,0 @@
...
-use OpenTelemetry\Sdk\Trace\Links;
-use OpenTelemetry\Sdk\Trace\SpanContext;
-use OpenTelemetry\Sdk\Trace\SpanLimits;
-use OpenTelemetry\Sdk\Trace\SpanLimitsBuilder;
-use OpenTelemetry\Sdk\Trace\Tracer;
-use OpenTelemetry\Sdk\Trace\TracerProvider;
-use OpenTelemetry\Trace\SpanKind;
-use PHPUnit\Framework\TestCase;
-
-class SpanLimitsTest extends TestCase
```
Blacksmoke16, line 20: I added equivalent tests to `SpanTest`.

sdk/Trace/SpanBuilder.php
```diff
@@ -0,0 +1,212 @@
...
+            return $this;
+        }
+
+        $this->links->addLink(new Link($context, $attributes));
+
+        return $this;
+    }
+
+    /** @inheritDoc */
+    public function setAttribute(string $key, $value): API\SpanBuilder
```
lalex, line 93: The `SpanLimits` are missed here. As soon as we have `$this->spanLimits` set in the constructor we may use it.
`Span` applies the limits only when initializes an empty `Attributes` collection: https://github.com/open-telemetry/opentelemetry-php/blob/189166e477a065520e438a7a705b28650595de9c/sdk/Trace/Span.php#L269-L271
Some tests on `SpanBuilder::setAttribute` could help to cover this issue.

Blacksmoke16: Thanks! I handled and added tests for both of these cases. See latest commit.

bobstrecansky (approved):

Issue #326, opened 2021-05-05
OTLP/GRPC exporter implementation - Test improvements


bobstrecansky:
the OTLP / GRPC Exporter could use some more tests (either acceptance or mocking the Grpc Client)

Grunet:
I would like to take a crack at this (to also help give me a better sense for the repo/library)

From just a code review (not looked at the coverage data yet) it seems like the case where the [gRPC call succeeds](https://github.com/open-telemetry/opentelemetry-php/blob/main/contrib/OtlpGrpc/Exporter.php#L150) may be untested. 

I may start with trying to get a test case mocking the TraceServiceClient working, then work on improving it to actually make the gRPC call (to a local endpoint?) instead

bobstrecansky:
Thanks @Grunet - assigned to you.  Let us know if you have any questions!

Grunet:
Oh this one can actually be closed. I think I made the PR and it was merged but at the time didn't know how to reference this issue so it would auto-close on-merge.

I made a note to follow up on it but never did...

tidal:
@Grunet 
Ok, thanks for the info.
Closing.

PR #430, opened 2021-09-30
Add 2 More Tests through OtlpGrpc\Exporter

Grunet:
Addresses [issue 326](https://github.com/open-telemetry/opentelemetry-php/issues/326).

Per Slack conversation, it sounded like there weren't any tests already going "downstream" and making real gRPC calls, and I didn't see any "upstream" tests calling into this class (outside of the examples) so I concluded that the unit-y level here was still appropriate.


tests/Contrib/Unit/OTLPGrpcExporterTest.php
```diff
@@ -4,15 +4,104 @@
...
+                                ->shouldReceive('wait')
+                                ->andReturn(
+                                    [
+                                        'unused response data',
+                                        new class() {
+                                            public $code;
+                        
+                                            public function __construct()
+                                            {
+                                                $this->code = 'An unexpected status';
```
Blacksmoke16, line 80: Thoughts on moving some of this logic to private functions? E.g. one to get the mock client where you can pass in status, and another to get the exporter instance?

tests/Contrib/Unit/OTLPGrpcExporterTest.php
```diff
@@ -4,15 +4,104 @@
...
+            '',
+            '',
+            false,
+            10,
+            $mockClient
+        );
+               
+        $exporterStatusCode = $exporter->export([new SpanData()]);
+
+        $this->assertEquals(Exporter::FAILED_NOT_RETRYABLE, $exporterStatusCode);
```
Blacksmoke16, line 101:
```suggestion
        $this->assertSame(Exporter::FAILED_NOT_RETRYABLE, $exporterStatusCode);
```

Minor, but is more ideal to use `assertSame`, otherwise things like `'1' == 1` would not be detected.

Grunet:
Hmm, weird that PHP 7.4 isn't passing here for the 2nd of the new tests, but 8.0 is (at least in my local setup when running "make test". 

I wonder if the Mockery::mock is somehow leaking between the 2 tests, so the 1st one returning a success is getting re-used in the 2nd where it shouldn't be? I'm not sure

Blacksmoke16:
> I wonder if the Mockery::mock is somehow leaking between the 2 tests, so the 1st one returning a success is getting re-used in the 2nd where it shouldn't be? I'm not sure

It shouldn't because you're inheriting from `MockeryTestCase`. That would clean them up after every test case.

tests/Contrib/Unit/OTLPGrpcExporterTest.php
Grunet: Yeah I think I was leaning onto the [AHA (Avoid Hasty Abstraction)](https://kentcdodds.com/blog/aha-testing) side of things for this, but I can go ahead and split them off into a little helper

tests/Contrib/Unit/OTLPGrpcExporterTest.php
Grunet: That's super good to know! Went ahead and updated them

tests/Contrib/Unit/OTLPGrpcExporterTest.php
Grunet: I went ahead and split the mock creation into a helper and definitely think they read a lot nicer now, but I wasn't sure if splitting the Exporter constructor off into one would be over-doing it. Don't have a strong opinion either way so just lmk what y'all are feeling on this!

Blacksmoke16: Oh wow, that's pretty fancy :P I was just thinking like `private function createMockTraceServiceClient(int $statusCode)` and call it a day :sweat_smile:.

Grunet:
I still can't quite reason this 7.4 vs 8.0 difference out, so I think in one of the next few evenings I will re-build locally with a 7.4 image, make sure things work with the otel-php main branch, and then try with my fork.

If I can reproduce it there, I'll probably add in a lot of print statements around this code to try and suss out why it's sending "success" in this case when it shouldn't be.

In the meantime if anyone looking this over happens to see any 7.4 vs 8 differences that could be involved, feel free to leave a comment! Would be much appreciated

Blacksmoke16:
@Grunet It's because the semantics of comparison between strings and numbers changed between PHP 7 and 8. I.e. your tests uncovered a bug in the library when using PHP 7.4. The gist of it is `'An unexpected status' == 0 # => true`. We really should be using strict comparison here. I.e. `$status->code === \Grpc\STATUS_OK` which would solve the issue.

Grunet:
> 
> 
> @Grunet It's because the semantics of non-strict comparison between strings and numbers changed between PHP 7 and 8. I.e. your tests uncovered a bug in the library when using PHP 7.4. The gist of it is `'An unexpected status' == 0 # => true`. We really should be using strict comparison here. I.e. `$status->code === \Grpc\STATUS_OK` which would solve the issue.
> 
> Refs:
> 
>     * https://wiki.php.net/rfc/string_to_number_comparison#introduction
> 
>     * https://stackoverflow.com/a/6843075

Ahhh nice, thank you!! Learning so much about PHP every evening :) 

Looks like that got all the checks passing!

bobstrecansky (approved):

PR #432, opened 2021-10-01
Events & Links improvements

### LINKED ISSUE #424: Replace`API\Events` and `API\Links`
The `API\Events` and `API\Links` objects feel a bit redundant versus just using arrays, or `ArrayObject`.

Still TBD which we want to go with data model wise. The main benefit of `ArrayObject` is it would be passed by reference by default, which may or not be helpful.

_Originally posted by @Blacksmoke16 in https://github.com/open-telemetry/opentelemetry-php/pull/422#discussion_r712602938_

Blacksmoke16:
* Replace `Links` and `Events` types with `array`
  * Also removes the ability to add attributes to an existing `Event` to preserve immutability 
* Apply limits to links and event attribute lengths
  * Removes one `Span` test as it is tested in `SpanBuilder`
* Some minor changes to `API\Event`
  * Rename `getTimestamp` to `getEpochNanos` to better match span API

Resolves #424


sdk/Trace/SystemClock.php
```diff
@@ -24,6 +24,10 @@ public static function getInstance(): self
     /** @inheritDoc */
     public function now(): int
     {
+        /**
+         * `microtime` returns a unix timestamp _WITH_ microseconds, not _IN_ microseconds.
```
bobstrecansky, line 28: Thank you for adding a comment here.  Time is confusing.

bobstrecansky (approved):

PR #434, opened 2021-10-02
Reorganize types

### LINKED ISSUE #425: PHP organization/naming pass
Currently all of the source PHP code lives within their own folders in the root directory. From what I remember talking with @bobstrecansky is that this was just how the original creator of the lib set it up, and not some OTLP requirement.

Related to this, the API is defined under the `OpenTelemetry\Trace` namespace and are using names like `Span` for interfaces.

Because we are still in a pre-alpha state, I propose to makes some changes before it becomes impossible due to needing to maintain backwards compatibility.

These suggestions include:

* Follow https://www.php-fig.org/bylaws/psr-naming-conventions/
  * E.g. `Span` => `SpanInterface`
* Follow https://wiki.php.net/rfc/class-naming
* Redo the directory structure to have all source code under `src`. Something like:

```txt
.
├── src
│   ├── API
│   ├── Context
│   ├── Proto
│   └── SDK
└── tests
    ├── API
    ├── Context
    └── SDK
```
* Update the namespaces of the types to match the directory structure

_Originally posted by @Blacksmoke16 in https://github.com/open-telemetry/opentelemetry-php/pull/422#discussion_r712628145_

Blacksmoke16:
* Follow's PSR naming conventions
  * Add `Interface` suffix to interfaces
  * Add `Abstract` prefix to abstract classes
  * Add `Trait` suffix to traits
* Re-organize source code files into `src/`
  * `proto` directory was kept as is since that's technically not source code since it's generated
  * Update `composer.json` autoloading definitions
  * Update CI related stuff now that everything isn't in the root directory

Resolves #425


src/SDK/Metrics/Providers/GlobalMeterProvider.php
```diff
@@ -2,17 +2,19 @@
...
-namespace OpenTelemetry\Sdk\Metrics\Providers;
+namespace OpenTelemetry\SDK\Metrics\Providers;
 
-use OpenTelemetry\Metrics as API;
+use OpenTelemetry\API\Metrics as API;
 
 /**
  * https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#global-meter-provider
+ *
+ * @method static API\MeterInterface getMeter(string $name)
```
Blacksmoke16, line 12: This allows IDEs and such to be aware of the method even if it's not defined. Allowed me to remove the ignored error in the PHPStan config.

bobstrecansky (approved):

PR #437, opened 2021-10-05
Baggage propagation

### LINKED ISSUE #429: Support Baggage propagation
#410 laid the ground work but some work still needs to be done:
  * Create a new `TextMapPropagator`
  * Create a new `CompositeTextMapPropagator` type to wrap Trace and Baggage implementations
    * https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/context/api-propagators.md#composite-propagator
  * Move types to the API layer and expose global propagator types
    * https://github.com/open-telemetry/opentelemetry-specification/blob/v1.6.1/specification/context/api-propagators.md#global-propagators

Depends on #428

Blacksmoke16:
* Moves `Baggage` logic into the API namespace
  * As per [this](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/context/api-propagators.md#propagators-distribution), given I think they'll be pretty commonly used
    * > MAY alternatively be distributed as part of the OpenTelemetry API.
* Move `Propagation` related types into `Context` namespace
* Implement no-op and multi implementations of `TextMapPropagatorInterface`
  * As per [this](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/context/api-propagators.md#composite-propagator) and [this](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/context/api-propagators.md#global-propagators):
    * >  Implementations MUST offer a facility to group multiple Propagators from different cross-cutting concerns in order to leverage them as a single entity.
    * > The OpenTelemetry API MUST use no-op propagators unless explicitly configured otherwise

The spec says:

> The OpenTelemetry API MUST provide a way to obtain a propagator for each supported `Propagator` type.

Related to one of the above bullet point, it also says **unless explicitly configured otherwise** in regards to which propagators to use by default. So the main thing left to answer is how we want to go about handling configuring the defaults and exposing global instances of these propagators (and `TracerProvider`):

> Thus, the API SHOULD provide a way to set/register and access a global default `TracerProvider`.

Either way, wanted to get a draft opened to start getting some feedback.

Resolves #429


src/SDK/Trace/TraceState.php
```diff
@@ -146,7 +146,7 @@ private function parse(string $rawTracestate): array
             }
 
             foreach ($listMembers as $listMember) {
-                $vendor = explode(self::LIST_MEMBER_KEY_VALUE_SPLITTER, $listMember);
+                $vendor = explode(self::LIST_MEMBER_KEY_VALUE_SPLITTER, trim($listMember));
```
Blacksmoke16, line 149: This fixes the tracestate test case that has spaces within `BaggagePropatagorTest`.

src/API/Baggage/Propagation/Parser.php
```diff
@@ -0,0 +1,68 @@
...
+        foreach (explode(',', $this->baggageHeader) as $baggageString) {
+            if (empty(trim($baggageString))) {
+                continue;
+            }
+
+            $explodedString = explode(';', $baggageString, 2);
+
+            $keyValue = trim($explodedString[0]);
+
+            if (empty($keyValue) || !str_contains($keyValue, '=')) {
```
bobstrecansky, line 41: Is there going to be a time where the second conditional here rings true?  Should we potentially add additional tests for these?

src/API/Trace/Propagation/TraceContextPropagator.php
```diff
@@ -22,20 +27,31 @@
...
+    {
+        if (null === self::$instance) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
     /** {@inheritdoc} */
-    public static function fields(): array
```
bobstrecansky, line 32: What's the purpose of setting these as static functions?  Are you worried about errant constructors getting set here or something of the like?

src/Context/Propagation/MultiTextMapPropagator.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @no-named-arguments
+     *
+     * @param list<TextMapPropagatorInterface> $propagators
+     */
+    public function __construct(array $propagators)
+    {
+        $this->propagators = $propagators;
+        $this->fields = array_values(array_unique(array_merge(...array_map(static fn (TextMapPropagatorInterface $propagator) => $propagator->fields(), $propagators))));
```
bobstrecansky, line 36: This line reads very heavy to me.  Do you think it mike make sense to break it up a bit?

src/Context/Propagation/NoopTextMapPropagator.php
```diff
@@ -0,0 +1,35 @@
...
+        return [];
+    }
+
+    public function extract($carrier, PropagationGetterInterface $getter = null, Context $context = null): Context
+    {
+        return $context ?? Context::getRoot();
+    }
+
+    public function inject(&$carrier, PropagationSetterInterface $setter = null, Context $context = null): void
+    {
```
bobstrecansky, line 33: Is this missing a function body?

tests/API/Unit/Baggage/Propagation/BaggagePropagatorTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $context = $propagator->extract(['baggage' => $header]);
+
+        $this->assertEquals(
+            $expectedBaggage,
+            Baggage::fromContext($context)
+        );
+    }
+
+    public function headerProvider(): array
```
bobstrecansky, line 74: Nice helper function! TIL about `@dataProvider` in phpunit; that would have been super convenient to know about a bunch of different times 😺

bobstrecansky (commented):
A couple comments; no huge blockers I don't think.  Beautiful test coverage.

src/API/Baggage/Propagation/BaggagePropagator.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Baggage\Propagation;
+
+use OpenTelemetry\API\Baggage\Baggage;
+use OpenTelemetry\API\Baggage\BaggageBuilderInterface;
+use OpenTelemetry\API\Baggage\Entry;
```
bobstrecansky, line 9: phan in CI seems to believe this isn't used.
https://github.com/open-telemetry/opentelemetry-php/pull/437/checks?check_run_id=3798402726#step:9:47

src/API/Baggage/Propagation/Parser.php
bobstrecansky: Also, phan was complaining about `str_contains()` without a polyfill:
https://github.com/open-telemetry/opentelemetry-php/pull/437/checks?check_run_id=3798402726#step:9:48

src/Context/Propagation/NoopTextMapPropagator.php
Blacksmoke16: Mm, it's a no-op injector, so what is there to do?

src/API/Trace/Propagation/TraceContextPropagator.php
Blacksmoke16: They were previously static functions since they handled injecting/extracting in a stateless manner. However, with the introduction of the multi processor, you could have more than one instantiation with different combination of propagators. For this reason i made the interface require them to be instance methods.

src/API/Baggage/Propagation/Parser.php
Blacksmoke16: It's handled via the `valid complex value with some invalid` case. When it splits, the third baggage value, `a sdf9asdf-alue1` ends up being `$keyValue` which isn't valid since there is no key.

The spec does say:

> Parsers MUST NOT assume that the equal sign is only used to separate key and value.

So might not hurt to add some additional cases where there are extraneous `=`.

src/API/Baggage/Propagation/Parser.php
```diff
@@ -0,0 +1,68 @@
...
+namespace OpenTelemetry\API\Baggage\Propagation;
+
+use function explode;
+use OpenTelemetry\API\Baggage\BaggageBuilderInterface;
+use OpenTelemetry\API\Baggage\Metadata;
+use function str_contains;
+use function str_replace;
+use function trim;
+
+final class Parser
```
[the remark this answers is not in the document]
Blacksmoke16: I copied the excluded characters array from the Java implementation, so I'm pretty sure this covers it all.

src/API/Baggage/Propagation/BaggagePropagator.php
```diff
@@ -0,0 +1,89 @@
...
+use OpenTelemetry\Context\Propagation\PropagationSetterInterface;
+use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
+use function rtrim;
+
+/**
+ * @see https://www.w3.org/TR/baggage
+ */
+final class BaggagePropagator implements TextMapPropagatorInterface
+{
+    public const BAGGAGE = 'baggage';
```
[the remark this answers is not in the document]
Blacksmoke16: it's the baggage equivalent to the `TRACEPARENT` and `TRACESTATE` headers in `TraceContextPropagator`, which basically represent the name of the header those propagators will inject/extract.

src/Context/Propagation/MultiTextMapPropagator.php
Blacksmoke16: Gotta love PHP :sweat_smile:.  I'm open to alternative suggestions on how to get an array of unique values after concatenating a series of arrays.

I can move it to a private method.

src/Context/Propagation/NoopTextMapPropagator.php
bobstrecansky: Oh, duh! Thanks for the explanation 🙂

src/Context/Propagation/MultiTextMapPropagator.php
bobstrecansky: I may just be nitpicking here.  Fine to leave it as is.

bobstrecansky (approved):

Issue #431, opened 2021-09-30
Use `4318` as the default OLTP/HTTP port


Blacksmoke16:
The default HTTP OTLP port is now `4318`.

Ref: https://github.com/open-telemetry/opentelemetry-specification/pull/1970.

PR #438, opened 2021-10-07
Update OTLP/HTTP Default Port Per Changes to the Spec

Grunet:
Resolves [#431](https://github.com/open-telemetry/opentelemetry-php/issues/431)

My biggest concern has to do with the 2nd commit, in particular b/c I couldn't find evidence in the spec of a non-protobuf-ed HTTP transport mechanism, which it seemed like Contrib/Otlp/Exporter.php is trying to do (sending an ordinary HTTP request, maybe with a JSON body). So I wasn't sure if updating the default port was appropriate in this case.

A smaller concern also has to do with the 2nd commit, namely if updating the port on the collector container in the compose file was OK (seemed to me like it was per [this PR](https://github.com/open-telemetry/opentelemetry-collector/pull/3743) from the collector repo)


Grunet:
Not sure I quite understand why the first codecov/project check failed, but may be missing something subtle

Grunet:
Resolves #431

Issue #426, opened 2021-09-29
Revamp examples


Blacksmoke16:
As a result of #422, examples will need to be updated to work with the new API. A lot of them are also very similar. IMO it would be better to have the examples show off specific features, like a simple one, child spans, etc. Instead of every type of exporter getting its own example.

_Originally posted by @Blacksmoke16 in https://github.com/open-telemetry/opentelemetry-php/pull/422#discussion_r718126890_

tidal:
In addition it would be good to use snake case for the file names of the examples, as camel case is usually only used for classes, interfaces and traits.

brettmc:
It seems to me that most of the original examples are really developer-focussed integration/regression tests - any thoughts on moving them out of `examples` and into `regression` or something like that? (happy to do this, and get them working again, if they're still useful and required).

Blacksmoke16:
There's also an `Integration` namespaces in `tests/SDK` that could be a good place for some of these types of examples.

I still think the examples would be better served showing how to use higher level concepts of the library (e.g. simple versus parent/child) instead needing one for every tiny feature of the lib. Possibly this would also be a good place for textual docs that can just talk about those features versus needing an entirely new example file that's 95% the same, just with a diff exporter/extra config/data on the spans.

tidal:
@Blacksmoke16 
I agree. The exporter setup is kind of straightforward, especially when one has basic understanding of distributed tracing. Other parts of the SDK/API/Spec are not as intuitive to understand. So for the SF bundle I had to dig through a lot of code
and spec docs to make things work. There is nothing wrong with reading the spec (while it's not always an easy read for non native English speakers), but in the best case someone who builds an instrumentation on top of the SDK, can rely on the SDK documentation itself.
The problem is probably 99% of developers (including me) prefer writing code over writing docs, while most prefer to read good doumentation over digging through code. And keeping docs up to date in a pre alpha state can be kind of a tedious.

lalex:
Could we rework or totally remove examples named `AlwaysOn...` as they show bad pattern of exposing `Sampler`s implementation? Client's code should not rely on sampling decisions made by samplers.

I've noticed it while browsing the source of the Symfony bundle: https://github.com/tidal/otel-sdk-bundle-example-sf5/blob/6b84645171d7f4fac06c82ec857291fb58ffecad/src/EventSubscriber/OtelKernelSubscriber.php#L193

tidal:
[...]

PR #455, opened 2021-10-29
docs/updated examples

brettmc:
Adding some current examples which illustrate some features not contained in other examples (resources, http attributes), which partially addresses https://github.com/open-telemetry/opentelemetry-php/issues/426;
Adding more resource constants per the opentelemetry-collector spec.
Updating ConsoleExporter to output span resources, and compacting span attributes.


bobstrecansky (approved):
LGTM!  Thank you for your contribution!

PR #456, opened 2021-10-30
Add context detach out-of-order detection and support for execution context switching

Nevay:
#### Detach out-of-order detection:
Detaching a context returns a non-zero status if the detached context is not the current instance (see [detach-context](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/context/context.md#detach-context)).
Out-of-order behavior: this PR detaches only the associated context, an alternative would be to detach all intermediate contexts to ensure that scope detach restores the previous context (if the detached scope belongs to the current execution context).

The SDK should decorate the scopes to log non-zero return values.

#### Execution context switching (fibers):
Adds methods to `fork`, `switch` and `destroy` the context associated with an execution context.

Note that fiber support requires hooking into fiber events - as far as I know they are not available to userland, which means we cannot provide an implementation that works out-of-the-box.
This PR requires the FFI-extension and preloading the header file to enable fiber support.


tidal:
> Note that fiber support requires hooking into fiber events - as far as I know they are not available to userland

PHP8.1 will ship with [fibers](https://wiki.php.net/rfc/fibers) and there is an extension ([ext-fiber](https://github.com/amphp/ext-fiber)) which adds fiber support to PHP, however the extension only works with PHP8.0

bobstrecansky:
There were a couple of Phan errors in this PR.  You can validate locally with `make phan`.

We are going to want to support PHP 7.4 until at least EOL, so it'd be advantageous to not use use new language features unless there is an absolute need.

Nevay:
This PR is compatible with PHP 7.4, the fiber specific code is only used if the `Fiber` class exists.
I've supressed the undeclared class errors for now as the code checks for class existance, `make phan` succeeds locally, I'll try to fix the pipeline later.

bobstrecansky (approved):

PR #459, opened 2021-11-01
Bugfix/batch processor

brettmc:
The spec for `SpanProcessor::shutdown()` says 'MUST include the effects of `ForceFlush`', so implement that logic and add a test (also updated a couple of other tests to add a new expectation on forceFlush)
The spec for `SpanProcessor::forceFlush()` also says that built-in span processors MUST call `forceFlush()` on the exporter.
Bugfix: `SpanInterface::scheduledDelayMillis` clearly accepts milliseconds, but time checking arithmetic uses nanoseconds. Adding a constant to convert between the two, and unit tests.

### DIFF
--- a/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -102,7 +102,7 @@ class BatchSpanProcessor implements SpanProcessorInterface
         }
 
         if (null !== $this->exporter) {
-            return $this->forceFlush() && $this->exporter->shutdown();
+            $this->forceFlush() && $this->exporter->shutdown();
         }
         $this->running = false;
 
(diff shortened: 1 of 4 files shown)


src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -100,10 +101,10 @@ public function shutdown(): bool
             return true;
         }
 
-        $this->running = false;
         if (null !== $this->exporter) {
             return $this->forceFlush() && $this->exporter->shutdown();
         }
+        $this->running = false;
```
brettmc, line 107: NB this is the important change for allowing flush on shutdown - setting `shutdown` too early meant that the forceFlush was a no-op

bobstrecansky (approved):

Issue #299, opened 2021-03-31
Implement Error Handling: Error Handlers must be configurable


jodeev:
See [Error Handling spec](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/error-handling.md#error-handling-in-opentelemetry) -- [Configuring Error Handlers](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/error-handling.md#configuring-error-handlers). 

Basic overview (refer to the spec):

SDK implementations MUST allow end users to change the library's default error handling behavior for relevant errors. Application developers may want to run with strict error handling in a staging environment to catch invalid uses of the API, or malformed config. 

Note that configuring a custom error handler in this way is the only exception to the basic error handling principles outlined above. The mechanism by which end users set or register a custom error handler should follow language-specific conventions.

bobstrecansky:
We need to have a central entry point in order to have global error handling.

bobstrecansky:
This is configurable enough because it uses a PSR-3 Logger (this is similar to what Java does)

Issue #298, opened 2021-03-31
Implement Error Handling: Self-Diagnostics


jodeev:
See the [Error Handling](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/error-handling.md#error-handling-in-opentelemetry) in the spec, [Self-Diagnostics](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/error-handling.md#self-diagnostics). 

All OpenTelemetry libraries -- the API, SDK, exporters, instrumentations, etc. -- are encouraged to expose self-troubleshooting metrics, spans, and other telemetry that can be easily enabled and filtered out by default.

Whenever the library suppresses an error that would otherwise have been exposed to the user, the library SHOULD log the error using language-specific conventions. SDKs MAY expose callbacks to allow end users to handle self-diagnostics separately from application code.

tidal:
Related: https://github.com/open-telemetry/opentelemetry-php/pull/460#issuecomment-962429788

kpratyus:
I have started looking into the possible implementations for this issue. I had a discussion with @tidal in the last sig and there were few suggestions on how we could proceed. So I would be glad if this issue could be assigned to me.
Thanks

tidal:
@stale: 滚!

bobstrecansky:
@kpratyus are you still planning on finishing this?

bobstrecansky:
@brettmc mentioned looking at the [PSR](https://www.php-fig.org/psr/psr-14/) for this.

brettmc:
I think this is now done - we report on errors now. We can emit more errors in future, probably based on beta release feedback.

PR #460, opened 2021-11-04
Add Logger SpanExporter, Logger ExporterDecorator and other logging related components.

tidal:
This PR adds observabilty to the library itself by introducing several logging related components. Logging functionality is based on the [PSR-3: Logger standard](https://www.php-fig.org/psr/psr-3/) and a basic mapping between PSR3 and [OpenTelemetry SeverityNumbers](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/logs/data-model.md#field-severitynumber) is added, too (PSR uses [rfc5424 ](https://datatracker.ietf.org/doc/html/rfc5424#page-10) log levels. )

Only the PSR3 interface library is added as a production dependency, so Developers can use a PSR3 implementation of their liking, however the popular choice [Monolog ](https://github.com/Seldaek/monolog)is added as a dev dependency, and a very simple PSR3 implementation is provided, too. 

This PR doesn't take care of log enrichment with trace data, etc. according to the spec, but can be a base for future development.

Furthermore, according to the [specs on error handling](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/error-handling.md#basic-error-handling-principles) "OpenTelemetry implementations MUST NOT throw unhandled exceptions at run time." So components of this PR can help avoid leaking exceptions from the library in the future, while still proving feedback for developers via logs. This can be done by implemeting the PSR "LoggerAware" interface by using a trait. Many frameworks automatically inject a logger instance in classes implementing said interface.

- Adds PSR3 and Monolog libaries (prod / dev)
- Adds VfsStream as dev library to allow mocking of the file system in tests.
- Adds a SpanConverter interface to allow logging of SpanData in different formats.
- Adds a Noop SpanConverter class to avoid runtime errors.
- Extracts friendly JSON converter from ConsoleExporter into dedicated SpanConverter class for reusability.
- Adds a LoggerSpanExporter for debugging/testing purposes which uses the friendly JSON converter as a default.
- Adds an abstract SpanExporterDecorator, which can be extended to enhance/filter Span data or report on it.
- Adds a LoggerDecorator which reports Span data to a given Logger.
- Adds a simple PSR file logger (while a file can be a lot of things in PHP)
- Adds Mapping between PSR and OTEL log/severity levels.
- Adds examples for the usage of LoggerExporter and LoggerDecorator

### DIFF
--- a/src/SDK/Logs/SimplePsrFileLogger.php
+++ b/src/SDK/Logs/SimplePsrFileLogger.php
@@ -15,16 +15,21 @@ class SimplePsrFileLogger implements LoggerInterface
 {
     use LoggerTrait;
 
+    private const DEFAULT_LOGGER_NAME = 'otel';
+
     private static ?array $logLevels = null;
 
     private string $filename;
 
+    private string $loggerName ;
+
     /**
      * @param string $filename
      */
-    public function __construct(string $filename)
+    public function __construct(string $filename, string $loggerName = self::DEFAULT_LOGGER_NAME)
     {
         $this->filename = $filename;
+        $this->loggerName = $loggerName;
     }
 
     /**
@@ -61,8 +66,9 @@ class SimplePsrFileLogger implements LoggerInterface
         }
 
         return sprintf(
-            '[%s] %s: %s %s%s',
+            '[%s] %s %s: %s %s%s',
             date(DATE_RFC3339_EXTENDED),
+            $this->loggerName,
             $level,
             $message,
             $encodedContext,
--- a/src/SDK/Logs/SimplePsrFileLogger.php
+++ b/src/SDK/Logs/SimplePsrFileLogger.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Logs;
+
+use Psr\Log\InvalidArgumentException;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerTrait;
+use Psr\Log\LogLevel;
+use ReflectionClass;
+use Throwable;
+
+class SimplePsrFileLogger implements LoggerInterface
+{
+    use LoggerTrait;
+
+    private static ?array $logLevels = null;
+
+    private string $filename;
+
+    /**
+     * @param string $filename
+     */
+    public function __construct(string $filename)
+    {
+        $this->filename = $filename;
+    }
+
+    /**
+     * @param string $level
+     * @param string $message
+     * @param array $context
+     * @psalm-suppress MoreSpecificImplementedParamType
+     */
+    public function log($level, $message, array $context = []): void
+    {
+        $level = strtolower($level);
+
+        if (!in_array($level, self::getLogLevels(), true)) {
+            throw new InvalidArgumentException(
+                sprintf('Invalid Log level: "%s"', $level)
+            );
+        }
+
+        file_put_contents($this->filename, $this->formatLog((string) $level, (string) $message, $context), FILE_APPEND);
+    }
+
+    /**
+     * @param string $level
+     * @param string $message
+     * @param array $context
+     * @return string
+     */
+    private function formatLog(string $level, string $message, array $context = []): string
+    {
+        try {
+            $encodedContext = json_encode($context, JSON_THROW_ON_ERROR);
+        } catch (Throwable $t) {
+            $encodedContext = sprintf('(Could not encode context: %s)', $t->getMessage());
+        }
+
+        return sprintf(
+            '[%s] %s: %s %s',
+            date(DATE_RFC3339_EXTENDED),
+            $level,
+            $message,
+            $encodedContext
+        );
+    }
+
+    /**
+     * @return array
+     */
+    private static function getLogLevels(): array
+    {
+        return self::$logLevels ?? self::$logLevels = (new ReflectionClass(LogLevel::class))->getConstants();
+    }
+}
--- a/src/SDK/Logs/SimplePsrFileLogger.php
+++ b/src/SDK/Logs/SimplePsrFileLogger.php
@@ -61,11 +61,12 @@ class SimplePsrFileLogger implements LoggerInterface
         }
 
         return sprintf(
-            '[%s] %s: %s %s',
+            '[%s] %s: %s %s%s',
             date(DATE_RFC3339_EXTENDED),
             $level,
             $message,
-            $encodedContext
+            $encodedContext,
+            PHP_EOL
         );
     }
 
--- a/src/SDK/Trace/Behavior/LoggerAwareTrait.php
+++ b/src/SDK/Trace/Behavior/LoggerAwareTrait.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Trace\Behavior;
+
+use Psr\Log\LoggerAwareTrait as PsrTrait;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use Psr\Log\NullLogger;
+
+trait LoggerAwareTrait
+{
+    use PsrTrait;
+
+    private string $defaultLogLevel = LogLevel::INFO;
+
+    /**
+     * @param string $logLevel
+     */
+    public function setDefaultLogLevel(string $logLevel): void
+    {
+        $this->defaultLogLevel = $logLevel;
+    }
+
+    /**
+     * @param string $message
+     * @param array $context
+     * @param string|null $level
+     */
+    protected function log(string $message, array $context = [], ?string $level = null): void
+    {
+        $this->getLogger()->log(
+            $level ?? $this->defaultLogLevel,
+            $message,
+            $context
+        );
+    }
+
+    protected function getLogger(): LoggerInterface
+    {
+        return $this->logger instanceof LoggerInterface ? $this->logger : $this->logger = new NullLogger();
+    }
+}
--- a/src/SDK/Trace/Span.php
+++ b/src/SDK/Trace/Span.php
@@ -21,6 +21,86 @@ use Throwable;
 
 final class Span extends API\AbstractSpan implements ReadWriteSpanInterface
 {
+
+    /** @readonly */
+    private API\SpanContextInterface $context;
+
+    /** @readonly */
+    private API\SpanContextInterface $parentSpanContext;
+
+    /** @readonly */
+    private SpanLimits $spanLimits;
+
+    /** @readonly */
+    private SpanProcessorInterface $spanProcessor;
+
+    /**
+     * @readonly
+     *
+     * @var list<API\LinkInterface>
+     */
+    private array $links;
+
+    /** @readonly */
+    private int $totalRecordedLinks;
+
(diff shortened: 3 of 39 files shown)


bobstrecansky (approved):
This is so exciting!  Thank you for your hard work on this.

bobstrecansky:
One style issue was caught in the CI

tidal:
@bobstrecansky

My pleasure, and CS should be fixed now.

I'm getting some psalm errors locally with the OtlpGrpc/Exporter despite the grpc extension being installed (UndefinedClass). I hope the don't show up here, since I did not actually touch that code.

tidal:
Hmm, that's interesting, while the coverage for my PR is 100%, they overall coverage drops by 0.69%. 
I did not remove any tests, but I rewrote some of tests of code I was touching with mocking away the dependencies.
So I guess the coverage before was false positive, because of implicit "testing". 
It's a matter of opinion, but tbh for me personally coverage is a non metric  for unit tests if dependencies are not mocked. Otherwise it's very easy to paint everything green, it doesn't even need assertions.

However I honestly like writing tests, so wait a bit with merging, please, I will make up for it, not a problem at all.

src/SDK/Logs/SimplePsrFileLogger.php
```diff
@@ -0,0 +1,80 @@
...
+namespace OpenTelemetry\SDK\Logs;
+
+use Psr\Log\InvalidArgumentException;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerTrait;
+use Psr\Log\LogLevel;
+use ReflectionClass;
+use Throwable;
+
+class SimplePsrFileLogger implements LoggerInterface
```
brettmc, line 14: in a container world (docker, k8s), logging to stdout and stderr is the norm. Since you can file_put_contents to php://stdout, this could become `SimplePsrStreamLogger`, and default to stdout?

src/SDK/Trace/Span.php
```diff
@@ -21,6 +21,86 @@
 
 final class Span extends API\AbstractSpan implements ReadWriteSpanInterface
 {
+
```
brettmc, line 24: is this a formatting/EOL change? The code doesn't look different

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -0,0 +1,44 @@
...
+use Psr\Log\LoggerAwareTrait as PsrTrait;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use Psr\Log\NullLogger;
+
+trait LoggerAwareTrait
+{
+    use PsrTrait;
+
+    private string $defaultLogLevel = LogLevel::INFO;
```
brettmc, line 16: good time to implement `OTEL_LOG_LEVEL` per spec? (default is `info` if not set)

brettmc:
Any ideas yet on how to make the logger accessible throughout the codebase? Wanting access to it pretty much everywhere through the SDK says that it's a static instance somewhere, possibly provided as an optional when constructing a TracerProvider and stored in TracerSharedState?

src/SDK/Trace/Span.php
tidal: Don't really know. It's LF line breaks and I don't think it was different before.  That's probably something the CS fixer did, I just changed StatusData to StatusDataInterface in the file.

src/SDK/Logs/SimplePsrFileLogger.php
tidal: To be honest I'd rather rename this logger to `DoNotUseInProductionSeriouslyLogger`. It is just meant to be used for demo/example purposes or at max local development. It has some serious design flaws. Logging to stdout/err is easy and comfortable, but it hides a lot of complexity, as well as K8S does, and at the end of the day even "high availabilty" only means the system doesn't crash, just because you look at it. However this little logger implementation has the potential to create an indefinit amount of backpressure, as it sends everything to the configured destination, The log level is just a "tag" for it.
So I'd rather not recommend anyone to use it in any serious application, but use a battle tested and well maintained solution like Monolog, which comes with a lot of configuration options, diferrent [handlers, filters](https://github.com/Seldaek/monolog/blob/main/doc/02-handlers-formatters-processors.md), and so on.

Also this logger will not make it easier to implement enrichment of logs with tracing data, etc. in the future. It's easier to provide contrib plugins for popular thirds party loggers, than to reinvent the wheel in that case, especially because people will use those libraries anyway.

But these are just my thoughts, feel free to change the logger once the PR is merged.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: Hmm, the default is 'info' at the moment, so that's according to spec (by accident). 
However if you mean, the trait/class should pull the value of the env variable `OTEL_LOG_LEVEL` out of the global scope, that's not something I want to implement. I am personally a strong believer in the Single Responsibilty Principle. So a class should be stupid and concentrate on it's job. It should not even know env vars exist, unless its job is to deal with env vars. I would prefer a solution where such config variables are resolved by a dedicated component and are then populated to the places which need them.

But again, these are just my thoughts.

Grunet:
This looks really awesome! And I think there are also 3 existing issues in the repo I'm wondering if this in part (or fully?) addresses and can maybe be closed as a part of this 

- Allow for injecting a custom, consumer-created error handler, log manager, etc... - #299 
- Related user feedback around being more verbose in a "debug mode" - #219 
- Self-diagnostics of perf, errors, logs, etc... - #298

src/SDK/Logs/SimplePsrFileLogger.php
```diff
@@ -0,0 +1,80 @@
...
+    private function formatLog(string $level, string $message, array $context = []): string
+    {
+        try {
+            $encodedContext = json_encode($context, JSON_THROW_ON_ERROR);
+        } catch (Throwable $t) {
+            $encodedContext = sprintf('(Could not encode context: %s)', $t->getMessage());
+        }
+
+        return sprintf(
+            '[%s] %s: %s %s%s',
```
brettmc, line 64: if the logs were going to a destination that also contains other logs, it would be good to have an indication that the log was from `otel-php`...?

src/SDK/Trace/Behavior/LoggerAwareTrait.php
brettmc: That makes sense - I'm thinking ahead to when I am instrumenting code (we have literally a couple hundred small to mid-sized applications), where I want the library to configure itself as much as possible with environment variables. But I agree that it's not the responsibility of this logger to care about that.

src/SDK/Logs/SimplePsrFileLogger.php
tidal: Yap, sure, not a problem. I will just add a "name" parameter like Monolog has, that defaults to 'otel'.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: Ok, got you. I thought about something like that as well. But a lot of frameworks have their own way how to deal with env vars, which have to be taken into account as well. It's probably a bit out of the scope of this PR, we can deal with this later.

tidal:
@brettmc
> Any ideas yet on how to make the logger accessible throughout the codebase? Wanting access to it pretty much everywhere through the SDK says that it's a static instance somewhere, possibly provided as an optional when constructing a TracerProvider and stored in TracerSharedState?

I thought only a bit about it,  however I like your idea. I actually just wanted to have a way to debug the problems with the exporter factories and some things for the Symfony InstrumentationBundle. So if you'd like to put more thought into it and/or work on it, go for it.

tidal:
> This looks really awesome! And I think there are also 3 existing issues in the repo I'm wondering if this in part (or fully?) addresses and can maybe be closed as a part of this
> 
> * Allow for injecting a custom, consumer-created error handler, log manager, etc... - [Implement Error Handling: Error Handlers must be configurable #299](https://github.com/open-telemetry/opentelemetry-php/issues/299)
> * Related user feedback around being more verbose in a "debug mode" - [User Research Feedback Improvements - Missing Exception #219](https://github.com/open-telemetry/opentelemetry-php/issues/219)
> * Self-diagnostics of perf, errors, logs, etc... - [Implement Error Handling: Self-Diagnostics #298](https://github.com/open-telemetry/opentelemetry-php/issues/298)


Yes, I have seen those issues as well. That's why instead of just creating something quick for myself for debugging, I figured it would be nice to have something more reusable for the library and library development.

tidal:
@brettmc
Btw, this could also get rid of throwing exceptions in the library at all (no matter if they are caught), as exceptions are quite time-consuming (Actually throwing of exceptions is not time-consuming, but the calculation of the stack trace. I know, because I had the stupid idea in the past to use exceptions as kind of events and made some measurements)

brettmc:
> @brettmc Btw, this could also get rid of throwing exceptions in the library at all (no matter if they are caught), as exceptions are quite time-consuming (Actually throwing of exceptions is not time-consuming, but the calculation of the stack trace. I know, because I had the stupid idea in the past to use exceptions as kind of events and made some measurements)

Yes, I was looking at the [error handling](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md) part of the spec recently around `the library SHOULD log the error`, and so having easy access to the logger will help a lot there, IMO. I also thought I read something about stdout being the default, but now I think that I misinterpreted "Standard output (logging)" in the exporters section of the spec compliance matrix.

tidal:
@bobstrecansky 
Ok, try again please.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: @bobstrecansky
Are there any plans (or active development) regarding configuration of the SDK? The configuration of the SdkBundle could be extracted out for this purpose (with some changes) and this could also take care of the whole env var issue.

tidal:
One more this regarding error handling in general and https://github.com/open-telemetry/opentelemetry-php/issues/219, https://github.com/open-telemetry/opentelemetry-php/issues/298, https://github.com/open-telemetry/opentelemetry-php/issues/299.

While logging is nice, to much logging does not always help with finding issues, especially in development, where you may not use sophisticated tools like Kibana (or any other). Also logging can slow down performance, even the check if something should be logged.

As a handy tool for this regard PHP has the [assert](https://www.php.net/manual/en/function.assert.php) and [assert_options](https://www.php.net/manual/en/function.assert-options.php). The benefit of `assert` is, that it is a noop in PHP production configuration, while it will raise notices or exceptions in dev configuration.  Also an `assert_callback` can be configured, while this has to be taken with care, as it's a global callback.

In essence the following code would raise a notice or exception (depending on configuration) in development, while it will be ignored in production.

```php

assert(1 === 2);
```

I have changed the PHPUnit config, to turn the `assert` on during tests.

tidal:
@bobstrecansky
Looks good for me now, If you agree, you can merge it.

Issue #451, opened 2021-10-15
Exporter factory, factory methods and creation could use some rework/cleanup


tidal:
**Is your request related to a problem?**

At the moment the creation of span exporters (especially through the factory methods and the factory) class is inconsistent in behavior and in some scenarios broken. However in downstream instrumentations (for example Symfony Bundle) a generic way to instantiate exporters is needed to allow configuration.

1. ~~"fromConnectionString" factory methods in [Jaeger](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Jaeger/Exporter.php#L14), [Newrelic](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Newrelic/Exporter.php#L170), [OtlpHttp](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/OtlpHttp/Exporter.php#L221), [Zipkin ](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Zipkin/Exporter.php#L134)and [ZipkinToNewrelic ](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/ZipkinToNewrelic/Exporter.php#L158) exporters have a dependency on Guzzle library and will fail if other PSR7 implementations are used. see: [448](https://github.com/open-telemetry/opentelemetry-php/issues/)~~
2. "fromConnectionString" factory method in [OtlpGrpc ](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/ZipkinToNewrelic/Exporter.php#L158) and [OtlpHttp](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/OtlpHttp/Exporter.php#L221) exporters silently drop the passed "$endpointUrl" argument without using it and/or raising an error,
3.  [OtlpHttp](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/OtlpHttp/Exporter.php#L83) exporter provides no way of configuring the endpoint (and a bunch of other options) programmatically, the only way is to provide a env var or using the default.
[...]

Issue #448, opened 2021-10-09
Library has a dependency on PSR7/PSR17 interfaces - Exporter factory methods have a dependency on an concrete implementation (Guzzle)


tidal:
**Describe your environment** 
This bug occures in the 0.0.3 release as well as in current dev-main

**Steps to reproduce**
Run:
`composer require open-telemetry/opentelemetry nyholm/psr7 kriswallsmith/buzz`
Code:
`$factory = new OpenTelemetry\Sdk\Trace\ExporterFactory('foo)');`
`$exporter = $factory->fromConnectionString('jaeger+http://localhost:9412/api/v2/spans');`
or 
`$exporter = OpenTelemetry\Contrib\Jaeger\Exporter::fromConnectionString('http://localhost:9412/api/v2/spans');`
or trying to create other exporters using a PSR7 client in the same way.

**What is the expected behavior?**
An instance of an Exporter should be created

**What is the actual behavior?**
`PHP Fatal error:  Uncaught Error: Class "GuzzleHttp\Psr7\HttpFactory" not found`

**Additional context**
This issues can be fixed by using the [php-http/discovery](https://github.com/php-http/discovery) composer package.
The resulting code would look something like this:
```
[...]
$exporter = new JaegerExporter(
    'foo',
    'jaeger+http://localhost:9412/api/v2/spans',
    HttpClientDiscovery::find(),
    Psr17FactoryDiscovery::findRequestFactory(),
    Psr17FactoryDiscovery::findStreamFactory(),
);
```

Blacksmoke16:
Related: https://github.com/open-telemetry/opentelemetry-php/issues/296.

Jaeger exporter needs extracted into its own type, versus relying on Zipkin.

tidal:
@Blacksmoke16 
You might want to take a look at https://github.com/open-telemetry/opentelemetry-php/issues/451 where I outlined some of the issues with the factory. I'd appreciate your feedback/ideas.

tidal:
Also it might be a good idea to create a dedicated http client/wrapper as suggested in https://github.com/open-telemetry/opentelemetry-php/issues/223 to only have one dependency in the http based exporters. 
[...]

PR #461, opened 2021-11-05
Fixing issues around jaeger-exporter by adding the guzzlehttp 

aabri-g4m:
This PR fixed issues around jaeger-exporter by adding the guzzlehttp to the required dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,9 @@
         "google/protobuf": "^v3.3.0",
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
-        "nyholm/dsn": "^2.0.0"
+        "nyholm/dsn": "^2.0.0",
+        "guzzlehttp/guzzle": "^7.3",
+        "guzzlehttp/psr7": "^2.1.0"
     },
     "config": {
         "sort-packages": true
@@ -51,8 +53,6 @@
     "require-dev": {
         "composer/xdebug-handler": "^2.0",
         "friendsofphp/php-cs-fixer": "^3.0",
-        "guzzlehttp/guzzle": "^7.3",
-        "guzzlehttp/psr7": "^2.0@RC",
         "mockery/mockery": "^1.4",
         "nyholm/psr7": "^1.4",
         "phan/phan": "^5.0",


tidal:
If you are referring to #448, the problem is having the dependeny on Guzzle in the first place. This is just a left over from moving away from Guzzle, and there are ways to solve this. There are more issues with the exporter factory and exporter instatiation which all in some way correlate. Keep in mind, I already started to work on these issues, and once I'm done, there will be no dependency in the code on Guzzle anymore.

phpboyscout:
@tidal is there a branch or repo somewhere with your current changes that could be used for R&D development purposes, we are looking to use the php-sdk going forward (aware that its pre-alpha) and it knowing the direction it is going in would be beneficial... as it stands right now as of v0.0.3 its unusable without the guzzle deps

tidal:
@phpboyscout 
> as it stands right now as of v0.0.3 its unusable without the guzzle deps

The guzzle deps in general have already been removed in the main branch (not by me), so you could just reference it  with
```yaml
"require": {
   [...]
   "open-telemetry/opentelemetry": "dev-main",
   [...]
}
```
and use any HTTP Client/Factory implementation of your choice. For example:
```yaml
"require": {
   [...]
   "open-telemetry/opentelemetry": "dev-main",
   "kriswallsmith/buzz": "^1.2",
   "nyholm/psr7": "^1.4"
   [...]
}
```

The only remaining dependency on guzzle is left in the factory methods of the exporters and the exporter factory itself.
You can take a look at the [exporter factory](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/a668bcd0e3f8bd606fb0cd1b18447438c848e7b6/instrumentation/symfony/OtelSdkBundle/Trace/ExporterFactory.php) in the Symfony SdkBundle, on the approach I will take to solve this issue.

However there are some [more issues](https://github.com/open-telemetry/opentelemetry-php/issues/451) around the creation of the exporters (which are easy to spot in hindsight, but difficult to plan for in open source development, that's just how it is). So there might be changes to the factory and the creation of the exporters itself, however I would not bet on my proposals for this issues to silently go through, as I am not a maintainer of this library, plus I don't think there is a "best" solution anyway, and there might be something I did not think of.  

To cut a long story short, I don't think it's a good idea to make bets on my WIP. However I will be done with my changes within this week I think.

However a good advice to deal with any 3rd party library dependecies (and changes in them) is to have no dependency on them at all by implementing Dependency Inversion. (If you don't know, what I mean by that, I'm happy to explain)

Hope that helps for now.

phpboyscout:
@tidal that is great, thanks for the steer, we will take that into consideration. Oddly enough the only reason this issue reared it it's head for us is that we were working with the Jaeger exporter which still has a reference to the Guzzle packages, hence the PR. 

In the mean time we can work around this and I'll ask @aabri-g4m to close the PR

tidal:
@phpboyscout

The guzzle issues should be fixed now with https://github.com/open-telemetry/opentelemetry-php/pull/472

You will need a [php-http/async-client-implementation](https://packagist.org/providers/php-http/async-client-implementation), so my example with `kriswallsmith/buzz` won't work anymore.
(The reason for the async-client is to allow non blocking calls to the trace collectors in the future, so any issues on that side won't block main application execution)

PR #463, opened 2021-11-10
Upgrade phpstan packages to stable versions

tidal:


bobstrecansky:
Shall we fix these phpstan errors in this PR as well?

tidal:
@bobstrecansky 

I guess so, I wasn't quite sure _how_ to fix the issues, since it's all unused code:

The one in the LoggerExporter is simply a left over from when I refactored the exporter and decorator. I can simply remove that property.

The other ones looked likekind of  WIP to me. So maybe I just comment them out for now and add a todo (please check it that code it needed) for now. What do you think?

bobstrecansky:
Yeah that sounds good to me!

tidal:
Ok, did so. Only in BatchSpanProcessor the unused property is a constructor argument, and BatchSpanProcessor is used in a lot of places, so I only put "@phpstan-ignore-next-line" for now.

Also had to add a little test to make the coverage happy.

bobstrecansky (approved):

Issue #464, opened 2021-11-12
Several classes violate direction of dependency between (sub) packages


tidal:
**What is the expected behavior?**
The dependency between (sub) packages should create a DAC (Directed Acyclic Graph). [According to the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#opentelemetry-client-generic-design) (sub) packages should be distributable as separate packages.

**What is the actual behavior?**
At the moment several classes violate the direction of dependency:

![dep1](https://user-images.githubusercontent.com/57825/141519805-1b30a7fb-7c16-4fcb-b239-411bf54ef059.png)

![image](https://user-images.githubusercontent.com/57825/141518177-f5d822bb-e37b-4407-b127-5a5a7d204358.png)
![image](https://user-images.githubusercontent.com/57825/141518258-c6bc2d90-d06e-4ba0-8351-3f24e90d0ec0.png)
![image](https://user-images.githubusercontent.com/57825/141518344-3ce5dbac-524a-4b6d-a86f-3d574180270c.png)

**Additional context**
The tests and images have been created by using the [Deptrac](https://github.com/qossmic/deptrac) tool (which I will create a PR for to add it to the library)
The direction of dependency is my assumption, but any other configuration creates even more issues.

PR #470, opened 2021-11-14
Move TraceContextPropagator to SDK

tidal:
Fixes 5 issues from https://github.com/open-telemetry/opentelemetry-php/issues/464

I wasn't quite sure where to put this first, especially since its test was located under `Context`. However looking at the dependencies, it only makes sense for this class to live under `SDK`.

(The remaining 12 issues from https://github.com/open-telemetry/opentelemetry-php/issues/464 are caused by the exporter factory, and I will fix them with the factory rework, but this is not as easy as moving the class to `Contrib`, expecially, when https://github.com/open-telemetry/opentelemetry-php/pull/469 is relying on it to live in under `SDK`.)


tidal:
The phan errors seem to be due to phan's latest release (5.3.0).  
I will fix them.

tidal:
I tested if what Phan is complaining about could possibly happen. It cannot! It just seems that Phan is struggling with variadic arguments in the latest version, so I just supressed the errors.

bobstrecansky (approved):

Issue #283, opened 2021-03-15
Event Attributes not accessible on both Zipkin and Jaeger for Otel-PHP examples 


prondubuisi:
Otel-php provides examples for exporting traces to both Zipkin and Jaegar, and while span traces get displayed on both visualization tools, I have noticed that `attribute` arguments passed to events via `$span->addEvent()`, do not get displayed on both tools

On the other hand, attributes set to span using `$span->setAttribute()` appear as tags.

@zsistla I am wondering if you are able to see attributes passed via `addEvent()` with `AlwaysOnZipkinToNewrelicExample`

Here are steps to reproduce the error;
on the project repository

run `make examples`
navigate to `http://localhost:9411/zipkin/` to see example traces on Zipkin
navigate to `http://localhost:16686/search` to see example traces on Jaegar

It is worthy of note that the tests for attributes on both examples pass, also I have tried tweaking the timestamps for the [recordException method](https://github.com/open-telemetry/opentelemetry-php/blob/main/sdk/Trace/Span.php#L222) using the timestamp example here https://github.com/open-telemetry/opentelemetry-php/blob/bb0bbafec1d48509171aa12be68182075ef8e459/contrib/ZipkinToNewrelic/SpanConverter.php#L56 as suggested by @bobstrecansky 

notice that the attributes from both [Zipkin](https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/AlwaysOnZipkinExample.php#L51) and [Jaegar](https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/AlwaysOnJaegerExample.php#L51) are not visible 

Below is the screenshot for both Zipkin and Jaegar

![image](https://user-images.githubusercontent.com/22311928/111207483-fb8a5600-85c9-11eb-8d5e-3a5effe3191e.png)
![image](https://user-images.githubusercontent.com/22311928/111207621-25dc1380-85ca-11eb-8694-19e07dbf8656.png)

bobstrecansky:
@prondubuisi - just wanted to follow up with you and check on the status of this issue.

Issue #415, opened 2021-09-15
Use of "iterable" Type May Need Re-Evaluation Across the Repo 


Grunet:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your php version (`php -v` will tell you your current version), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on master.

The repo opened up in Gitpod (so the dev environment?)

**Steps to reproduce**
I haven't reproduced the problem directly in the repo yet, but it was pointed out to me that there are many instances in the repo where something is typed as an "iterable" but treated like an ordinary array , which may not always function as intended.

For example, on [this occasion](https://github.com/open-telemetry/opentelemetry-php/blob/45b7750ec3decb8c4d3c5cc7359717fa2efeef7a/contrib/ZipkinToNewrelic/Exporter.php#L112), "empty" is called on the iterable type, but if that iterable is something like an exhausted generator, it counterintuitively won't return "true" (as I tried to reproduce in [this fiddle](https://replit.com/@Grunet/MistyroseKnownDesigners#main.php)) 

**What is the expected behavior?**
For the specific case above, an iterable that doesn't have any content remaining should be considered the same as an empty array, not the opposite

**What is the actual behavior?**
For the specific case above, passing in an exhausted generator would cause it to skip the initial check and proceed to the rest of the function I think

**Additional context**
There was a long discussion around part of this [in Slack](https://cloud-native.slack.com/archives/C01NFPCV44V/p1631498438088800?thread_ts=1631244100.052100&cid=C01NFPCV44V) that may be valuable to review

bhaibel:
@Grunet Reading this and the thread for your PR #412, it sounds like you believe #412 makes a dent in the issue but doesn't completely solve it. This issue is also really open-ended.

You might not know what next steps for smoking out all other potential bugs would look like. But if you have even kind of a place you would suggest starting, would you mind leaving that as a breadcrumb for other developers? Right now this bug is written in a way where it seems hard to start work on, and anything that makes it easier to work will help it actually get worked.

Grunet:
[...]

PR #474, opened 2021-11-16
Finish Off Zipkin Spec Updates

### LINKED ISSUE #295: Update Zipkin Exporter to conform to latest spec changes

Grunet:
Resolves #295 and maybe addresses part of #283 ?

This just covers these 2 parts of [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/sdk_exporters/zipkin.md#opentelemetry-to-zipkin-transformation)

[Remote Endpoint](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/sdk_exporters/zipkin.md#remote-endpoint)
[Event Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/sdk_exporters/zipkin.md#events)

I co-opted Go's code heavily for both sets of changes, as well as the comments in and around them for context (some of which didn't line up with Go's implementation actually)

[Remote Endpoint](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L238)
[Event Attributes](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L141)

I haven't done any testing with an actual Zipkin instance (so idk how much of this actually works), but I tried to add unit tests as appropriate. These are the 4 untested lines I noticed in SpanConverter.php

- Line 177 - if json_encode fails on the attributes array
- Line 192 - the preferred attribute ends up not having a string value
- Line 232 - the preferred attribute ends up not having a string value
- Line 236 - if an invalid ip is passed (edited) 

Other notes on things I grappled with along the way:

- Go checks if attribute values are a string in some places, but not others (see [this comment](https://github.com/open-telemetry/opentelemetry-go/blob/4ba964bae77d9d32b4bb0167ed5533a0c05dcdf9/attribute/value.go#L200) that suggests it should always do so)
- Setting 'serviceName' to 'unknown' isn't done in the Go code nor mentioned in the spec, but it is called out in a comment [here](https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L125)
- I couldn't really figure out what the keys of the 'remoteEndpoint' node were supposed to be, nor what their values' units should be, but I loosely inferred them based off of the comments in Go [here](https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L127)
- Not done in code, but a [Go comment](https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L122) mentions port = 0 should be the default
- I couldn't figure out a clean way in PHP to limit a port to less than 65536, like Go seems to do [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L292)
- defined('AF_INET6') seems to be returning something falsy while running the tests, even while inet_pton seems to be working ok, so I wasn't able to use it as a detector for if PHP was compiled with ipv6 support (I'm not really sure what inet_pton will do in that case tbh. Tried googling but couldn't suss it out)
- Other parts of the file use a ```foreach($attributes as $k => $v)``` approach, but I couldn't figure out how that worked in general, with AttributesInterface only extending IteratorAggregate and Countable, and then AttributeInterface only declaring getKey and getValue methods

### DIFF
--- a/src/Contrib/Zipkin/SpanConverter.php
+++ b/src/Contrib/Zipkin/SpanConverter.php
@@ -184,53 +184,12 @@ class SpanConverter
         //ip address handling from the spec, like is done here - https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L264-L271
         switch($preferredAttr->getKey()) {
             case "net.peer.ip":
-                $ipString = $preferredAttr->getValue();
-
-                if (filter_var($ipString, FILTER_VALIDATE_IP)) {
-
-                    $remoteEndpointArr = [
-                        //Not in the Go code but mentioned in a comment here - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L125
-                        'serviceName' => "unknown"
-                    ];
-
-                    if (filter_var($ipString, FILTER_VALIDATE_IP,FILTER_FLAG_IPV4)) {
-                        //Key casing/value units loosely inferred from Go comments around here - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L127
-                        $remoteEndpointArr["ipv4"] = ip2long($ipString);
-                    }
-
-                    if (filter_var($ipString, FILTER_VALIDATE_IP,FILTER_FLAG_IPV6)) {
-                        //if (defined('AF_INET6')) { //TODO - figure out why this was never getting dropped into during the tests
-                            //This won't work (oddly) if ipv6 has been disabled for PHP and the server it's running on
-                            //Where this idea came from - https://www.php.net/manual/en/function.inet-pton.php#104917
-                            $remoteEndpointArr["ipv6"] = inet_pton($ipString);
-                        //}
-                        //TODO - does the else case need handling here?
-                    }
-
-                    //Go comment (but not code...), mentions port = 0 should be the default - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L122
-                    $remoteEndpointArr["port"] = 0;
-                    foreach ($span->getAttributes() as $attr) {
-                        if ($attr->getKey() === "net.peer.port") {
-                            //TODO - take care of the cases where this isn't a string
-                            $portVal = $attr->getValue();
-                            $portInt = intval($portVal); //TODO - find out if Go's setting of 16 for bit size is implicitly the case here - https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L292
-
-                            $remoteEndpointArr["port"] = $portInt;
-
-                            break;
-                        }
-                    }
-
-                    return $remoteEndpointArr;
-                }
-                break;
+                return SpanConverter::getRemoteEndpointDataFromIpAddressAndPort($preferredAttr, $span);
             default:
                 return [
                     'serviceName' => $preferredAttr->getValue(),
                 ];
         }
-
-        return null; //Determine if this is safe to remove
     }
 
     private static function findRemoteEndpointPreferredAttribute(SpanDataInterface $span): ?Attribute {
@@ -249,5 +208,49 @@ class SpanConverter
         }
 
         return $preferredAttr;
+    }
+
+    private static function getRemoteEndpointDataFromIpAddressAndPort(Attribute $preferredAttr, SpanDataInterface $span): ?array { //TODO - switch the span out for just the list of attributes
+        $ipString = $preferredAttr->getValue();
+
+        if (filter_var($ipString, FILTER_VALIDATE_IP)) {
+
+            $remoteEndpointArr = [
+                //Not in the Go code but mentioned in a comment here - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L125
+                'serviceName' => "unknown"
+            ];
+
+            if (filter_var($ipString, FILTER_VALIDATE_IP,FILTER_FLAG_IPV4)) {
+                //Key casing/value units loosely inferred from Go comments around here - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L127
+                $remoteEndpointArr["ipv4"] = ip2long($ipString);
+            }
+
+            if (filter_var($ipString, FILTER_VALIDATE_IP,FILTER_FLAG_IPV6)) {
+                //if (defined('AF_INET6')) { //TODO - figure out why this was never getting dropped into during the tests
+                    //This won't work (oddly) if ipv6 has been disabled for PHP and the server it's running on
+                    //Where this idea came from - https://www.php.net/manual/en/function.inet-pton.php#104917
+                    $remoteEndpointArr["ipv6"] = inet_pton($ipString);
+                //}
+                //TODO - does the else case need handling here?
+            }
+
+            //Go comment (but not code...), mentions port = 0 should be the default - https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L122
+            $remoteEndpointArr["port"] = 0;
+            foreach ($span->getAttributes() as $attr) {
+                if ($attr->getKey() === "net.peer.port") {
+                    //TODO - take care of the cases where this isn't a string
+                    $portVal = $attr->getValue();
+                    $portInt = intval($portVal); //TODO - find out if Go's setting of 16 for bit size is implicitly the case here - https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L292
+
+                    $remoteEndpointArr["port"] = $portInt;
+
+                    break;
+                }
+            }
+
+            return $remoteEndpointArr;
+        }
+
+        return null;
     }
 }
--- a/src/Contrib/Zipkin/SpanConverter.php
+++ b/src/Contrib/Zipkin/SpanConverter.php
@@ -147,7 +147,7 @@ class SpanConverter
     }
 
     private static function toAnnotation(EventInterface $event): array {
-        $value = $event->getName();
+        
 
         if (count($event->getAttributes()) > 0) {
             $attributesArray = [];
@@ -161,6 +161,9 @@ class SpanConverter
 
                 $value = "\"{$eventName}\": {$attributesAsJson}";
             }
+        } else {
+            $eventName = $event->getName();
+            $value = "\"{$eventName}\"";
         }
 
         $annotation = [
--- a/src/Contrib/Zipkin/SpanConverter.php
+++ b/src/Contrib/Zipkin/SpanConverter.php
@@ -227,12 +227,7 @@ class SpanConverter
         }
 
         if (filter_var($ipString, FILTER_VALIDATE_IP,FILTER_FLAG_IPV6)) {
-            //if (defined('AF_INET6')) { //TODO - figure out why this was never getting dropped into during the tests
-                //This won't work (oddly) if ipv6 has been disabled for PHP and the server it's running on
-                //Where this idea came from - https://www.php.net/manual/en/function.inet-pton.php#104917
-                $remoteEndpointArr["ipv6"] = inet_pton($ipString);
-            //}
-            //TODO - does the else case need handling here?
+            $remoteEndpointArr["ipv6"] = inet_pton($ipString);
         }
 
         $remoteEndpointArr["port"] = $portNumber ?? 0;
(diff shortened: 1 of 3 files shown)


Grunet:
Sorry for the super messy commit history. Lmk if I should squash it down before merging (couldn't quite see if Gitpod's editor had a tool for that, or if I need to suss out the CLI)

tidal:
Thank you!


> * Line 177 - if json_encode fails on the attributes array

json_encode should use the JSON_THROW_ON_ERROR flag to make it throw exceptions on an error. 
You can provoke and error by passing a file pointer (preferrable to smth like std out) as a value.


> * Line 236 - if an invalid ip is passed (edited)

Why should the exporter/converter care, what a valid IP is? And if it's an incorrect one, is it silently dropped? That would make it harder to pinpoint errors than just passing through an incorrect IP imo.



> * Setting 'serviceName' to 'unknown' isn't done in the Go code nor mentioned in the spec, but it is called out in a comment [here](https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L125)

The [service.name](https://github.com/open-telemetry/opentelemetry-specification/blob/48ea76ccf06a027278e885e774b1371b91d6454d/specification/resource/semantic_conventions/README.md#service) attribute is mandatory, and it's the `$name` parameter in the exporter's constructor, so it there should be alsways a value for it. However the converter should probably not know about this behavior.



> * I couldn't figure out a clean way in PHP to limit a port to less than 65536, like Go seems to do [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L292)

There is no simple way, as (vanilla) PHP only has integers, but is this important in this case?




> * Other parts of the file use a `foreach($attributes as $k => $v)` approach, but I couldn't figure out how that worked in general, with AttributesInterface only extending IteratorAggregate and Countable, and then AttributeInterface only declaring getKey and getValue methods

If you a referring to this,
```php
        foreach ($span->getAttributes() as $k => $v) {
            $row['tags'][$k] = $this->sanitiseTagValue($v->getValue());
        }
```
The key is set on `Attributes::setAttribute`.
However I'm not really sure, why the Iterator construct is needed in the first place.

tidal:
> couldn't quite see if Gitpod's editor had a tool for that

Nothing against Gitpod, it's awesome, but sadly(!) no other IDE/Editor comes even close to PHPStorm in case of working with PHP. (I know this is not a PHP 'problem')

src/Contrib/Zipkin/SpanConverter.php
```diff
@@ -131,4 +145,127 @@ private static function toSpanKind(SpanDataInterface $span): ?int
         
         return null;
     }
+
+    private static function toAnnotation(EventInterface $event): array
+    {
+        $eventName = $event->getName();
+        $attributesAsJson = SpanConverter::convertEventAttributesToJson($event);
```
tidal, line 152: If you would check here for ``count($event->getAttributes()) === 0``, you could simplify it a bit.

src/Contrib/Zipkin/SpanConverter.php
```diff
@@ -131,4 +145,127 @@ private static function toSpanKind(SpanDataInterface $span): ?int
...
+
+        $value = ($attributesAsJson !== null) ? "\"{$eventName}\": {$attributesAsJson}" : "\"{$eventName}\"";
```
tidal, line 154: You could use ``sprintf('"{%s}"', $eventName)`` to avoid escaping the quotes.

src/Contrib/Zipkin/SpanConverter.php
```diff
@@ -98,13 +111,14 @@ public function convert(SpanDataInterface $span)
...
 
         foreach ($span->getEvents() as $event) {
-            if (!array_key_exists('annotations', $row)) {
-                $row['annotations'] = [];
+            $row['annotations'][] = SpanConverter::toAnnotation($event);
+        }
+
+        if (($span->getKind() === SpanKind::KIND_CLIENT) || ($span->getKind() === SpanKind::KIND_PRODUCER)) {
+            $remoteEndpointData = SpanConverter::toRemoteEndpoint($span);
+            if ($remoteEndpointData != null) {
```
tidal, line 119: Did you use a weak comparison on purpose?

src/Contrib/Zipkin/SpanConverter.php
```diff
@@ -131,4 +145,127 @@ private static function toSpanKind(SpanDataInterface $span): ?int
...
+        if ($preferredAttr === null) {
+            return null;
+        }
+
+        if (!is_string($preferredAttr->getValue())) {
+            return null;
+        }
+
+        switch ($preferredAttr->getKey()) {
+            case 'net.peer.ip':
```
tidal, line 196: Could you create/use a constant for this, please?

tests/Contrib/Unit/ZipkinSpanConverterTest.php
```diff
@@ -176,6 +181,54 @@ public function shouldConvertAnOTELSpanOfUnknownKindToAZipkinSpanOfUnspecifiedKi
         $this->assertArrayNotHasKey('kind', $row);
     }
 
+    /**
+     * @test
+     */
+    public function shouldConvertAnEventWithoutAttributesToAnAnnotationWithOnlyItsName()
+    {
+        $span = (new SpanData())
+            ->addEvent('event.name', new Attributes());
```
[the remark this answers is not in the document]
Grunet: Updated, though not sure I fully understand what this is for tbth (closest I can imagine is to make tests more unit-y so failures are easier to pinpoint? Or something similar? Would be curious to understand this a bit better). Also not entirely sure what an easy way is to know if it worked

src/Contrib/Zipkin/SpanConverter.php
Grunet: Yup, done

src/Contrib/Zipkin/SpanConverter.php
Grunet: Nope, nice catch!

src/Contrib/Zipkin/SpanConverter.php
Grunet: Ah nice, that looks much better!

src/Contrib/Zipkin/SpanConverter.php
Grunet: I feel like I should get what you mean by this, but am struggling to at the moment. Is your idea to move the check on line 168 into this method, maybe using it in the ternary instead? 

I'm not sure how to then handle the json_encode failure case, but maybe that's a separate thing

Grunet:
> > * Line 177 - if json_encode fails on the attributes array
> 
> json_encode should use the JSON_THROW_ON_ERROR flag to make it throw exceptions on an error. You can provoke and error by passing a file pointer (preferrable to smth like std out) as a value.
> 
> > * Line 236 - if an invalid ip is passed (edited)
> 
> Why should the exporter/converter care, what a valid IP is? And if it's an incorrect one, is it silently dropped? That would make it harder to pinpoint errors than just passing through an incorrect IP imo.
> 
> 
> > * I couldn't figure out a clean way in PHP to limit a port to less than 65536, like Go seems to do [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/zipkin/model.go#L292)
> 
> There is no simple way, as (vanilla) PHP only has integers, but is this important in this case?

Yeah for these 3 points that all have to do with input validation (in some form or another) I'm not sure exactly what the right approach is, either at a library level or even OTEL-level (I vaguely recall seeing part of a spec around this maybe?). 

I assume giving some sort of feedback to the consumer would be ideal, but I'm not too sure in what form they should be

Grunet:
> > * Setting 'serviceName' to 'unknown' isn't done in the Go code nor mentioned in the spec, but it is called out in a comment [here](https://github.com/open-telemetry/opentelemetry-go/blob/7ce58f355851d0412e45ceb79d977bc612701b3f/exporters/jaeger/internal/gen-go/zipkincore/zipkincore.go#L125)
> 
> The [service.name](https://github.com/open-telemetry/opentelemetry-specification/blob/48ea76ccf06a027278e885e774b1371b91d6454d/specification/resource/semantic_conventions/README.md#service) attribute is mandatory, and it's the `$name` parameter in the exporter's constructor, so it there should be alsways a value for it. However the converter should probably not know about this behavior.

I did some google searching and found [this Zipkin doc](https://zipkin.io/zipkin/2.11.0/zipkin/zipkin2/Span.html#localEndpoint--) that tries to define what "localEndpoint" and "remoteEndpoint" mean for it.

Not sure I fully understand them, but given that the $this->serviceName member variable is already being used for the "localEndpoint", based on those definitions it seems like it might be odd to try and re-use it for the "remoteEndpoint" perhaps.

Grunet:
> > * Other parts of the file use a `foreach($attributes as $k => $v)` approach, but I couldn't figure out how that worked in general, with AttributesInterface only extending IteratorAggregate and Countable, and then AttributeInterface only declaring getKey and getValue methods
> 
> If you a referring to this,
> 
> ```
>         foreach ($span->getAttributes() as $k => $v) {
>             $row['tags'][$k] = $this->sanitiseTagValue($v->getValue());
>         }
> ```
> 
> The key is set on `Attributes::setAttribute`. However I'm not really sure, why the Iterator construct is needed in the first place.

Yup that's the one. And ahhh ok I think I see how the Attributes class does that magic a little better now (ty!) though I'm still a little perplexed. Would any implementation of AttributesInterface have to satisfy this "associative array"-style requirement?

(Also FWIW the iterator point seems related to #415 at a high-level)

src/Contrib/Zipkin/SpanConverter.php
tidal: Yeah, something like that, but it's honestly not that important. :)

tests/Contrib/Unit/ZipkinSpanConverterTest.php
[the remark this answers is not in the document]
tidal: Code coverage normally records all code which has been **executed** during a test, whether it has been explicitly tested or not. So when dependecies are not mocked, it will create false postives in code coverage. In other words in this case code coverage is not a good metric anymore to see what parts of the code has been covered by unit tests. (It's not a good metric to show _what_ has been tested anyway, only what has _not_ been tested.)

In order to avoid these false positives in PHPUnit there are basically tow ways. Either you mock away _all_ dependencies of the tested library, or you tell PHPUnit with the '@covers' annotation, what part of the code you actually trying to test.

You can execute phpunit like this to create a HTML view of the code coverage: 
```bash
vendor/bin/phpunit --coverage-html=tests/coverage/html
```
You can find the HTML code coverage report in `tests/coverage/html`

(I have created a PR (https://github.com/open-telemetry/opentelemetry-php/pull/477) to add a make target to create the report)

tidal:
> Would any implementation of AttributesInterface have to satisfy this "associative array"-style requirement? (maybe that's what IteratorAggregate enforces? Can't quite tell from the doc)
 
`IteratorAggregate` itself simply exists to avoid `getIterator`calls in loops, that's all. 
(I personally try to avoid Iterators, when possible, because the code tends to be easier to read without them. However they can of course make code more type safe.)

tests/Contrib/Unit/ZipkinSpanConverterTest.php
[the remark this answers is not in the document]
tidal: Maybe one more thing regarding mocks vs. covers annotation. 

A lot of people do not like writing mocks, which is understandable since they can get very tricky. Especially when you deal with a lot of nested dependencies and have to replicate a lot of the behavior of the mocked classes, it can get complicated.

However , I prefer writing mocks for one reason. If I find myself writing a lot of complicated mocks, it can be an indicator for a subpar design choice I did (for example "violating" [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) or [LoD](https://en.wikipedia.org/wiki/Law_of_Demeter) ) . So the solution is refactoring the code to be less complicated, and the mocks will be less complicated as well. 
(Needless to say, I only know this, because I made a lot of bad design choices in the past)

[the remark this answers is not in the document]
Grunet: Yup I think that makes sense to me, making the code coverage data better highlight which units aren't being directly, intentionally tested vs having them only being incidentally indirectly covered via other tests (is how I'm conceptualizing it now at least).

And yeah I feel that too, when mocking gets really complicated I've felt before the real problem is often elsewhere and I'm just working around it.

tidal (approved):

bobstrecansky:
Is this ready for review / merge?

Grunet:
> Is this ready for review / merge?

@bobstrecansky Yes, the main outstanding question for me is how to handle invalid input from the input spans (e.g. invalid ip addresses, port numbers) and possibly give feedback to the user (when in a dev environment).

But after chatting with @brettmc in the SIG meeting it sounds like this might be out of scope to figure out for this PR, in which case it should be fully ready to review

bobstrecansky:
Cool I'll review tomorrow.  Thanks!

bobstrecansky (approved):

Issue #452, opened 2021-10-21
Http based exporters require a port and a path in enpoint url


tidal:
**Steps to reproduce**

Example for Zipikin exporter:

```
$exporter = new OpenTelemetry\Contrib\Zipkin\Exporter(
    'foo',
    'http://zipkin',
    $client,
    $requestFactory,
    $streamFactory
);
```

**What is the expected behavior?**

An exporter instance to be created.

**What is the actual behavior?**

`InvalidArgumentException: Endpoint should have scheme, host, port and path`

**Additional context**

Collector endpoints should be accessible through reverse proxies, load balancers, etc., which may hide the endpoint port and/or path

tidal:
related: https://github.com/open-telemetry/opentelemetry-php/issues/451 , https://github.com/open-telemetry/opentelemetry-php/issues/448

PR #476, opened 2021-11-17
Exporter Refactoring

tidal:
- Closes: https://github.com/open-telemetry/opentelemetry-php/issues/452
- Fixes 10) from https://github.com/open-telemetry/opentelemetry-php/issues/451
- Adds composer package dg/bypass-finals to allow mocking of final classes/methods

This is mainly a refactoring and abstraction of exporters and respective tests to make the code easier maintainable (removing most of the duplicate code) and a setup for comimg changes to the exporter factory etc. Besides allowing URLs without ports and/or paths, this PR doesn't change any behavior.

An example, how the ZipkinExporterTest looks like now:

```php
<?php

declare(strict_types=1);

namespace OpenTelemetry\Tests\Contrib\Unit;

use OpenTelemetry\Contrib\Zipkin\Exporter;

class ZipkinExporterTest extends AbstractHttpExporterTest
{
    protected const EXPORTER_NAME = 'test.zipkin';

    /**
     * @psalm-suppress PossiblyInvalidArgument
     */
    public function createExporterWithDsn(string $dsn): Exporter
    {
        return new Exporter(
            self::EXPORTER_NAME,
            $dsn,
            $this->getClientInterfaceMock(),
            $this->getRequestFactoryInterfaceMock(),
            $this->getStreamFactoryInterfaceMock()
        );
    }
}
```


tidal:
The tests seem to fail, because the `assert` functions are not executed in CI.
I will quick fix this, and we can later investigate the reason.

tidal:
The tests failed because the `assert` functions have not been excecuted in the CI run. I have temporarily fixed this by always throwing exceptions, so we can later find out how to make CI execute  `assert` (Its probably a production ini setting).

tidal:
Please do not merge, yet. I want to add (or actually remove) some more things.

tidal:
It's ok for me now. 
(However I had to resolve the merge conflits twice, because github was pulling in changes from other PRs into my dev branch. That's why there are all these duplicate commits).
The Psalm Code Scanning job seems to be stuck though.

bobstrecansky (approved):

Issue #382, opened 2021-07-19
Full otel example


hrodic:
**Is your feature request related to a problem?**
What is the recommended way to propagate trace between 2 distributed applications?

**Describe the solution you'd like**
I would like to have a full example (if possible docker-composed)  in examples/ folder that does something like this
1. service 1 trace starting and generation of spans
2. service 1 injects trace context to the headers and sends request to service 2
3. service 2 receives headers and extracts them to trace continuing it and adding more spans.
4. service 2 finishes
5. service 1 finishes
6. backend could be a otel-collector with debug active so you can see the trace (spans) coming in.

**Describe alternatives you've considered**
Using Guzzle, for example.

**Additional context**
This will help a lot of people testing this library and giving more feedback.

bobstrecansky:
@ohamuy is planning on working on a full fledged AWS example, I'm going to tag him on this.

ohamuy:
Hello @hrodic! I have implemented two example applications in the contributor repo. There is a README for all the necessary installations. These examples should work with other backends as well if you would want to use something alternative to AWS X-Ray

ohamuy:
I hope it is sufficient for your request! In the future there should be more robust examples with actual web applications coming out

brettmc:
@hrodic - there is an example now that covers your request. Please check it out, feedback welcome. Can this issue be closed now?

brettmc:
Closing due to no feedback, I think the issue has been fixed now.

PR #483, opened 2021-11-21
distributed tracing example

brettmc:
Adding a self-contained example for distributed tracing into both jaeger and zipkin.


examples/distributed-tracing/src/index.php
```diff
@@ -0,0 +1,123 @@
...
+        //a guzzle middleware to wrap http calls in a span, and inject trace headers
+        $stack->push(function(callable $handler) {
+            return function (RequestInterface $request, array $options) use ($handler): PromiseInterface {
+                $span = TracerProvider::getDefaultTracer()
+                    ->spanBuilder(sprintf('%s %s', $request->getMethod(), $request->getUri()))
+                    ->setSpanKind(SpanKind::KIND_CLIENT)
+                    ->setAttribute('http.method', $request->getMethod())
+                    ->setAttribute('http.url', $request->getUri())
+                    ->startSpan();
+                $ctx = $span->storeInContext(Context::getCurrent());
```
brettmc, line 42: This works, but it looks like a few revisions ago you could just pass a SpanContext for trace propagation, which would be simpler since a SpanContext is what it ultimately works with...

tidal:
You beat me to it. :) 
However, I guess two examples can't hurt, and mine will be very different (no framework and using envoy to simulate a service mesh)

Btw: you can close https://github.com/open-telemetry/opentelemetry-php/issues/382 with this.

examples/distributed-tracing/README.md
```diff
@@ -0,0 +1,25 @@
+# Distributed Tracing Example
+This example uses `docker-compose`, and illustrates the distributed tracing functionality of OpenTelemetry. An HTTP request to service-one will make multiple con current HTTP requests, each of which is injected with a `traceparent` header.
```
tidal, line 2: "con current" -  is this a pun, I don't get?

examples/distributed-tracing/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+All trace data is exported via grpc to an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/), where they are forwarded to zipkin and jaeger.
+
+The example is presented as a [slim framework](https://www.slimframework.com/) single-page application for simplicity, and uses Guzzle as an HTTP client. The same application source is used for all services.
```
tidal, line 6: "single file application"?

examples/distributed-tracing/src/composer.json
```diff
@@ -0,0 +1,15 @@
...
+  "require": {
+    "php": ">=8",
+    "open-telemetry/opentelemetry": "dev-main",
+    "guzzlehttp/guzzle": "*",
+    "php-http/guzzle7-adapter": "*",
+    "slim/slim": "~4",
+    "php-di/php-di": "^6.3",
+    "php-di/slim-bridge": "^3.2"
+  }
+}
```
tidal, line 15: Could you please commit your composer.lock, to make sure this example always works?

examples/distributed-tracing/README.md
brettmc: nope, just a misplaced space. I changed it to "asynchronous"

examples/distributed-tracing/src/composer.json
brettmc: I can. I'd better also change "composer update" to "composer install" if we want it to always work...

brettmc: and, to be able to commit a lock file I'm waiting on #484 - dev-main in packagist does not have some recent changes that this example uses.

tidal: I see. 
I never noticed that composer is using packagist to resolve the latest commit for branches.

brettmc: I can change the composer config so that it pulls the library from github rather than packagist? The only reason I didn't do it that way was because I saw that packagist is the official distribution method (and avoids potentially having to deal with github tokens to install)

.github/workflows/psalm-security-analysis.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         uses: docker://vimeo/psalm-github-actions
         with:
           composer_require_dev: true
-          composer_ignore_platform_reqs: false # https://github.com/psalm/psalm-github-actions/pull/29
+          composer_ignore_platform_reqs: true
```
brettmc, line 17: Note that this got merged a couple of days ago, so undoing the existing workaround.

examples/distributed-tracing/src/composer.json
tidal: Yes, I guess that's a good temporary workaround until auto-update is enabled and/or packagist catches up.

bobstrecansky (approved):

PR #487, opened 2021-11-24
Optimize base PHP Dockerfile

tidal:
This PR reduces the size of the base php image from 627mb to 129mb, and the containers are now run by non privileged user (cotainers were writing as root on local file-system. Some security focused K8S distributions like OpenShift/OKD don't allow privileged users to run containers). Also the PHP version is now 7.4 (instead of 8.0) to ensure compatibility with minimum supported PHP version.

- Uses Alpine (instead of Debain) as base image with smaller footprint
- Uses docker-php-extension-installer
- Strips debug symboly from grpc and xdebug extensions
- Uses a single RUN command to reduce layer size
- ~~Runs containers as alpines standard user "guest"~~

Changes to fix issues mentioned by @brettmc :

The image creates a system user ("php"), which is used to run the container per default. However "root" is still the default
user in docker-compose, but the user can be set to "php/guest" by env var "PHP_USER". I added a `.env.dist` file, which can be copied to `.env` (which is ignored by git) to provide the user for docker-compose. This should solve the issues, the `composer.lock` have been written by root, and the container can still be run as a non privileged user.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,10 @@
-FROM php:7.4-cli-alpine
+
+FROM php:7.4-cli-alpine as php_build
 
 ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
 
 RUN chmod +x /usr/local/bin/install-php-extensions; \
-    apk add --no-cache bash git binutils; \
+    apk add --update binutils; \
     install-php-extensions \
         ast-1.0.14 \
         xdebug \
@@ -14,10 +15,20 @@ RUN chmod +x /usr/local/bin/install-php-extensions; \
     ; \
     # strip debug symbols from extensions to reduce size
     strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/grpc.so; \
-    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so; \
-    rm /usr/local/bin/install-php-extensions; \
-    apk del --no-cache binutils
+    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so;
 
-USER guest
+FROM php_build
 
 WORKDIR /usr/src/myapp
+
+RUN   apk add --no-cache bash git; \
+      chmod +x -R /usr/local/lib/php/extensions/no-debug-non-zts-20190902; \
+      addgroup -g "1000" -S php; \
+      adduser --system \
+        --gecos "" \
+        --ingroup "php" \
+        --uid "1000" \
+        "php";
+
+USER php
+
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,17 +1,23 @@
-FROM php:8-buster
+FROM php:7.4-cli-alpine
 
-RUN apt-get -y update && apt-get -y install git zip && \
-    curl -sS https://getcomposer.org/installer | php && \
-    mv composer.phar /usr/local/bin/composer && \
-    chmod +x /usr/local/bin/composer && \
-    pecl install ast-1.0.14 xdebug && \
-    docker-php-ext-enable ast xdebug && \
-    # The pcntl extension is used for speeding up `make phan`
-    docker-php-ext-install pcntl
+ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
 
-# install GRPC
-RUN apt install -y --no-install-recommends zlib1g-dev && \
-    pecl install grpc && \
-    docker-php-ext-enable grpc
+RUN chmod +x /usr/local/bin/install-php-extensions; \
+    apk add --no-cache bash git binutils; \
+    install-php-extensions \
+        ast-1.0.14 \
+        xdebug \
+        zip \
+        pcntl \
+        grpc \
+        @composer \
+    ; \
+    # strip debug symbols from extensions to reduce size
+    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/grpc.so; \
+    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so; \
+    rm /usr/local/bin/install-php-extensions; \
+    apk del --no-cache binutils
+
+USER guest
 
 WORKDIR /usr/src/myapp
--- a/.env.dist
+++ b/.env.dist
@@ -1,0 +1,1 @@
+PHP_USER=php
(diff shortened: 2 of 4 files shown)


brettmc:
Looks good to me - there is a `make bash` in the Makefile which will need to be updated (or, could install a bash interpreter in alpine?)

docker/Dockerfile
```diff
@@ -1,17 +1,23 @@
...
+ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
 
-# install GRPC
-RUN apt install -y --no-install-recommends zlib1g-dev && \
-    pecl install grpc && \
-    docker-php-ext-enable grpc
+RUN chmod +x /usr/local/bin/install-php-extensions; \
+    apk add --no-cache bash git binutils; \
+    install-php-extensions \
+        ast-1.0.14 \
```
tidal, line 8: This version of AST was used in the Dockerfile before. I didn't know if this is needed for something, so I left it how it was.

docker/Dockerfile
```diff
@@ -1,17 +1,23 @@
...
-    # The pcntl extension is used for speeding up `make phan`
-    docker-php-ext-install pcntl
+ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
 
-# install GRPC
-RUN apt install -y --no-install-recommends zlib1g-dev && \
-    pecl install grpc && \
-    docker-php-ext-enable grpc
+RUN chmod +x /usr/local/bin/install-php-extensions; \
+    apk add --no-cache bash git binutils; \
```
tidal, line 6: I only added git, because it comes with the Debian image and did not know if someone may use it. It's probably not needed, but not that much of an overhead.

brettmc:
I have a couple of minor issues with bash and permissions using this image:
```
$ make bash
docker-compose run --rm php bash
bash: /dev/null/.bashrc: Not a directory
bash-5.1$
```
I haven't seen the "guest" user before, but I think it might need uid:gid 1000:1000 (at least, for Linux, which is the default). As it stands, I can't "composer install" from within a container because I don't have write permissions

brettmc:
This feels a little hacky, but resolves permissions and bashrc complaints for me:
```
RUN deluser guest; \
    addgroup --gid 1000 guest; \
    adduser --system --ingroup guest --uid 1000 guest; \
    mkdir /usr/src/myapp; \
    chown guest:guest /usr/src/myapp
```
It does complain when deleting "guest" but succeeds:
```
Step 4/6 : RUN deluser guest;     addgroup --gid 1000 guest;     adduser --system --ingroup guest --uid 1000 guest;     mkdir /usr/src/myapp;     chown guest:guest /usr/src/myapp
 ---> Running in 417b0b32894d
deluser: can't find guest in /etc/group
Removing intermediate container 417b0b32894d
 ---> 5e5d63dd42f8
```

tidal:
> I have a couple of minor issues with bash and permissions using this image:
> 
> ```
> $ make bash
> docker-compose run --rm php bash
> bash: /dev/null/.bashrc: Not a directory
> bash-5.1$ touch composer.lock
> touch: composer.lock: Permission denied
> ```
> 
> I haven't seen the "guest" user before, but I think it might need uid:gid 1000:1000 (at least, for Linux, which is the default). As it stands, I can't "composer install" from within a container because I don't have write permissions

If that's how you used to call composer before as well, the issue is that the composer.lock file has been written by root, same as all the cache files and so on. 

I guess the "guest" user is inherited from busybox.

I haven't had bash complain about .bashrc , especially not at  `/dev/null/.bashrc` (but it's not really surprising that it's not a directory) . However I tested with `nerdctl run -it opentelemetry-php_php bash`, but that shoudln't make a difference.

Adding new users with IDs starting at 1000 is more a convention of some distributions. RHEL (and others) for example start at 500. I don't know about alpine, since it has no `/etc/login.defs`, and I don't know, where the information would be located otherwise.

tidal:
> This feels a little hacky, but resolves permissions and bashrc complaints for me:
> 
> ```
> RUN deluser guest; \
>     addgroup --gid 1000 guest; \
>     adduser --system --ingroup guest --uid 1000 guest; \
>     mkdir /usr/src/myapp; \
>     chown guest:guest /usr/src/myapp
> ```
> 
> It does complain when deleting "guest" but succeeds:
> 
> ```
> Step 4/6 : RUN deluser guest;     addgroup --gid 1000 guest;     adduser --system --ingroup guest --uid 1000 guest;     mkdir /usr/src/myapp;     chown guest:guest /usr/src/myapp
>  ---> Running in 417b0b32894d
> deluser: can't find guest in /etc/group
> Removing intermediate container 417b0b32894d
>  ---> 5e5d63dd42f8
> ```

Maybe, it's easier to just add a new user, instead of deleting "guest". I will have a look later and see, if I can reproduce the `.bashrc`issue.

However `chown` should not make a difference here, since the file system is not mounted at container bulld time.

tidal:
@brettmc 
Could you please check, if this update works for you? 
(I have added more information to the PR description)

brettmc:
@tidal - it works perfectly for me now. It maybe needs a sentence in the development/docker section of readme on how to set the user, but nothing else from me.

docker/Dockerfile
bobstrecansky: I believe it was necessary for the Phan installation of I remember correctly

bobstrecansky:
This deviates from the tooling we use in CI:

https://github.com/open-telemetry/opentelemetry-php/blob/main/.github/workflows/php.yml

That being said, I'm okay with this change if we think it will help with developer friction.

tidal:
> This deviates from the tooling we use in CI:
> 
> https://github.com/open-telemetry/opentelemetry-php/blob/main/.github/workflows/php.yml
> 
> That being said, I'm okay with this change if we think it will help with developer friction.
@bobstrecansky 
I'm not quite sure, if I understand what you mean by tooling here. (given English is not my native language). 
Could you elaborate on this, please? Do you mean the OS, php version or something else?

I personally can work around any of the issues I mentioned, so it's more about sharing, in case someone else runs into the same issues. However, there is no need to replace the current docker file/image. We could also have/provide two docker files and/or images and make the image configurable in the docker file via env var. What about that?
The alpine image doesn't need to be build via an github action either, though just configuring an image name and or tag? 

Maybe we just talk freeze this PR fow now and talk about it in more detail in next weeks meeting?  I guess that's a bit easier than to discuss all the options and  pros/cons here.

bobstrecansky:
Yeah that sounds good to me!  Happy to talk about it on Wednesday.  I very much like the way this PR is moving, just want to make sure we are all on the same page.

bobstrecansky:
@tidal wanted to remove additional debug symbols here and then we will merge.

tidal:
@bobstrecansky 
For some reasons (which I don't want to investigate right now) my attempts ended up in a bigger image.
It's 129mb how it is right now. I think that's a reasonable size, so let's just go ahead and merge it how it is atm.

bobstrecansky (approved):

Issue #383, opened 2021-07-21
Merge HTTP and GRPC OTLP Exporters into a single interface


SeanHood:
**Is your feature request related to a problem?**
As a telemetry implementer, I'd like to easily switch between OTLP/HTTP and OTLP/GRPC

**Describe the solution you'd like**
There should be a single OTLP Exporter interface, which based on a configuration option based on `OTEL_EXPORTER_OTLP_PROTOCOL` should accept `http/protobuf` or `grpc` and use the correct Exporter accordingly.

What this interface should look like, I'm not sure or if it's feasible. Currently the two Exporters take different sets of parameters.

**Describe alternatives you've considered**
Having the two exporters separately I believe makes it harder to switch as this requires a code change rather than a config change.

**Additional context**
This is a draft issue to be filled out with examples from around the Otel community on how it's been done in other languages.

bobstrecansky:
@SeanHood - are you planning on working on this ticket or are you hoping someone else will work on it?

SeanHood:
Happy to work on it, had a few ideas over the weekend.

I'm thinking a new class e.g. `OpenTelemetry\Contrib\Otlp\ConfigOpts()` which could then be passed to either the OtlpGrpc or OtlpHttp Exporters.

End users would see some interfaces like so:

```php
$exporterOpts = new ConfigOpts([
    'endpoint'    => 'api.example.com:4317',
    'compression' => 'gzip'
]);

$exp = OTLPExporter(..., $exporterOpts);
```

or 

```php
$exporterOpts = new ConfigOpts();
$exporterOpts->endpoint = 'api.example.com';
$exporterOpts->compression = 'gzip';

$exp = OTLPExporter(..., $exporterOpts);
```

`ConfigOpts()` would also take care of figuring out all of the Env var configuration and precedence.

This idea doesn't quite go far enough to normalise the HTTP or GRPC Exporter signatures to make them easily interchangeable but it wouldn't be far off.

SeanHood:
Just had a thought: I was going to do this as part of this issue but maybe could be split into a new issue(?). Would be to merge `OtlpHttp/SpanConverter` and `OtlpGrpc/SpanConverter` into `Otlp/SpanConverter`. From when I worked on them I think they're 99-100% the same but this would save some duplication of effort between updates to either.

SeanHood:
[...]

PR #488, opened 2021-11-26
Otlp centralised config

brettmc:
This is stage 1 of implementing #383 - taking SeanHood's initial work and expanding on it to centralise the configuration of http and grpc flavours of otlp protocol:
* move all parameters from constructors of each exporter into ConfigOpts, and make that the only parameter to each exporter
* also move exporter-based environment variable config into ConfigOpts
* adjust "insecure" and "compression" header defaults to align to specification
* update all factories and tests which interact with these exporters to use ConfigOpts

Next steps:
* combine SpanConverters (they are almost identical)
* try to combine the Exporters


tidal:
I'm sorry, this PR adds one more issue to https://github.com/open-telemetry/opentelemetry-php/issues/464, while I'm working on solving the remainig issues. There is sadly  is no simple solution, that's why it has taken a while. 

This will also break the Symfony SDKBundle. Breaking the Bundle is not the real issue, but it's breaking it in a way that is hard to fix, since the Config object creates more drift between the different exporters. 

It also makes my work on the exporter factory (including HTTP Client) more difficult. That's why [I had asked Sean](https://github.com/open-telemetry/opentelemetry-php/issues/464), if he is still working on the  options implementation. And that has been the reason for doing the [refactoring](https://github.com/open-telemetry/opentelemetry-php/pull/476) to get closer to having a streamlined approach for all exporters, while introducing a config object only for the Otlp exporters does the opposite. 

The different exporters have different configuration requirements, which additionally may belong to different logical layers/responsibilities (serialization, encryption, compression, transport, authentification, etc.), so one config class to rule them all would be no solution either. However, creating abstractions for the different layers/responsibilities and making those configurable could be.

brettmc:
Thanks @tidal  - I'll park this for a couple of days and try to think through it a bit more. It seemed like a logical way to move towards merging the two OTLP exporters into one and getting rid all of the existing duplication, but I don't want to clobber your efforts!

tidal:
@brettmc 
Thank you, too!!!


> It seemed like a logical way to move towards merging the two OTLP exporters into one and getting rid all of the existing duplication

Yes, I totally agree! However there should be a unified way to instantiate all exporters, especially for the use in  downstream packages/projects. I have been thinking about this for a while now, and tried a few things. I will try create a PR in the next days, and then we can have a look how can intergrate with moving the OTLP exporters into one. 

Anyway, the SpanConverters  are nor really touched by that, so you can still work on them, if you like. Maybe you have seen, that I introduced a [SpanConverterInterface](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Trace/SpanConverterInterface.php) for all the other converters, but it could not be used for the OTLP converters, since there is no "convert" method like in the other converters. If you could make that happen, that would be awesome!!!

Maybe at the end we could have just one single exporter, which uses composition to get configured with different converters, transports and so on and the factory or a builder will take care of plugging all the things together? In that way we could also reduce dupilcation of zipkin, newrelic, zipkin+newrelic and whatever there may be be in the future.

bobstrecansky (approved):

brettmc:
resurrecting for further discussion, but not ready to be merged yet.

PR #489, opened 2021-11-27
PHP 8.1 compatibility

tidal:
PHP 8.1 has been[ released on Nov. 25th](https://www.php.net/archive/2021.php#2021-11-25-1).

So this PR just adds PHP 8.1 to the test matrix.


tidal:
Seems as some packages have to chatch up with PHP8.1.
We can just leave this PR open until the checks work.

bobstrecansky:
Looks like there are some 8.1 build dependencies.  Something we should probably assess.

bobstrecansky:
Looks like the composer package for `qossmic/deptrac-shim` was updated recently; we probably just need to update the version we use (there is `v0.18.0` now)

tidal:
> Run vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php --dry-run --stop-on-violation --using-cache=no -vvv
  vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.php --dry-run --stop-on-violation --using-cache=no -vvv
  shell: /usr/bin/bash -e {0}
PHP needs to be a minimum version of PHP 7.2.5 and maximum version of PHP 8.0.*.
Current PHP version: 8.1.0.
To ignore this requirement please set `PHP_CS_FIXER_IGNORE_ENV`.
If you use PHP version higher than supported, you may experience code modified in a wrong way.

Looks like php-cs-fixer still has problems with 8.1. There are changes to adress this in the master branch, but they are not live, yet. I'll temporarily set the php-cs-fixer version to 'master-dev' just to see, if there are more issues.

tidal:
Ok, PHP 8.1 support in php-cs-fixer `dev-master` seems not to be fully there as well, so I'll revert the version again, and check back in a few days.

bobstrecansky:
Looks like php-cs-fixer requires up to 8.0, but does not have 8.1 as a requirement:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json#L17

I've opened an issue with them in order to get PHP 8.1 support added.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/6145

bobstrecansky:
PHP CS Fixer doesn't yet suport 8.1:

 https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5891

We will either have to wait for that or use a different library.

brettmc:
php-cs-fixer + 8.1 support dropped yesterday (v3.4.0) so this should be good to go, now.
It would be nice to have an "allowed to fail" setting, but I haven't been able to find a way to implement it for github actions...

brettmc:
Getting closer - now it looks like we are waiting on a protobuf fix for 8.1: https://github.com/protocolbuffers/protobuf/issues/9293
The complaints in integration tests look reasonable and easy enough to fix (move required parameters to before optional)

brettmc:
I've been playing with allowing some jobs to fail and found https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run
I think this is the way to do it:
```yaml
strategy:
  fail-fast: false
  matrix:
    php-versions: ['7.4', '8.0']
    experimental: [false]
    include:
      - php-versions: 8.1
        experimental: true
runs-on: ubuntu-latest
    continue-on-error: ${{ matrix.experimental }}
```

tidal:
> I've been playing with allowing some jobs to fail and found https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run I think this is the way to do it:
> 
> ```yaml
> strategy:
>   fail-fast: false
>   matrix:
>     php-versions: ['7.4', '8.0']
>     experimental: [false]
>     include:
>       - php-versions: 8.1
>         experimental: true
> runs-on: ubuntu-latest
>     continue-on-error: ${{ matrix.experimental }}
> ```

If we allow the PHP8.1 checks to fail, there is no much value in running them in the first place imo.

tidal:
> Getting closer - now it looks like we are waiting on a protobuf fix for 8.1: [protocolbuffers/protobuf#9293](https://github.com/protocolbuffers/protobuf/issues/9293) The complaints in integration tests look reasonable and easy enough to fix (move required parameters to before optional)

I don't want to wait on  protobuf to fix things. The is just an indicator for our code being to tightly coupled to the protobuf library, so stuff is leaking into our unit tests. I will address this in a different PR. 

However there is still a problem with with this [line](https://github.com/open-telemetry/opentelemetry-php/blob/77a103b38c66e4c029d50bc26ece0cea95a089ea/src/Context/fiber/zend_observer_fiber.php#L34), which lets the unit test run crash. We should discuss in the next meeting, if we want to support such edge cases like context switching inside of a Fiber in the main library by using FFI. I don't think the dev images have even FFI installed, but I have tested this on my local system.

brettmc:
> If we allow the PHP8.1 checks to fail, there is no much value in running them in the first place imo, or maybe I'm just not understanding, what you are trying to achieve by doing this.
It means we can start running our tests against new PHP versions as soon as possible (before they go GA), and find issues earlier. In the case of 8.1, once it does go green, we switch off "experimental" and then it must always pass.

tidal:
> It means we can start running our tests against new PHP versions as soon as possible (before they go GA), and find issues earlier. In the case of 8.1, once it does go green, we switch off "experimental" and then it must always pass.

Ok, got you.  let's talk about this in tomorrows meeting, as the FFI issue will not just turn green, the tests are crashing after 51%.

brettmc:
Related issue: https://github.com/open-telemetry/opentelemetry-php/issues/515

brettmc (approved):

PR #492, opened 2021-11-29
Convert and send zipkin span kind in string format

PauliusB:
Zipkin span exporter currently serializes and sends span kind in integer format. However, Zipkin span api endpoint seems to only accept them in strings format (https://zipkin.io/zipkin-api/#/default/post_spans). This results in rejected method calls with status code 400 when using any span kind other that internal. This PR converts spans to required format to fix rejected method calls to Zipkin endpoints.

### DIFF
--- a/src/Contrib/Zipkin/SpanConverter.php
+++ b/src/Contrib/Zipkin/SpanConverter.php
@@ -170,12 +170,12 @@ class SpanConverter implements SpanConverterInterface
         if (count($event->getAttributes()) === 0) {
             return null;
         }
-        
+
         $attributesArray = [];
         foreach ($event->getAttributes() as $attr) {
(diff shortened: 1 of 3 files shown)


Grunet:
@PauliusB Other than the CLA being annoying, was there a particular reason you closed this PR? (If you join in the slack.cncf.io channel I think we can find someone to get that sorted out)

AFAICT this looks spot on, as it seems like in Java enum values default to their name (so the [Java implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/SpanKind.java) would do this by default) and the [JS implementation](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-exporter-zipkin/src/transform.ts#L23) seems to be doing an explicit conversion like this as well.

PR #493, opened 2021-11-29
Convert and send zipkin span kind in string format

PauliusB:
Zipkin span exporter currently serializes and sends span kind in integer format. However, Zipkin span api endpoint seems to only accept them in strings format (https://zipkin.io/zipkin-api/#/default/post_spans). This results in rejected method calls with status code 400 when using any span kind other that internal. This PR converts spans to required format to fix rejected method calls to Zipkin endpoints.


src/Contrib/Zipkin/SpanKind.php
```diff
@@ -0,0 +1,27 @@
...
+namespace OpenTelemetry\Contrib\Zipkin;
+
+final class SpanKind
+{
+    public const KIND_SERVER = 'SERVER';
+    public const KIND_CLIENT = 'CLIENT';
+    public const KIND_PRODUCER = 'PRODUCER';
+    public const KIND_CONSUMER = 'CONSUMER';
+
+    public static function getChoices(): array
```
brettmc, line 14: this looks like it is not used? and the following constructor is redundant too

tidal: @brettmc 
This is following the current implementation of [API/Trace/SpanKind](https://github.com/open-telemetry/opentelemetry-php/blob/01c388e02cd0dd94915475d4b4032bbd8aaa8e93/src/API/Trace/SpanKind.php)

However the choices there could be a constant as well:
```php
public const CHOICES = [
            self::KIND_CLIENT,
            self::KIND_SERVER,
            self::KIND_PRODUCER,
            self::KIND_CONSUMER,
        ];
        
```
While I don't know if if "choices" is the best wording here. These are more like roles imho.

brettmc: Right, that makes sense. I don't see any evidence of them being used (and it is not covered by any tests) - but I could see it being potentially useful for SDK users etc.

PauliusB: Yeah, I've tried making similar implementation of [API/Trace/SpanKind](https://github.com/open-telemetry/opentelemetry-php/blob/01c388e02cd0dd94915475d4b4032bbd8aaa8e93/src/API/Trace/SpanKind.php) but I'm not too sure that was necessary. Currently there's no real usage of `getChoices` so I could simply leave this class with constants only. Potentially it could become a pure enum in latter versions of php.

Oh and another thought about these constants - having `KIND_` prefix on them is a bit redundant since constant purpose is already implied with class name. Maybe I should remove it to have usages like `SpanKind::CLIENT` instead of `SpanKind::KIND_CLIENT`?

tidal: > Oh and another thought about these constants - having KIND_ prefix on them is a bit redundant since constant purpose is already implied with class name. Maybe I should remove it to have usages like SpanKind::CLIENT instead of SpanKind::KIND_CLIENT?

I have noticed that, too. Just go ahead and remove the "SPAN_" prefixes. I Think it looks cleaner that way.

PauliusB: I've renamed SpanKind constants and removed the `getChoices` method. Let me know if there's any additional changes you want to be done :)

tidal: Look good to me. Thanks for your contribution!

>  Let me know if there's any additional changes you want to be done :)

You could change the original [API/Trace/SpanKind](https://github.com/open-telemetry/opentelemetry-php/blob/01c388e02cd0dd94915475d4b4032bbd8aaa8e93/src/API/Trace/SpanKind.php) accordingly, if you want, but that would probably be better done in another PR.

Grunet:
Just a thought, not a requirement to get this merged, but would it make sense to update the [SpanKind's constants](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/API/Trace/SpanKind.php) to use strings instead of integers? That's what it seems like [Java is doing by default](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/SpanKind.java), but I'm not sure what ripple effects it might have.

FWIW I noticed [JS](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-exporter-zipkin/src/transform.ts#L23) is doing this mapping approach, so it's probably nbd.

bobstrecansky:
@PauliusB - what do you think of @Grunet 's idea?

PauliusB:
Yeah, I would agree on string format. Integers currently imply protocol specific values which is probably not the case here. Using strings in the generic SpanKind class would be more abstract approach and could be mapped to required values on each span converter if needed. If other people agree with this, I could add related changes to this (or a new) PR

tidal:
I would prefer any changes to [SpanKind ](https://github.com/open-telemetry/opentelemetry-php/blob/01c388e02cd0dd94915475d4b4032bbd8aaa8e93/src/API/Trace/SpanKind.php) to go into a separate PR. 
This PR is mainly about fixing a bug, and changing the signature of the SpanKind class would need a lot more changes through out the library.

bobstrecansky:
One small merge conflict as well.

tidal:
@bobstrecansky 
LGTM!

bobstrecansky (approved):

PR #495, opened 2021-11-30
creating a global logger holder

brettmc:
since logging errors can live in all parts of the sdk (logs, metrics, traces), creating a singleton
in the top-level SDK namespace seems the best place (and avoids passing loggers all through the codebase).
This commit just gets the mechanism in place and implements a couple of log messages to get feedback on whether this is a good way to go


tidal:
> since logging errors can live in all parts of the sdk (logs, metrics, traces), creating a singleton
> in the top-level SDK namespace seems the best place (and avoids passing loggers all through the codebase).
> This commit just gets the mechanism in place and implements a couple of log messages to get feedback on whether this is a good way to go

To be honest, I am not a big fan of global states, singletons, etc. at all. They may solve some problems, but create others down the line. For example the benchmark metrics are kind of bogus (when run in the same process) with all the global states already in the library. That's true for classic PHP runtimes. (I nowadays perefer Swoole/Swoft as they outperform NodeJS and Go, but there you have different ways to hold state, and even can share state between proccesses.)

I know people have different opinions regarding that matter, but I think global state creates more problems than it solves, otherwise nobody would use dependency injection, but throw everything into global state (I can remember, when things were done that way, and I personally don't miss those times) 

You could achieve the same by using the [LoggerAwareTrait](https://github.com/open-telemetry/opentelemetry-php/blob/d497f4e745204161b90b9ee455476e19ba945f59/src/SDK/Trace/Behavior/LoggerAwareTrait.php). I don't see any problem passing the logger to other objects. This is not much overhead compared to querying a global logger, and on the other hand it could allow for more fine grained control, if needed. If you are worrying about code duplication, there could be some kind of "LoggerInjector" class, which takes an object and injects the Logger, when the class implements the PSR `LoggerInterface`

However, those are just my €0.02.

bobstrecansky:
This seems like a good topic to discuss in the SIG meeting tomorrow.  I'll add it to the agenda!

bobstrecansky:
@brettmc - are there other things you'd like to discuss with this PR?  I know we covered a bit in the SIG meeting; just wanted to make sure you didn't have any open ended questions.

brettmc:
@bobstrecansky - this is still waiting on me. In SIG we discussed whether `LoggerAwareTrait` was a better approach to singleton, and I was going to prototype that for comparison.

tidal (approved):
For the record: 
Me approving this PR does not change my opinion on global states. :)

PR #496, opened 2021-11-30
removing span processor factory

brettmc:
feedback from spec SIG group in https://github.com/open-telemetry/opentelemetry-specification/pull/2172#issuecomment-982312473 is that defaulting to BatchSpanProcessor is preferred, and not to have an env var for OTLP_TRACES_PROCESSOR


brettmc:
I think that the code coverage drop is just due to having removed code - the changed files have 100% coverage

tidal:
Maybe instead of removing the factory, we can use a [Language Specific Environment Variable](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#language-specific-environment-variables).

Let's talk about this in the meeting as well, as I simply don't buy any of the given argumentation.

brettmc:
Discussed in SIG meeting today, will switch over to a language-specific env var, ie OTEL_PHP_TRACES_PROCESSOR

Issue #301, opened 2021-03-31
Configuration Interface and default configuration


jodeev:
We need to implement the [Default SDK Configuration](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/sdk-configuration.md#default-sdk-configuration).

This has two pieces:
* The SDK MUST provide a programmatic interface for all configuration. This interface SHOULD be written in the language of the SDK itself. All other configuration mechanisms SHOULD be built on top of this interface. An example of this programmatic interface is accepting a well-defined struct on an SDK builder class. From that, one could build a CLI that accepts a file (YAML, JSON, TOML, ...) and then transforms into that well-defined struct consumable by the programatic interface.
* Establish a default configuration

tidal:
I can take this on (after I finished the exporter factory issues, which however relate to this), and I guess no one is working this yet.
Basically I solved a lot of he tasks involved already in the [Symfony SDKBundle](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/bcd4c4a5996e41fe47615928f2e032bc05b04203/instrumentation/symfony/OtelSdkBundle), though there need to be some adjustments.

Regarding the example from the spec:

Structs will only ship with PHP8.1 as first class citizens, so that's not an option, however in principle that's what the SF configuration does, only that it's a bit more flexible than structs themselve. I don't even see many places where structs would make sense, if we want to leave the SDK open for extension as mentioned somewhere else in the spec.

Creating a CLI is very simple , defacto standard is to use the [Console Component](https://symfony.com/doc/current/components/console.html), however the config can be cached anyway, so there might be no real need for that 

I think the best way is to go with some dependeny injection container to build the SDK. Symfony of course uses it's own [PSR11 (Container Interface)](https://www.php-fig.org/psr/psr-11/) implementation which is very powerful, but I would not be set to this one. There are many PSR11 implementations, which all have their pros and cons, so I would rather evaluate appropriate options. However I think DI could also help remove some of the code smells currently in the library (like inline creation of instances for example, which is a problem with testing, debugging and SRP).

tidal:
Another option for configuration might be [thephpleague/config](https://github.com/thephpleague/config). It's a more simplistic approach than [symfony/config](https://symfony.com/doc/current/components/config.html) and creates a data structure, which comes close to a 'well-defined struct'.

brettmc:
[...]

PR #498, opened 2021-12-02
experimental: config builder + nette schema

brettmc:
using nette/schema to generate a config from environment + user-supplied variables
- user vars replace env vars
- some defaults provided in config (more could be?)
- enforces config structure and types (where possible)
- switching some factories to `fromConfig` (more to do)


src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+    {
+        //@phan-ignore PhanUndeclaredClassMethod
+        $schema = Expect::structure([
+            'log' => Expect::structure([
+                'stream' => Expect::string('php://stdout'),
+                'level' => Expect::string('info'),
+            ]),
+            'propagators' => Expect::string('tracecontext,baggage'),
+            'service' => Expect::structure([
+                'name' => Expect::string(),
```
brettmc, line 71: we could add all defaults here, and remove from the various classes that look up env vars

tidal:
Looks good on first sight. However I will add some feeback.

One thing thing I found in the repos of other SIGs is, that they often have a `experimental` directory besides the `src` one. Maybe we can adport this, so it would be easier to collaborate on certain things, and once everybody is happy, we move it to source?
One reason is that I find it way easier to review code locally, than in github's PR GUI (There are tools to help with this, but they don't work with the restrictions of the OTEL repos afaik)

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK;
+
+use Nette\Schema\Expect;
+use Nette\Schema\Processor;
+
+/** @phan-file-suppress PhanUndeclaredClassMethod */
+
+class ConfigBuilder
```
tidal, line 12: I think this ConfigBuilder is doing to much. It's a) defining a configuration schema, and b) acting upon user definded values via env vars. I think the providers you outlined [here](https://github.com/open-telemetry/opentelemetry-php/issues/301#issuecomment-984091668) are a cleaner approach to deal with user provided values and allow for single responsibility.

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+                        'export_timeout' => Expect::string()->castTo('int'),
+                        'max_queue_size' => Expect::string()->castTo('int'),
+                        'max_export_batch_size' => Expect::string()->castTo('int'),
+                    ]),
+                    'simple' => Expect::structure([]),
+                    'noop' => Expect::structure([]),
+                ]),
+                'exporter' => Expect::string('grpc'),
+                'exporters' => Expect::structure([
+                    'zipkin' => Expect::structure([
```
tidal, line 100: I would prefer to not have the schemas for the contrib exporters here. This is of course not a direct dependency on the Contrib "package", but a indirect one. It's expecting something which may not be there.

Also this looks like it's always expecting configuration for all possible exporters, no?

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+                    'attribute_value_length' => Expect::string()->castTo('int'),
+                    'event_count' => Expect::string()->castTo('int'),
+                    'link_count' => Expect::string()->castTo('int'),
+                    'attribute_per_event' => Expect::string()->castTo('int'),
+                    'attribute_per_link' => Expect::string()->castTo('int'),
+                ]),
+            ]),
+        ]);
+        $env = $this->env();
+        $resourceAttributes = $this->resourceAttributes();
```
tidal, line 136: I don't think temporary variables help readability.

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Retrieve all OTEL_* environment variables, and map them to a nested array
+     */
+    private function env(): array
+    {
+        $vars = array_filter(getenv(), function ($key) {
+            return strpos($key, 'OTEL_') !== false;
+        }, ARRAY_FILTER_USE_KEY);
+        $output = [];
+        foreach ($vars as $key => $val) {
```
tidal, line 150: A long time ago I read a [book on code readebilty](https://www.oreilly.com/library/view/the-art-of/9781449318482/), so you can blame the authors for me complaininga bout this stuff. :)

 However while a lot of things there seam to be nit-picky at first, they make sense, when one thinks about and get used to them. One point is to not use "short variable names". While you and maybe me as well know what's meant with "$val", it might not be obviouss to everyone.

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+
+    public function build(array $data = []): object
+    {
+        //@phan-ignore PhanUndeclaredClassMethod
+        $schema = Expect::structure([
+            'log' => Expect::structure([
+                'stream' => Expect::string('php://stdout'),
+                'level' => Expect::string('info'),
+            ]),
+            'propagators' => Expect::string('tracecontext,baggage'),
```
tidal, line 69: Why do you expect a CSV here? I Think it's cleaner to expect an array and normalize values from the env var accordingly.

src/SDK/ConfigBuilder.php
```diff
@@ -0,0 +1,199 @@
...
+            ]),
+            'resource' => Expect::structure([
+                'limits' => Expect::structure([
+                    'attribute_count' => Expect::string()->castTo('int'),
+                    'attribute_value_length' => Expect::string()->castTo('int'),
+                ]),
+                'attributes' => Expect::array(),
+            ]),
+            'trace' => Expect::structure([
+                'sampler' => Expect::string('parentbased_always_on'),
```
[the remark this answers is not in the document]
brettmc: I found this method easier to reason about, because the config options for different types of sampler (or exporter/processor/etc) are easily found within their respective key in "samplers", where "sampler" is the list of samplers to be used.
So, iterate over "sampler" to work out which to create, and then look up sampler-specific config in samplers[type] as required.

src/SDK/ConfigBuilder.php
brettmc: Yes, that makes sense. It does raise the question though - I'm starting to think that the two otlp exporters should not be in `contrib` but rather in `SDK` - this is the opentelementry project, after all.

src/SDK/ConfigBuilder.php
brettmc: I had two different techniques going during development, and originally CSV values were extracted into an array as you've suggested. I went back to the CSV method because arrays did not work with default values (ie, if you define default propagator(s) as array values, they are not removable).
So it looks like with nette/schema, we really need to choose between having it handle default values, or having it just enforce the schema (and responsibility for default values stays in the various factories/constructors). I'm happy for it to just enforce the structure.

src/SDK/ConfigBuilder.php
tidal: >  this is the opentelementry project, after all.

That's true, however from the perspective of the library the OTLP collectors are 3rd party applications, which just happen to live under the same umbrella organization. Moving the grpc exporter into the SDK would also make the corresponding extension a must have dpendency for the SDK again. And since there must be a solution to handle Zipkin & Jaeger and Co. in the contrin directory, it's not much of a hassle to do the same for the otlp exporters.

brettmc:
@tidal - a change of direction. I've created an `experimental` dir as you suggested, and have two different approaches:
* `Experimental\NetteConfig` - the first approach (unchanged), using nette schema to create one big schema of all possible config
* `Experimental\Config` - instead of nette schema, use classes and interfaces and a ConfigBuilder. I like this approach more:
  * exporters and span processors must be made available to the builder (but the main ones are there by default), which I think will help to remove the sdk->contrib dependency
  * config classes are responsible for only configuring themselves
  * `experimental\examples\ConfigExample.php` demonstrates creating a config from a mixture of env and user-provided input. The output is a mixture of arrays and objects - lazy but it's experimental :)
I have not tried very hard to make it pretty, since it might all end up on the scrap heap.

experimental/examples/ConfigExample.php
```diff
@@ -0,0 +1,36 @@
...
+putenv('OTEL_EXPORTER_ZIPKIN_ENDPOINT=http:zipkin/v1/traces');
+putenv('OTEL_EXPORTER_ZIPKIN_TIMEOUT=99');
+putenv('OTEL_PHP_TRACES_PROCESSOR=batch,simple,noop');
+putenv('OTEL_BSP_SCHEDULE_DELAY=10000');
+putenv('OTEL_ATTRIBUTE_COUNT_LIMIT=111');
+putenv('OTEL_PROPAGATORS=tracecontext');
+
+echo 'Creating Config From Environment and user config' . PHP_EOL;
+$config = (new ConfigBuilder())
+    ->withExporterConfig(ZipkinConfig::class)
```
tidal, line 26: What's the point of expecting a class, when this has to be added programmatically anyway?

experimental/examples/ConfigExample.php
```diff
@@ -0,0 +1,36 @@
...
+    ->withExporterConfig(NewRelicConfig::class)
+    ->withUserConfig([
+        'span.processor.batch.max_queue_size' => 333,
```
tidal, line 29: What's the point of this dot-notation? This has to be resolved on every request?

experimental/examples/ConfigExample.php
```diff
@@ -0,0 +1,36 @@
...
+        'resource.limits.attribute_value_length' => 444,
+        'exporter.new_relic.license_key' => 'secret',
```
tidal, line 31: So I can only have one newrelic exporter?

experimental/examples/ConfigExample.php
```diff
@@ -0,0 +1,36 @@
...
+putenv('OTEL_TRACES_SAMPLER_ARG=0.95');
+putenv('OTEL_TRACES_EXPORTER=otlp,zipkin,newrelic');
+putenv('OTEL_EXPORTER_ZIPKIN_ENDPOINT=http:zipkin/v1/traces');
+putenv('OTEL_EXPORTER_ZIPKIN_TIMEOUT=99');
+putenv('OTEL_PHP_TRACES_PROCESSOR=batch,simple,noop');
+putenv('OTEL_BSP_SCHEDULE_DELAY=10000');
+putenv('OTEL_ATTRIBUTE_COUNT_LIMIT=111');
+putenv('OTEL_PROPAGATORS=tracecontext');
+
+echo 'Creating Config From Environment and user config' . PHP_EOL;
```
tidal, line 24: What is "user config"? are env vars not provided by users?

experimental/src/Config/Config.php
```diff
@@ -0,0 +1,41 @@
...
+class Config implements ConfigInterface
+{
+    public array $log;
+    public array $exporters = [];
+    public array $spanProcessors = [];
+    public array $sampler = [];
+    public array $service;
+    public ResourceConfig $resource;
+
+    public function __construct(array $userConfig, array $environmentConfig)
```
tidal, line 16: What's the point of having two arguments, istead of an array of configurations?

experimental/src/Config/Config.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $this->log = [
+            'level' => $userConfig['log.level'] ?? $environmentConfig['OTEL_LOG_LEVEL'] ?? null,
```
tidal, line 19: Why does the environment config have the env var names as keys?

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Find exporter config provider for requested exporter(s)
+     */
+    private function buildExporters(): array
+    {
+        $configs = [];
+        foreach ($this->userConfig['exporters'] ?? explode(',', $this->environmentConfig['OTEL_TRACES_EXPORTER'] ?? '') as $name) {
+            foreach ($this->exporters as $klass) {
```
tidal, line 76: Why do you use the Swedish word for "class"?

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+        }, ARRAY_FILTER_USE_BOTH);
+    }
+
+    /**
+     * Find exporter config provider for requested exporter(s)
+     */
+    private function buildExporters(): array
+    {
+        $configs = [];
+        foreach ($this->userConfig['exporters'] ?? explode(',', $this->environmentConfig['OTEL_TRACES_EXPORTER'] ?? '') as $name) {
```
tidal, line 75: What's the point of this expensive double loop?

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+        return $configs;
+    }
+
+    /**
+     * Find span processor config provider for requested span processor(s)
+     */
+    private function buildSpanProcessors(): array
+    {
+        $configs = [];
+        foreach ($this->userConfig['span.processors'] ?? explode(',', $this->environmentConfig['OTEL_PHP_TRACES_PROCESSOR'] ?? '') as $name) {
```
tidal, line 93: see above

experimental/src/Config/ConfigInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Experimental\Config;
+
+interface ConfigInterface
+{
+    public function __construct(array $userConfig, array $environmentConfig);
```
tidal, line 9: What's the point of having an interface with a constructor (and only a constructor)?

experimental/src/Config/Config.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Experimental\Config;
+
+class Config implements ConfigInterface
+{
+    public array $log;
```
tidal, line 9: This can create a runtime error without an default empty error.

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+namespace OpenTelemetry\Experimental\Config;
+
+use OpenTelemetry\Experimental\Config\Exporter\Otlp\Config as OtlpConfig;
+use OpenTelemetry\Experimental\Config\SpanProcessor\BatchSpanProcessorConfig;
+use OpenTelemetry\Experimental\Config\SpanProcessor\NoopSpanProcessorConfig;
+use OpenTelemetry\Experimental\Config\SpanProcessor\SimpleSpanProcessorConfig;
+
+class ConfigBuilder
+{
+    //initial list of known exporter types (exluding exotic contrib exporters, which must be added)
```
tidal, line 14: Who "knows" those exporters?

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+            $config->addExporter($name, $exporterConfig);
+        }
+        foreach ($this->buildSpanProcessors() as $name => $spanProcessorConfig) {
+            $config->addSpanProcessor($name, $spanProcessorConfig);
+        }
+
+        return $config;
+    }
+
+    public function withUserConfig(array $config): self
```
tidal, line 45: With a "with" prefix people usually  expect an immutable class/object.

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Retrieve all non-empty OTEL_* environment variables
+     */
+    private function getEnvironmentConfig(): array
+    {
+        return array_filter(getenv(), function ($value, $key) {
```
tidal, line 64: This is like "Select * from database". 
We just cannot make assumptions on how many env vars there are, and this seem like it has to be done with every request again. This could be easily solved, if the env vars were constants.

experimental/src/Config/Exporter/Otlp/Config.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function __construct(array $userConfig, array $environmentConfig)
+    {
+        $this->endpoint = $userConfig['exporter.otlp.endpoint'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_ENDPOINT'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.insecure'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_INSECURE'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.certificate_file'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_CERTIFICATE'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.headers'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_HEADERS'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.compression'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_COMPRESSION'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.timeout'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_TIMEOUT'] ?? null;
+        $this->endpoint = $userConfig['exporter.otlp.protocol'] ?? $environmentConfig['OTEL_EXPORTER_OTLP_PROTOCOL'] ?? null;
```
tidal, line 28: Looks like everyone wants to be an endpoint.

experimental/src/Config/Exporter/NewRelic/Config.php
```diff
@@ -0,0 +1,30 @@
...
+    public ?string $endpoint;
+    public ?string $licenseKey;
+
+    public function __construct(array $userConfig, array $environmentConfig)
+    {
+        $this->endpoint = $userConfig['exporter.new_relic.endpoint'] ?? $environmentConfig['OTEL_EXPORTER_NEWRELIC_ENDPOINT'] ?? null;
+        $this->licenseKey = $userConfig['exporter.new_relic.license_key'] ?? $environmentConfig['OTEL_EXPORTER_NEWRELIC_LICENSE_KEY'] ?? null;
+    }
+
+    public static function provides(string $exporterName): bool
```
tidal, line 21: Whats's the point of this method, when there is a method, which provides the name?

experimental/src/Config/ConfigBuilder.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function __construct()
+    {
+        $this->environmentConfig = $this->getEnvironmentConfig();
+    }
+
+    public function build(): object
+    {
+        $config = new Config($this->userConfig, $this->environmentConfig);
+        foreach ($this->buildExporters() as $name => $exporterConfig) {
```
tidal, line 35: What#s the point of passing the whole config tree into dedicated configs? What's the point of having them then in the first place, as this is hard coupling them with the global config schema.

experimental/src/Config/Exporter/Zipkin/Config.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Experimental\Config\Exporter\Zipkin;
+
+use OpenTelemetry\Experimental\Config\ConfigInterface;
+use OpenTelemetry\Experimental\Config\ExporterConfigInterface;
+
+class Config implements ConfigInterface, ExporterConfigInterface
+{
+    public ?string $endpoint;
```
tidal, line 12: This will create a runtime error without defaulting to null.

experimental/src/Config/ResourceLimitsConfig.php
```diff
@@ -0,0 +1,22 @@
...
+    public ?int $attributeCount;
+    public ?int $attributeValueCount;
+
+    public function __construct(array $userConfig, array $environmentConfig)
+    {
+        $this->attributeCount = $this->intOrNull($userConfig['resource.limits.attribute_count'] ?? $environmentConfig['OTEL_ATTRIBUTE_COUNT_LIMIT'] ?? null);
+        $this->attributeValueCount = (int) ($userConfig['resource.limits.attribute_value_length'] ?? $environmentConfig['OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT'] ?? 128);
+    }
+
+    private function intOrNull($value): ?int
```
tidal, line 18: Is this a question?

experimental/src/Config/SpanProcessor/NoopSpanProcessorConfig.php
```diff
@@ -0,0 +1,20 @@
...
+namespace OpenTelemetry\Experimental\Config\SpanProcessor;
+
+use OpenTelemetry\Experimental\Config\ConfigInterface;
+use OpenTelemetry\Experimental\Config\SpanProcessorConfigInterface;
+
+class NoopSpanProcessorConfig implements ConfigInterface, SpanProcessorConfigInterface
+{
+    public function __construct(array $userConfig, array $environmentConfig)
+    {
+    }
```
tidal, line 14: ???

src/SDK/ConfigBuilder.php
tidal: I don't understand. Why should there be default propagators and why should there be defaults for array items? 
However I think it's better anyway to have classes or factories handle defaults than maintaining them in two places, including possible bugs.

tidal:
Nos têtes sont rondes pour que nos pensées puissent changer de direction. (Francis Picabia)

However your new approach looks a bit rough ,so I can only provide some "rough" comments, I'm afraid. And it looks like it's only tracing configuration in a way that metrics and logger can't be added.

In general here is no need for the for league/config or nette/schema to be one big ball of mud, they can be 
merged as well. 

> * which I think will help to remove the sdk->contrib dependency

Not like it's implemented at the moment. This just adds an additional layer of indirection and possibly other dependency problems.

> config classes are responsible for only configuring themselves

Maybe, but they seems to have dependencies on the complete config tree.

experimental/examples/ConfigExample.php
brettmc: My thinking here is to allow making contrib exporters available without directly referencing `contrib` from `sdk`, and potentially to make them settable from an env var (OTLP_PHP_EXPORTER_CLASS ?).
Without some sort of discovery mechanism I don't see a way to have something in `sdk` that is capable of configuring classes from `contrib`. Perhaps a `class_exists` for something like `Contrib\ExporterBuilder`, and the configurations for these guys actually does live in `contrib`, and if that class does not exist then only the SDK SpanExporters are available.

experimental/examples/ConfigExample.php
brettmc: that's a "todo" - it should require a proper nested config as might be created from yaml/json/whatever.

experimental/examples/ConfigExample.php
brettmc: It never occurred to me that somebody would want to have multiple instances of the same exporter. In fact, the Batch and Simple span processors only allow for 0..1 exporter. MultiSpanProcessor does look like it allows any number of span processors, each of which could have their own exporter though - so I guess you can!

experimental/examples/ConfigExample.php
brettmc: True, they are. I could merge them all together, but haven't because it requires sdk->contrib references - at least, the way I'm thinking of it does. As mentioned in an earlier comment, this might be resolved by either having that logic sit in `contrib`, or even removing env vars from feature and keeping that knowledge where it is now, in the various constructors.

experimental/src/Config/ConfigBuilder.php
brettmc: I think I see an improvement here. I think that the config builder should creates all of the different sub-configs and assemble them, passing to each only the config that they need.

experimental/src/Config/SpanProcessor/NoopSpanProcessorConfig.php
brettmc: It's because the constructor is defined in an interface...I will change it.

experimental/examples/ConfigExample.php
tidal: > It never occurred to me that somebody would want to have multiple instances of the same exporter.

I guess you never had to develop packages/libraries to be used by customers/clients. They sometimes can come up with let's say kind of "exotic" requests or requierements, that's why I'm used to at least consider "exotic" requirements to be prepared. ;)

experimental/examples/ConfigExample.php
tidal: Ok, got you, I just find the constructor a bit clunky in that case. But I think you commented something regarding this below.

experimental/src/Config/SpanProcessor/NoopSpanProcessorConfig.php
tidal: @brettmc 
I know, this was just a lazy comment from me.  :)

brettmc:
Discussed in SIG, I will close this and put the related issue on hold. https://github.com/open-telemetry/opentelemetry-specification/issues/2207 may help a lot here, by defining an SDK configuration schema.

Issue #482, opened 2021-11-19
Proposal: Make gRPC PHP  extension optional


tidal:
Proposed change: 
- Move grpc-ext to dev requirements in composer.json
- Add a suggest entry for grpc-ext in composer.json
- Document in README (and docs). see: https://github.com/open-telemetry/opentelemetry-php/issues/481

Reasons for the proposal:
- The gRPC extension is not a dependency of the API or the SDK, but only one contrib exporter (atm).
- Compiling the extension can takes up to 10-20 minutes (Yes, this can be cached in container layers etc, but it can still be an issue in cascading build systems and other places)
- The binary of the gRPC extension can be as large as 100mb. That's not huge, it's massive.! (On my local system the binary was 101mb after compiling for PHP8. All other extension are around 1mb. Only swoole is 30mb, but it does a few more things than gRPC. Yes, the size of the binary can be reduced, [see this thread](https://github.com/grpc/grpc/issues/23626), but I don't think users should be expected to dig through github issues)
- A lot of providers and systems (Heroku or OpenShift's S2I for PHP for example), etc. still regard the grpc extension
 as kind of 'exotic', or in other words the extension is not (or cannot be) installed.
- Because of the above issues, there will propbably be Ops people, who will refuse or be reluctant to install the extension, when it's not really needed. (I have asked some), and thus the whole SDK cannot be used.

I am not aware, why the gRPC extension is a standard requirement atm, so there might be something I'm missing here.

brettmc:
I assumed this would go ahead, and so made http+protobuf the default in #488

Issue #481, opened 2021-11-19
Update README


tidal:
The readme is a bit out of sync with latest development and could use some polishing as well.

- [X] Current requirements to install the library should be documented.
- Maybe add a short documentation of additional (or all) make targets.
- There is a lot of redundancy, which does not really help readability.

tidal:
@stale
jehwoandersmakiern!

yuktea:
@tidal I'll work on this 🙂

PR #501, opened 2021-12-05
Make gRPC PHP extension optional

tidal:
closes: https://github.com/open-telemetry/opentelemetry-php/issues/482

As discussed at the meeting, this PR will remove the hard dependecy on the PHP gRPC extension and move it to be a dev requirement. Additionally a "suggest" section is added to composer.json with information, when the ext-grpc is needed.

The gRPC extension may be a hard dependency of a separate Contrib package in the future, if packages are published individually.

The documentation of this change will be addressed in https://github.com/open-telemetry/opentelemetry-php/issues/481, as this depdendency hasn't been documentated before anyway.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,6 @@
     "license": "Apache-2.0",
     "require": {
         "php": "^7.4 || ^8.0",
-        "ext-grpc": "*",
         "ext-json": "*",
         "google/protobuf": "^v3.3.0",
         "grpc/grpc": "^1.30",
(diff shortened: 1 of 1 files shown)


brettmc:
Does it need an `extension_exists` check somewhere (probably the exporter constructor?)

tidal:
@brettmc 
> Does it need an extension_loaded check somewhere (probably the exporter constructor?)

Not in the exporter constructor. What could else could we do there besides throwing an exception? 
I'd prefer to have a hard error in this case.

There are a lot of packages which work in that way, that you need additionally dependencies to use certain components/features. If I want to use `Doctrine` with `SQLite` and don't have the sqlite extsnion installed it just creates an error as well. I think we should expect basic troubleshooting skills from users, especially when the requirement will be added to the readme/docs

However there may could be such checks in configuration handling, factories, instrumenting ocde etc.

bobstrecansky (approved):

PR #502, opened 2021-12-05
Add semantic conventions for Resource and Trace attributes

### LINKED ISSUE #376: Semantic Convention Generator
*Is your feature request related to a problem?*
This is an issue that covers similar ideas to issue #297 

Currently there is no semantic convention generator to generate constants

*Describe the solution you'd like*

A semantic convention generator should be added as detailed in the following specifications:

https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions

*Describe alternatives you've considered*
No other alternatives. This is necessary for GA release

lalex:
This PR adds semantic conventions for Resource and Trace attributes.

Fixes #376.

As spec [says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#semantic-conventions), semconv is a candidate for the separated package. So generated classes are in their own namespace `\OpenTelemetry\SemConv`.

There is a script to generate PHP classes from yaml semantic conventions files, (usage `SEMCONV_VERSION=1.8.0 make semconv`).

Deprected class `\OpenTelemetry\SDK\Resource\ResourceConstants`.

### DIFF
--- a/script/semantic-conventions/templates/AttributeValues.php.j2
+++ b/script/semantic-conventions/templates/AttributeValues.php.j2
@@ -8,6 +8,11 @@ namespace {{ namespace }};
 
 class {{ class }}AttributeValues
 {
+    /**
+     * The URL of the OpenTelemetry schema for these keys and values.
+     */
+    public const SCHEMA_URL = '{{schemaUrl}}';
+
 {%- for attribute in attributes | unique(attribute="fqn") %}
 {%- if attribute.attr_type.members %}
 {%- for member in attribute.attr_type.members %}
--- a/script/semantic-conventions/templates/AttributeValues.php.j2
+++ b/script/semantic-conventions/templates/AttributeValues.php.j2
@@ -1,0 +1,30 @@
+<?php
+
+// DO NOT EDIT, this is an Auto-generated file from script/semantic-convention{{template}}
+
+declare(strict_types=1);
+
+namespace {{ namespace }};
+
+class {{ class }}AttributeValues
+{
+{%- for attribute in attributes | unique(attribute="fqn") %}
+{%- if attribute.attr_type.members %}
+{%- for member in attribute.attr_type.members %}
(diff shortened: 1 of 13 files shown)


script/semantic-conventions/templates/AttributeValues.php.j2
```diff
@@ -0,0 +1,35 @@
+<?php
```
bobstrecansky, line 1: should we perhaps add this to the `.gitignore` file, or are we purposefully vendoring these into a commit?

bobstrecansky (commented):
One open question about committing generated files, but otherwise I very much like the direction.  Thanks for your hard work on this @lalex !

script/semantic-conventions/templates/AttributeValues.php.j2
lalex: This is a template for the [code generation](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator). It will be used every time the semantic conventions are updated in [spec repo](https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions) to generate an updated PHP code.

A workflow will be like this:
- checkout opentelemetry-php repo
- install composer dev requirements
- make semconv
- commit and push an updated generated files

bobstrecansky: Figured as much, just wanted to verify.  We should probably add that to the README.md as well.  Would you like to include that in this PR or a subsequent PR?

lalex: Done. I've added some lines in README.md.

bobstrecansky (approved):

brettmc:
@lalex  - it looks like we're still waiting on CLA signing before this can be merged.

bobstrecansky:
He has definitely signed the CLA before; wonder if easyCLA is being annoying.

PR #509, opened 2021-12-09
Logging via Logger aware trait

brettmc:
As an alternative approach to singleton logger, this adds logging capability to the library by using the existing `LoggerAwareTrait`.
- adds `LoggerAwareTrait` and `LoggerAwareInterface` to some classes that might/should do logging
- extends LoggerAwareTrait to be able to inject its logger into other classes, makes `setLogger()` chainable for convenience, methods for `error`/`warning`/`debug`/`info` logging
- extends existing factories so that they inject their logger into instances they create

### DIFF
--- a/src/Contrib/OtlpGrpc/Exporter.php
+++ b/src/Contrib/OtlpGrpc/Exporter.php
@@ -15,8 +15,9 @@ use OpenTelemetry\SDK\Trace\Behavior\LoggerAwareTrait;
 use OpenTelemetry\SDK\Trace\Behavior\SpanExporterTrait;
 use OpenTelemetry\SDK\Trace\Behavior\UsesSpanConverterTrait;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+use Psr\Log\LoggerAwareInterface;
 
-class Exporter implements SpanExporterInterface
+class Exporter implements SpanExporterInterface, LoggerAwareInterface
 {
     use LoggerAwareTrait;
     use EnvironmentVariablesTrait;
@@ -111,7 +112,10 @@ class Exporter implements SpanExporterInterface
             return self::STATUS_SUCCESS;
         }
 
-        $this->log('Error exporting span', ['error' => $status->details], \Psr\Log\LogLevel::ERROR);
+        $this->error('Error exporting span', [
+            'error' => $status->details ?? 'unknown grpc error',
+            'code' => $status->code,
+        ]);
         if (in_array($status->code, [
             \Grpc\STATUS_CANCELLED,
             \Grpc\STATUS_DEADLINE_EXCEEDED,
@@ -126,7 +130,6 @@ class Exporter implements SpanExporterInterface
             return self::STATUS_FAILED_RETRYABLE;
         }
 
-        
         return self::STATUS_FAILED_NOT_RETRYABLE;
     }
 
--- a/src/Contrib/OtlpGrpc/Exporter.php
+++ b/src/Contrib/OtlpGrpc/Exporter.php
@@ -11,12 +11,14 @@ use OpenTelemetry\Contrib\Otlp\SpanConverter;
 use Opentelemetry\Proto\Collector\Trace\V1\ExportTraceServiceRequest;
 use Opentelemetry\Proto\Collector\Trace\V1\TraceServiceClient;
 use OpenTelemetry\SDK\EnvironmentVariablesTrait;
+use OpenTelemetry\SDK\Trace\Behavior\LoggerAwareTrait;
 use OpenTelemetry\SDK\Trace\Behavior\SpanExporterTrait;
 use OpenTelemetry\SDK\Trace\Behavior\UsesSpanConverterTrait;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
 
 class Exporter implements SpanExporterInterface
 {
+    use LoggerAwareTrait;
     use EnvironmentVariablesTrait;
     use SpanExporterTrait;
     use UsesSpanConverterTrait;
@@ -109,6 +111,7 @@ class Exporter implements SpanExporterInterface
             return self::STATUS_SUCCESS;
         }
 
+        $this->log('Error exporting span', ['error' => $status->details], \Psr\Log\LogLevel::ERROR);
         if (in_array($status->code, [
             \Grpc\STATUS_CANCELLED,
             \Grpc\STATUS_DEADLINE_EXCEEDED,
@@ -123,6 +126,7 @@ class Exporter implements SpanExporterInterface
             return self::STATUS_FAILED_RETRYABLE;
         }
 
+        
         return self::STATUS_FAILED_NOT_RETRYABLE;
     }
 
--- a/src/Contrib/OtlpGrpc/Exporter.php
+++ b/src/Contrib/OtlpGrpc/Exporter.php
@@ -11,7 +11,6 @@ use OpenTelemetry\Contrib\Otlp\SpanConverter;
 use Opentelemetry\Proto\Collector\Trace\V1\ExportTraceServiceRequest;
 use Opentelemetry\Proto\Collector\Trace\V1\TraceServiceClient;
 use OpenTelemetry\SDK\EnvironmentVariablesTrait;
-use OpenTelemetry\SDK\Trace\Behavior\LoggerAwareTrait;
 use OpenTelemetry\SDK\Trace\Behavior\SpanExporterTrait;
 use OpenTelemetry\SDK\Trace\Behavior\UsesSpanConverterTrait;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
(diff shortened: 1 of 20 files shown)


src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -11,12 +11,15 @@
...
+use OpenTelemetry\SDK\Trace\Behavior\LoggerAwareTrait;
 use OpenTelemetry\SDK\Trace\Behavior\SpanExporterTrait;
 use OpenTelemetry\SDK\Trace\Behavior\UsesSpanConverterTrait;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+use Psr\Log\LoggerAwareInterface;
 
-class Exporter implements SpanExporterInterface
+class Exporter implements SpanExporterInterface, LoggerAwareInterface
 {
+    use LoggerAwareTrait;
```
tidal, line 22: If you move the `LoggerAwareTrait` to the `SpanExporterTrait`, you don't have to add it here again.

src/Contrib/OtlpHttp/Exporter.php
```diff
@@ -12,18 +12,21 @@
...
 use Psr\Http\Message\StreamFactoryInterface;
+use Psr\Log\LoggerAwareInterface;
 
-class Exporter implements SpanExporterInterface
+class Exporter implements SpanExporterInterface, LoggerAwareInterface
 {
     use EnvironmentVariablesTrait;
     use UsesSpanConverterTrait;
     use HttpSpanExporterTrait;
+    use LoggerAwareTrait;
```
tidal, line 29: see above

src/SDK/Trace/Behavior/HttpSpanExporterTrait.php
```diff
@@ -17,13 +17,12 @@
...
 use Throwable;
 
-//use Throwable;
-
 /**
  * @phan-file-suppress PhanTypeInvalidThrowsIsInterface
  */
 trait HttpSpanExporterTrait
 {
+    use LoggerAwareTrait;
```
tidal, line 25: see above

src/SDK/Trace/Behavior/HttpSpanExporterTrait.php
```diff
@@ -46,14 +45,20 @@ protected function doExport(iterable $spans): int
         try {
             $response = $this->dispatchSpans($spans);
         } catch (ClientExceptionInterface $e) {
+            $this->error('Error exporting span', ['error' => $e->getMessage()]);
```
tidal, line 48: Why don't you add the stack trace here? It has been generated anyway.

src/SDK/Trace/Behavior/HttpSpanExporterTrait.php
```diff
@@ -46,14 +45,20 @@ protected function doExport(iterable $spans): int
...
+
             return $e instanceof RequestExceptionInterface
                 ? SpanExporterInterface::STATUS_FAILED_NOT_RETRYABLE
                 : SpanExporterInterface::STATUS_FAILED_RETRYABLE;
         } catch (Throwable $e) {
+            $this->error('Error exporting span', ['error' => $e->getMessage()]);
```
tidal, line 54: I don't think `STATUS_FAILED_NOT_RETRYABLE` and `STATUS_FAILED_RETRYABLE`should have the same severity level.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -23,20 +27,65 @@ public function setDefaultLogLevel(string $logLevel): void
         $this->defaultLogLevel = $logLevel;
     }
 
+    /**
+     * @return static
+     */
+    public function withLogger(LoggerInterface $logger)
```
tidal, line 33: What's the purpose of this method? If you want to have a fluid interface you can simply override the `setLogger`method.

```php

    public function setLogger(LoggerInterface $logger): self
    {
        $this->logger = $logger;

        return $this;
    }
```

(I don't like fluid interface anyway, because it's semantically incorrect, but I know most people like it.)

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -23,20 +27,65 @@ public function setDefaultLogLevel(string $logLevel): void
...
+
     /**
      * @param string $message
      * @param array $context
      * @param string|null $level
      */
     protected function log(string $message, array $context = [], ?string $level = null): void
     {
+        //TODO basic info on which class created the message, cheaper than calculating from a stack trace
+        $context['source'] = __CLASS__;
```
tidal, line 48: It's safer to use ``` $context['source'] = get_class($this); ``` here, since when for whatever reason a class is extended, ``` __CLASS__``` in a Trait will point to the class importing the trait.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -23,20 +27,65 @@ public function setDefaultLogLevel(string $logLevel): void
...
         $this->getLogger()->log(
             $level ?? $this->defaultLogLevel,
             $message,
             $context
         );
     }
 
+    protected function debug(string $message, array $context = []): void
```
tidal, line 56: Could you rename these methods with a `log` prefix (eg. `logDebug`), as these names have no meaning outside of logging.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -23,20 +27,65 @@ public function setDefaultLogLevel(string $logLevel): void
...
+    {
+        $this->log($message, $context, LogLevel::WARNING);
+    }
+
+    protected function error(string $message, array $context = []): void
+    {
+        $this->log($message, $context, LogLevel::ERROR);
+    }
+
+    protected function info(string $message, array $context = []): void
```
tidal, line 71: I think there are a some [severity levels](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Logs/PsrSeverityMapperInterface.php) missing.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
```diff
@@ -23,20 +27,65 @@ public function setDefaultLogLevel(string $logLevel): void
...
+    {
+        $this->log($message, $context, LogLevel::INFO);
+    }
+
+    /**
+     * Inject the logger into another class that implements LoggerAwareInterface, and
+     * return the instance.
+     */
+    protected function injectLogger($instance)
```
tidal, line 80: Can you please add `object` type hints to the parameter and return value?

src/SDK/Trace/Behavior/LoggerAwareTrait.php
brettmc: I added NOTICE, but I feel like critical and higher should not be needed...? (or, can be added if we do have a use-case)

src/SDK/Trace/Behavior/LoggerAwareTrait.php
brettmc: NB required some psalm-suppress annotations

src/SDK/Trace/Behavior/LoggerAwareTrait.php
brettmc: I've overridden `setLogger` for now (phan + psalm were unhappy). Whether the fluid interface gets to stay is up for debate, it just makes setting up via sdk a little easier.

src/SDK/Trace/Behavior/HttpSpanExporterTrait.php
brettmc: done. I wasn't sure what the performance penalty might be for sending the entire exception.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
brettmc: (I'm not sure that the fluid interface aids readability much, but it does allow you to still create a tracer in one go)

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: Ok.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: What does "NB" mean?

src/SDK/Trace/Behavior/HttpSpanExporterTrait.php
tidal: The biggest performance penalty is generating the stack trace in the first place, however some "log inspection tools" will provide additional features, if the stack trace is provided.

src/SDK/Trace/Behavior/LoggerAwareTrait.php
tidal: Fluid interface is already used in many places in the library. Just because I don't like it, doesn't mean I need to have everything my way ;)

brettmc:
discussed in SIG today, this will be closed in favour of #495

PR #512, opened 2021-12-13
build base image for multiple php versions

brettmc:
Allow building for multiple php versions by making Dockerfile php version an argument, and relaxing extension location (by using find instead of hard-coded location).
Updating the github workflow to build and tag base image for 7.4,8.0,8.1 (thus dropping "latest" from further updates).
Updating Makefile to accept `PHP_VERSION` for local testing against other versions.
Relax a couple of dev dependencies so they can install for 7.4->8.1.
Adding xdebug example config to .env.dist, and allow configurating xdebug via environment.
Adding a couple of whitespace-related php-cs-fixer rules.


.github/workflows/publish-otel-php-base-docker-image.yml
```diff
@@ -15,16 +18,22 @@ jobs:
...
-      - name: Push to ghcr.io
+      - name: Build and push ${{ matrix.php-version }} to ghcr.io
         uses: docker/build-push-action@v2
         with:
           push: true
           file: docker/Dockerfile
-          tags: ghcr.io/open-telemetry/opentelemetry-php/opentelemetry-php-base:latest
+          build-args: PHP_VERSION=${{ matrix.php-version }}
+          tags: ghcr.io/open-telemetry/opentelemetry-php/opentelemetry-php-base:${{ matrix.php-version }}
+          cache-from: type=gha,scope=${{ matrix.php-version }}
```
brettmc, line 38: Using github action cache - this speeds up builds substantially (grpc!) but as a side-effect we might not get updated dependencies. Is it worth the trade-off?

.env.dist
```diff
@@ -1 +1,3 @@
 PHP_USER=php
+XDEBUG_MODE=debug
+XDEBUG_CONFIG=client_host=172.17.0.1 idekey=netbeans-xdebug
```
tidal, line 3: Could you please remove this line?

README.md
```diff
@@ -141,6 +141,16 @@ from your bash compatible shell.  This does the following things:
 * Runs all of our [phpunit](https://phpunit.de/) unit tests.
 * Performs static analysis with [Phan](https://github.com/phan/phan), [Psalm](https://psalm.dev/) and [PHPStan](https://phpstan.org/user-guide/getting-started)
 
+### Other PHP versions
+
+We current try to support PHP 7.4, 8.0 and 8.1 - the developer image `ghcr.io/open-telemetry/opentelemetry-php/opentelemetry-php-base` is tagged as `7.4`, `8.0` and `8.1` respectively, with `7.4` being the default.
+You can simulate the important github actions against other PHP versions by running the following command:
```
tidal, line 147: Could you phrase this differently? The github actions are (luckily) not important at all, they merely excute our QA checks, which could be done by any other CI system.

.github/workflows/publish-otel-php-base-docker-image.yml
tidal: It Looks like there is no way to configure a TTL or invalidate the cache. Maybe add some extra key to the cache-key to allow for invalidating when needed?

tidal (changes requested):

docker/Dockerfile
```diff
@@ -14,15 +14,15 @@ RUN chmod +x /usr/local/bin/install-php-extensions; \
         @composer \
     ; \
     # strip debug symbols from extensions to reduce size
-    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/grpc.so; \
-    strip --strip-debug /usr/local/lib/php/extensions/no-debug-non-zts-20190902/xdebug.so;
+    find /usr/local/lib/php/extensions -name grpc.so -exec strip --strip-debug {} \;; \
+    find /usr/local/lib/php/extensions -name xdebug.so -exec strip --strip-debug {} \;;
```
tidal, line 18: +1 for using find, that's what I wanted to use in the first place, but I guess I messed smth up, that's why it did not work.
However if we go for a more abstracted way, we could also simply strip the debug signals in all extensions, so any .so file in the extension directory. That should not hurt if there is nothing to strip, but can probably lower the size of other extensions as well (maybe we need others in the future).

.github/workflows/publish-otel-php-base-docker-image.yml
brettmc: The easiest way I can see is to add an extra variable to the key, that we can change (eg CACHE_VERSION) - but, that just sounds like troubleshooting and extra work for maintainers.
Given that the build takes around 20 minutes, and only runs against `main`, I think we will have less surprises in the future without cache.

tidal: Yap, I agree. Actually the image build doesn't even need to be run after each merge, since it's not using the PHP code anyway, so once a week should be enough, no?

README.md
```diff
@@ -141,6 +141,16 @@ from your bash compatible shell.  This does the following things:
 * Runs all of our [phpunit](https://phpunit.de/) unit tests.
 * Performs static analysis with [Phan](https://github.com/phan/phan), [Psalm](https://psalm.dev/) and [PHPStan](https://phpstan.org/user-guide/getting-started)
 
+### Other PHP versions
+
+We aim to support PHP 7.4, 8.0 and 8.1 - the developer image `ghcr.io/open-telemetry/opentelemetry-php/opentelemetry-php-base` is tagged as `7.4`, `8.0` and `8.1` respectively, with `7.4` being the default.
```
brettmc, line 146: "support officially supported PHP versions" per https://www.php.net/supported-versions.php ??

.github/workflows/publish-otel-php-base-docker-image.yml
brettmc: Ooh, that's a good idea. It could run to schedule, and we have the option of manually running it as required.

README.md
tidal: Yap, however it would be nice if you replace "per" with "according to". I'm not trying to nit-pick, but despite the loanword "per" being used like that in English, this is not what the Latin word "per" means, so using it in that way often creates confusion for non native English speakers.

tidal (approved):

PR #514, opened 2021-12-16
adding php8.1 as an experimental build

brettmc:


bobstrecansky (approved):

brettmc:
This is very close but not quite right - the [workflow is successful](https://github.com/open-telemetry/opentelemetry-php/actions/runs/1585531311) (with one allowed failure), but the checks seem to have their own opinion on whether that is ok.

PR #517, opened 2021-12-18
fixing fiber test crash

brettmc:
instead of serializing FFI into a const, stash it in a static as suggested by @Nevay 
in passing, fixing a deprecation notice from context propagation test.


Nevay:
The problem is caused by something outside of the fiber related code that tries to serialize all defined constants, the approach with the static property only works around this problem.
I currently cannot run the tests to verify, but the root cause seem to be tests with `@runInSeparateProcess`; adding a [`@preserveGlobalState disabled`](https://phpunit.readthedocs.io/en/9.5/annotations.html#preserveglobalstate) annotation to the `@runInSeparateProcess` annotated tests in [TracerProviderTest](https://github.com/open-telemetry/opentelemetry-php/blob/c464706ff42e01cf4f8b63d5fcadce0fe1f60dd9/tests/SDK/Unit/Trace/TracerProviderTest.php) should resolve this issue without changing the usage of a constant as reference holder.

src/Context/fiber/zend_observer_fiber.php
```diff
@@ -30,8 +29,16 @@
             return false;
         }
 
-        // Has to keep reference alive
-        define(__NAMESPACE__ . '\\_ffi_fibers', $fibers);
+        // @internal
+        class FFIFiberHolder
```
tidal, line 33: - Can you please move `class FFIFiberHolder` into a dedicated file?
- Can you please protect this feature with a constant and/or env var (at least until we can functional test this in different runtimes, with different opcache settings, etc.) ?
- Can you please remove the try-catch-block around `FFI::scope`and check for `FFIFiberHolder::$fibers` instead? The chances that users preload an undocumented .h file from a package is very low anyway, which means the exception is almost always thrown.

tests/Context/Unit/Propagation/ArrayAccessGetterSetterTest.php
```diff
@@ -80,8 +80,12 @@ public function offsetExists($offset): bool
                 return true;
             }
 
-            public function offsetGet($offset)
+            /**
+             * @psalm-suppress ReservedWord
+             */
+            public function offsetGet($offset): mixed
```
tidal, line 86: This might work, because the method is actually not called, still `mixed` is not a valid return type in PHP7.4, so can you please remove the stub methods by extending the class from `ArrayObject`, I guess this should work as well.
(Not sure what that interface is for anyway, since it's not used anywhere)

tidal (changes requested):

src/Context/fiber/zend_observer_fiber.php
brettmc: I've moved this code into a class, and protected it behind an env var. also removed try/catch. It turns out that setting up fibers happens much earlier than I believed (I think because it's defined in composer.json, at least I know how it gets pre-loaded now), and that's why I've set the env var in phpunit.xml - setting it in a test, or in the `--ENV--` section of a phpt is too late.

tests/Context/Unit/Propagation/ArrayAccessGetterSetterTest.php
brettmc: replaced with a mock

brettmc:
@Nevay - you were right, phpunit was serializing global state and your suggested fix worked. I hope the changes I've made are an improvement.

src/Context/ZendObserverFiber.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        if (null === self::$fibers) {
+            try {
+                $fibers = FFI::load(__DIR__ . '/fiber/zend_observer_fiber.h');
+                $fibers->zend_observer_fiber_init_register(fn (int $initializing) => Context::storage()->fork($initializing)); //@phpstan-ignore-line
+                $fibers->zend_observer_fiber_switch_register(fn (int $from, int $to) => Context::storage()->switch($to)); //@phpstan-ignore-line
+                $fibers->zend_observer_fiber_destroy_register(fn (int $destroying) => Context::storage()->destroy($destroying)); //@phpstan-ignore-line
+                self::$fibers = $fibers;
+            } catch (Exception $e) {
+                $this->logWarning('Error setting up FFI fiber observer', ['error' => $e]);
```
Nevay, line 48: Logging a warning on startup is likely too extreme (code might not use fibers at all), the `FiberNotSupportedContextStorage` already triggers warnings on global `Context` usage within a fiber.

src/Context/ZendObserverFiber.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * @psalm-suppress UndefinedClass
+     */
+    public function init(): bool
+    {
+        if (null === self::$fibers) {
+            try {
+                $fibers = FFI::load(__DIR__ . '/fiber/zend_observer_fiber.h');
```
Nevay, line 42: Removing `FFI::scope()` breaks this for php-fpm etc. with the default FFI configuration ([ffi.enable=preload](https://www.php.net/manual/en/ffi.configuration.php#ini.ffi.enable)).

README.md
```diff
@@ -77,6 +77,11 @@ In order to use compression in HTTP requests you should install
 the  PHP zlib extension. You can use the same install methods as described for the gRPC extension above,
 however most OS` package managers provide a package for the extension.
 
+#### 4.) Install PHP [ext-ffi](https://www.php.net/manual/en/book.ffi.php)
+
+_Experimental_ support for using fibers in PHP 8.1 for Context storage requires the `ffi` extension, and can
+be enabled by setting the `OTEL_PHP_FIBERS_ENABLED` environment variable to a truthy value (`1`, `true`, `on`).
```
Nevay, line 83: Should add a comment about preloading for SAPIs besides CLI - to use the fiber support under different SAPIs the binding may have to be preloaded, e.g. by including `vendor/autoload.php` in `opcache.preload`.
(Should consider moving the header file to the toplevel directory if we want to document the usage with `ffi.preload`, but not sure whether that is something we should support/document for now.)

Nevay (changes requested):

src/Context/ZendObserverFiber.php
brettmc: removed

src/Context/ZendObserverFiber.php
brettmc: ok, put this back to how it was try(scope)/try(load).

README.md
brettmc: I haven't considered that these changes working with other SAPIs. I've updated the readme but will try to add a working example with FPM to verify what actually works.

tidal (approved):

bobstrecansky (approved):

PR #521, opened 2021-12-24
Restructure API / SDK components (Attributes, Clock, Trace)

Nevay:
- moves `Attributes` and `Clock` from SDK/Trace to SDK, needed by metrics too
- moves `Event` and `Link` from API to SDK, only needed for readable spans, not used by API
- moves `NoopTracer` from SDK to API as required by specification
- moves `AttributesInterface` from API to SDK by widening types in API to iterable
  previous implementation required API consumers to depend on the SDK implementation to provide attributes to API methods
  bc-breaks:
  - `Attribute` abstraction is removed, `Attributes` now return only the value instead of {name,value} pairs
  - `iterable $attributes` arguments are now non-nullable
  - (todo: rename `Attributes::setAttribute()` to `Attributes::set()` to be consistent with `Attributes::get()` ?)
- moves `ClockInterface` from API to SDK
  currently only used in SDK, interface+implementation should either be fully available in API or just in SDK


src/SDK/Trace/ImmutableSpan.php
```diff
@@ -131,7 +132,7 @@ public function getEvents(): array
         return $this->events;
     }
 
-    public function getAttributes(): API\AttributesInterface
+    public function getAttributes(): AttributesInterface
```
brettmc, line 135: I'm a little confused about the mixed usage of `iterable` and `AttributesInterface` for attributes in different parts of the code. I just want to make sure that's intentional?

tests/SDK/Util/SpanData.php
```diff
@@ -10,20 +10,23 @@
...
+    /** @var list<EventInterface> */
     private array $events = [];
 
-    /** @var list<API\LinkInterface> */
+    /** @var list<LinkInterface>
+     */
     private array $links = [];
 
-    private API\AttributesInterface $attributes;
+    private \OpenTelemetry\SDK\AttributesInterface $attributes;
```
brettmc, line 29: I think the project style is to have these as `use` statements

src/SDK/Trace/ImmutableSpan.php
Nevay: Its intentional, API uses `iterable` (only parameters), SDK uses `AttributesInterface`.

tests/SDK/Util/SpanData.php
Nevay: Missed this one, PhpStorm changed some use-statements to fqcn on moving these classes.

brettmc:
I've taken a first pass over this - I like! Couple of comments to be looked at, and I'm in favour of renaming `setAttribute` to `set` - we're still pre-alpha so breaking things for the better is fine.

tidal (changes requested):

tidal (approved):

tidal:
Thanks for your changes (sorry for the delay, I was busy over the holidays).
There are some merge conflics now, otherwise It's looking good imo.

brettmc:
Hi @Nevay - are you happy for this to be merged now?

Nevay:
@brettmc yes, this should be finished now.
I kept `Attributes::setAttribute()` unchanged for now - method will be removed when we split `Attributes` into mutable builder and immutable attributes to ensure that readonly objects (e.g. `SpanDataInterface`) are not indirectly mutable.

PR #524, opened 2021-12-25
Add FFI Fiber support preload example

### LINKED ISSUE #523: create FFI + fiber example
Create a working example for FFI preloading of fiber code. eg when used with php-fpm.

> Should add a comment about preloading for SAPIs besides CLI - to use the fiber support under different SAPIs the binding may have to be preloaded, e.g. by including `vendor/autoload.php` in `opcache.preload`.
(Should consider moving the header file to the toplevel directory if we want to document the usage with `ffi.preload`, but not sure whether that is something we should support/document for now.)

_Originally posted by @Nevay in https://github.com/open-telemetry/opentelemetry-php/pull/517#discussion_r771930476_

Nevay:
Adds php-fpm example for FFI Fiber support using `php-fpm -d opcache.preload=vendor/autoload.php`.
(Tried to remove duplication between this example and the prometheus example.)

Resolves #523.


bobstrecansky (approved):

bobstrecansky:
Thanks for this!  Glad to have these updates.

Issue #519, opened 2021-12-22
Move PrometheusExporter into Contrib package


tidal:
The PrometheusExporter is currently located inside of the SDK package/directory.
It should live inside of the Contrib package/directory.

Issue #507, opened 2021-12-09
Create separate composer packages for contrib repository and SDK components.


tidal:
_This topic has been discussed in different threads so this an issue for this task to be referenced._

According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md#opentelemetry-client-generic-design) the OpenTelemetry client should be distributed as (at least) 4 packages.
The individual components of the [contrib repo](https://github.com/open-telemetry/opentelemetry-php-contrib) should be distributed as separate packages as well.

Composer as the package manager does not allow for several packages to live in the same git repository. So the usual way to deal with this in Mono-Repos is to create Many-Repos mirrors, which can be referenced by composer.

The goal is to setup the infrastructure and tools to allow the package split and make it as automated as possible.
This does not only include technical tasks, but also dealing with restrictions by OpenTelemetry and the access to repository configuration by mainatainers, as well as yet to be solved dependency conflicts in the main library.

tidal:
All issues in https://github.com/open-telemetry/opentelemetry-php/issues/464 have to be solved to allow splitting the main PHP libary.

tidal:
See discussions in: https://github.com/open-telemetry/opentelemetry-php/issues/220 , https://github.com/open-telemetry/opentelemetry-php-contrib/issues/21

tidal:
related: https://github.com/open-telemetry/opentelemetry-php/issues/356

tidal:
@stale
Kennstedenübadenrobota?

PR #531, opened 2021-12-30
Initial package-split setup

tidal:
This PR adds an initial setup for splitting the library into separate composer packages. (see: https://github.com/open-telemetry/opentelemetry-php/issues/507)

- Closes https://github.com/open-telemetry/opentelemetry-php/issues/519
- Adds composer.json files for the individual packages.
- Adds dependency checks for 3rd party libraries.
- Moves OTLP protobuf files into `proto/otel` directory.
- Resolves a naming confilct in `make` by renaming `make proto` to `make protobuf`
- Changes generating of protobuf files to use latest stable release of source repo and publish version in a `VERSION` file.


proto/otel/composer.json
```diff
@@ -0,0 +1,27 @@
...
+        "google/protobuf": "^v3.3.0"
+    },
+    "autoload": {
+        "psr-4": {
+            "Opentelemetry\\Proto\\": "Opentelemetry/Proto/",
+            "GPBMetadata\\Opentelemetry\\": "GPBMetadata/Opentelemetry/"
+        }
+    },
+    "suggest": {
+        "ext-protobuf": "For better performance, when dealing with protobuf file",
```
brettmc, line 24: file doesn't seem like the right word.... "format"?

tests/Contrib/Unit/PrometheusExporterTest.php
```diff
@@ -15,7 +14,7 @@ class PrometheusExporterTest extends TestCase
 {
     public function testEmptyMetricsExportReturnsSuccess()
     {
-        $exporter = new PrometheusExporter($this->createMock(CollectorRegistry::class));
+        $exporter = new \OpenTelemetry\Contrib\Prometheus\PrometheusExporter($this->createMock(CollectorRegistry::class));
```
brettmc, line 17: is this an auto-generated change? I think it should have a "use"...

composer.json
```diff
@@ -35,11 +38,19 @@
             "email": "levim@php.net"
```
brettmc, line 38: is this a good time to generalise to otel php contributors ?

tests/Contrib/Unit/PrometheusExporterTest.php
tidal: Yap, it's probably caused by auto-refactoring. I'll fix it.

brettmc (approved):
LGTM - a couple of minor issues and questions, but otherwise great.

composer.json
tidal: Don't know, what do you excatly mean? Using the same entry as in the other composer files?

brettmc: I was thinking the same as your recent addition to the sub-packages - but it might need more discussion so as not to offend anybody

tidal: Yeah, that was the reason I did not change anything in the main composer file. Let's just postpone this decision. It's such a minor change. We can throw it in any time, but I guess it would be better to discuss it upfront.

tidal:
Thanks for the swift review @brettmc! <3

PR #532, opened 2022-01-02
Allow psr/log ^2.0 and ^3.0

tidal:
In PHP >= 8 psr/log versions ^2.0 or ^3.0 are loaded by default (stronger typed than version1).
Allowing these packages to be loaded allows opentelemetry-php to be added without having to downgrade psr/log.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "php-http/discovery": "^1.14",
         "promphp/prometheus_client_php": "^2.2.1",
         "psr/http-factory-implementation": "^1.0",
-        "psr/log": "^1.1",
+        "psr/log": "^1.1|^2.0|^3.0",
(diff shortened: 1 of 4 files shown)


brettmc (approved):

tidal:
Thanks @brettmc !
Kind of strange, why this [reported decreased code-coverage.](https://app.codecov.io/gh/open-telemetry/opentelemetry-php/compare/532/changes). This change should have zero side-effects on tests.
Maybe it's flaky because of what PHP version reports the coverage (first/last)?
At least something to keep an eye on in the future.

brettmc:
I thought that was strange, too. I looked into it, and the coverage changes seemed unrelated to this change - I guessed that it's used the wrong baseline, somehow (I'm assuming that it stores the latest coverage data for each branch)

PR #537, opened 2022-01-04
Move resource detection out of `ResourceInfo`

### LINKED ISSUE #473: `ResourceInfo` needs a dedicated factory
I believe `ResoerceInfo` needs a dedicated factory. It's overloaded with factory methods, while creating objects is not its main responsibility. However this could be changed in a later PR.

_Originally posted by @tidal in https://github.com/open-telemetry/opentelemetry-php/pull/469#r748851138_

Nevay:
Moves resource detection out of `ResourceInfo` and adds some additional attributes.

`ResourceInfo::merge()` no longer differs between empty and non-empty values ([the value of the updating resource MUST be picked (even if the updated value is empty)](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/sdk.md#merge)).
`ResourceInfo::create()` no longer adds the default resource attributes.

Fixes #473.

### DIFF
--- a/src/SDK/Resource/Detectors/Composer.php
+++ b/src/SDK/Resource/Detectors/Composer.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Resource\Detectors;
+
+use function class_exists;
+use Composer\InstalledVersions;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+
+final class Composer implements ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo
+    {
+        if (!class_exists(InstalledVersions::class)) {
+            return ResourceInfo::emptyResource();
+        }
+
+        $attributes = [
+            ResourceAttributes::SERVICE_NAME => InstalledVersions::getRootPackage()['name'],
+            ResourceAttributes::SERVICE_VERSION => InstalledVersions::getRootPackage()['pretty_version'],
+        ];
+
+        return ResourceInfo::create(new Attributes($attributes), ResourceAttributes::SCHEMA_URL);
+    }
+}
--- a/src/SDK/Resource/Detectors/Composite.php
+++ b/src/SDK/Resource/Detectors/Composite.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Resource\Detectors;
+
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+
+final class Composite implements ResourceDetectorInterface
+{
+    private iterable $resourceDetectors;
+
+    /**
+     * @param iterable<ResourceDetectorInterface> $resourceDetectors
+     */
+    public function __construct(iterable $resourceDetectors)
+    {
+        $this->resourceDetectors = $resourceDetectors;
+    }
+
+    public function getResource(): ResourceInfo
+    {
+        $resources = [];
+        foreach ($this->resourceDetectors as $resourceDetector) {
+            $resources[] = $resourceDetector->getResource();
+        }
+
+        return ResourceInfo::merge(...$resources);
+    }
+}
--- a/src/SDK/Resource/Detectors/Env.php
+++ b/src/SDK/Resource/Detectors/Env.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Resource\Detectors;
+
+use function explode;
+use function getenv;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+use function strpos;
+use function trim;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable
+ */
+final class Env implements ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo
+    {
+        $attributes = [];
+
+        $string = getenv('OTEL_RESOURCE_ATTRIBUTES');
+        if ($string && false !== strpos($string, '=')) {
+            foreach (explode(',', $string) as $pair) {
+                [$key, $value] = explode('=', $pair);
+                $attributes[trim($key)] = trim($value);
+            }
+        }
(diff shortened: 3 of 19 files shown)


src/SDK/Resource/Detectors/Env.php
```diff
@@ -0,0 +1,40 @@
...
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+use function strpos;
+use function trim;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable
+ */
+final class Env implements ResourceDetectorInterface
```
tidal, line 19: Could you rename this class to "Environment"? 
It's not super important, but "Environment" is used in other place as well, and it's a bit more descriptive.

src/SDK/Resource/Detectors/Os.php
```diff
@@ -0,0 +1,32 @@
...
+use OpenTelemetry\SemConv\ResourceAttributes;
+use const PHP_OS;
+use const PHP_OS_FAMILY;
+use function php_uname;
+use function strtolower;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/semantic_conventions/os.md
+ */
+final class Os implements ResourceDetectorInterface
```
tidal, line 19: Could you rename this class to "OperatingSystem"? 
Again, not super important.

src/SDK/Resource/Detectors/ProcessRuntime.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/semantic_conventions/process.md#process-runtimes
+ */
+final class ProcessRuntime implements ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo
+    {
+        $attributes = [
+            ResourceAttributes::PROCESS_RUNTIME_NAME => php_sapi_name(),
```
tidal, line 22: Could we also support [php-runtime/runtime](https://github.com/php-runtime/runtime)?
The term "runtime" is very ambigous in general, however in a lot of cases which could be considered a "runtime" ´php_sapi_name()` just returns `cli`.

src/SDK/Resource/Detectors/ProcessRuntime.php
```diff
@@ -0,0 +1,28 @@
...
+            ResourceAttributes::PROCESS_RUNTIME_VERSION => PHP_VERSION,
```
tidal, line 23: Maybe just add this attribute when it's CLI SAPI or we can get the version of other runtimes in another way??

src/SDK/Resource/Detectors/Sdk.php
```diff
@@ -0,0 +1,48 @@
...
+    ];
+
+    public function getResource(): ResourceInfo
+    {
+        $attributes = [
+            ResourceAttributes::TELEMETRY_SDK_NAME => 'opentelemetry',
+            ResourceAttributes::TELEMETRY_SDK_LANGUAGE => 'php',
+        ];
+
+        if (class_exists(InstalledVersions::class)) {
```
tidal, line 31: Same as above (first comment)

src/SDK/Resource/ResourceDetectorInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Resource;
+
+interface ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo;
+}
```
tidal, line 10: Just for reference.
We will have to change the AWS detectors accordingly: https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/Aws

tidal (approved):

tidal:
Thank you, this is looking good.
I just have some nit-picky stuff and some possible enhancements, but feel freeto ignore them for now.

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -0,0 +1,29 @@
...
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+
+final class Composer implements ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo
+    {
+        if (!class_exists(InstalledVersions::class)) {
```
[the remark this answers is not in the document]
Nevay: I would prefer to keep this as is, considering that composer 1.x is [deprecated](https://github.com/composer/composer/issues/10340). Users can add their own detector if they have to use an old versions (or just specify these attributes manually).

src/SDK/Resource/Detectors/Composite.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public function getResource(): ResourceInfo
+    {
+        $resources = [];
+        foreach ($this->resourceDetectors as $resourceDetector) {
+            $resources[] = $resourceDetector->getResource();
+        }
+
+        return ResourceInfo::merge(...$resources);
```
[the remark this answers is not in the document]
Nevay: Resource detectors should be called only once during startup of the SDK, caching here is likely not useful.

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -0,0 +1,29 @@
...
+namespace OpenTelemetry\SDK\Resource\Detectors;
+
+use function class_exists;
+use Composer\InstalledVersions;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+
+final class Composer implements ResourceDetectorInterface
```
brettmc, line 14: This class doesn't have any test coverage

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
bobstrecansky, line 1: Perhaps we could come up with another naming convention here?  When I read `Composer`, I immediately went to [The PHP dependency manager.](https://getcomposer.org/)

src/SDK/Resource/Detectors/ProcessRuntime.php
Nevay: Is it possible to determine the runtime / runner without checking the stack trace for `RunnerInterface::run()`? (Could be unreliable depending on the usage of the runtime component.)

Should be implemented as additional detector to allow opting in/out separately.

bobstrecansky (approved):

bobstrecansky:
I think we just need to fix the merge conflicts now and we're good to go @Nevay

bobstrecansky:
Hey @Nevay - Would you like us to fix these merge conflicts and merge this PR now?

Nevay:
@bobstrecansky rebased again, should be ready to merge now.

PR #540, opened 2022-01-06
adding covers annotations to all tests

brettmc:
- Adding `@covers` to all test classes, and enforcing coverage for all future tests in phpunit config. Integration tests are all marked as `@coversNothing`, and they already do not contribute to code coverage due to recent splitting of unit/integration github actions).
- Adding and updating tests to cover some easier now-uncovered code
- Putting back `test-coverage` make target (which I recently removed but almost immediately regretted)

### DIFF
--- a/tests/Unit/SDK/Metrics/ValueRecorderTest.php
+++ b/tests/Unit/SDK/Metrics/ValueRecorderTest.php
@@ -4,10 +4,10 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Tests\Unit\SDK\Metrics;
 
-use InvalidArgumentException;
 use OpenTelemetry\API\Metrics as API;
 use OpenTelemetry\SDK\Metrics\ValueRecorder;
 use PHPUnit\Framework\TestCase;
+use TypeError;
 
 /**
  * @covers OpenTelemetry\SDK\Metrics\ValueRecorder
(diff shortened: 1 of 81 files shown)


tests/Unit/SDK/Metrics/ValueRecorderTest.php
```diff
@@ -5,149 +5,162 @@
...
-        $metric->record(2);
-        $this->assertEquals(2, $metric->getCount());
-        $this->assertEquals(5, $metric->getMax());
-        $this->assertEquals(2, $metric->getMin());
-        $this->assertEquals(7, $metric->getSum());
+        $this->metric->record(5);
+        $this->assertEquals(1, $this->metric->getCount());
+        $this->assertEquals(5, $this->metric->getMax());
+        $this->assertEquals(5, $this->metric->getMin());
+        $this->assertEquals(5, $this->metric->getSum());
```
tidal, line 35: Could we convert this into a custom assertion?
```php assertMetrics(float $count, float $max, float $min, float $sum, ValueRecorder $recorder)```
Would be better readable.

File: tests/Unit/SDK/SystemClockTest.php [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+class SystemClockTest extends TestCase
+{
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState false
+     */
+    public function test_get_instance(): void
+    {
+        $clock = SystemClock::getInstance();
+        $this->assertSame($clock, SystemClock::getInstance());
```
tidal, line 22: `assertEquals`

tests/Unit/SDK/Metrics/ValueRecorderTest.php
```diff
@@ -5,149 +5,162 @@
...
-        $this->assertEquals(1, $metric->getCount());
-        $this->assertEquals(5.2222, $metric->getMax());
-        $this->assertEquals(5.2222, $metric->getMin());
-        $this->assertEquals(5.2222, $metric->getSum());
-        $metric->record(2.6666);
-        $this->assertEquals(2, $metric->getCount());
-        $this->assertEquals(5.2222, $metric->getMax());
-        $this->assertEquals(2.6666, $metric->getMin());
-        $this->assertEquals(7.8888, $metric->getSum());
+        $this->metric->record(5.2222);
```
tidal, line 87: The tested metthod (`record`) could simply expect an float and all the above tests would not be needed, as integers are floats as well.

tidal (approved):

tidal:
Thank you.
Code-coevrage dropped a lot (quod erat demonstrandum), but now we have a more realistic coverage metric.

tidal:
However I'm not sure we need to tag the integration tests with `coversNothing`. 
They essentially cover code, and it might be interesting to see what code is actually executed.
Executing them separately is enough imo.

tests/Unit/SDK/SystemClockTest.php
brettmc: this is testing the singleton functionality, so ensuring it's the same object that comes back a second time (have updated the test name to be more descriptive)

brettmc:
@tidal - removing the `@coversNothing` will trip up the `forceCoversAnnotation` I added (which will mark any future tests without covers annotations as risky). So, we can have one or the other - forcing coverage annotations is not that important though. WDYT?

tests/Unit/SDK/Metrics/ValueRecorderTest.php
brettmc: done - added a `float` type-hint, and removed some casting and checks that are no longer needed

tests/Unit/SDK/SystemClockTest.php
tidal: Yap, sorry, I was aware of that. I did not mean to post that comment, must have been by accident.

tidal:
> @tidal - removing the `@coversNothing` will trip up the `forceCoversAnnotation` I added (which will mark any future tests without covers annotations as risky). So, we can have one or the other - forcing coverage annotations is not that important though. WDYT?

Ok, got you. I'd rather have the `forceCoversAnnotation`. However we could create a separate phpunit config for the integration tests and call it with `-c|--configuration <file>   Read configuration from XML file`

brettmc:
> Ok, got you. I'd rather have the forceCoversAnnotation. However we could create a separate phpunit config for the integration tests and call it with -c|--configuration <file> Read configuration from XML file

I can see some use in seeing code coverage from integration tests...however maintaining a duplicate phpunit config doesn't feel right. If it helps, the [phpunit docs](https://phpunit.readthedocs.io/en/9.5/code-coverage-analysis.html#code-coverage-analysis-specifying-covered-parts-examples-guestbookintegrationtest-php) recommend not generating coverage for integration tests:

> It is also possible to specify that a test should not cover any method by using the @coversNothing annotation (see @coversNothing). This can be helpful when writing integration tests to make sure you only generate code coverage with unit tests.

I haven't been able to find another way to enforce covers for only some files (there is not a phpunit command-line switch for it, and php-cs-fixer has a fixer for covers, but you can't apply rules only to some paths).

tidal:
> I can see some use in seeing code coverage from integration tests...however maintaining a duplicate phpunit config doesn't feel right. If it helps, the [phpunit docs](https://phpunit.readthedocs.io/en/9.5/code-coverage-analysis.html#code-coverage-analysis-specifying-covered-parts-examples-guestbookintegrationtest-php) recommend not generating coverage for integration tests:

The "coverage report" for the integration test would just have informational value, but it's not super important.
Maybe we find a way to create such a report once the servers are stup.

PR #520, opened 2021-12-23
Add metric aggregation implementation

Nevay:
Adds synchronous and asynchronous metric aggregations with support for multiple readers per aggregation.
Intended as internal implementation of meter instruments; on instrument creation: `Meter` would create a `MetricStream` for each registered `View`, `MetricReader`s would register as reader for each `MetricStream`.

Creating this as draft to gather some feedback before I start working on the todos. Are there objections to implementing the metrics SDK this way?

---

TODOs:
- [ ] remove/resolve compatibility.php
- [ ] add PHP7.4 compatibility (currently ^8.1 only)
- [ ] add tests
- [ ] fix cs / psalm / phpstan / phan

PR #542, opened 2022-01-06
Make `AttributesInterface` immutable

Nevay:
Makes `AttributesInterface` immutable by introducing `AttributesBuilderInterface` as mutable builder.

Related:
https://github.com/open-telemetry/opentelemetry-php/pull/521#issuecomment-1003330123
https://github.com/open-telemetry/opentelemetry-php/pull/520


src/SDK/AttributesBuilder.php
```diff
@@ -0,0 +1,134 @@
...
+    {
+        return isset($this->attributes[$offset]);
+    }
+
+    public function offsetGet($offset)
+    {
+        return $this->attributes[$offset] ?? null;
+    }
+
+    public function offsetSet($offset, $value): void
```
lalex, line 95: An `$offset` which is an Attribute key "must be a non-null and non-empty string" ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/common/common.md#attributes)). It's better to add type hint `string`.

src/SDK/Trace/Span.php
```diff
@@ -361,9 +360,9 @@ public function toSpanData(): SpanDataInterface
             $this->name,
             $this->links,
             $this->events,
-            $this->getImmutableAttributes(),
-            (null === $this->attributes) ? 0 : $this->attributes->getTotalAddedValues(),
-            $this->totalRecordedEvents,
+            $this->attributes->build(),
+            max(0, $this->droppedEventsCount),
```
lalex, line 364: I've spent a while to understand why the count could be negative number. Very interesting approach to store the limit value.

src/SDK/Trace/SpanBuilder.php
```diff
@@ -36,24 +34,39 @@ final class SpanBuilder implements API\SpanBuilderInterface
...
         InstrumentationLibrary $instrumentationLibrary,
-        TracerSharedState $tracerSharedState,
-        SpanLimits $spanLimits
+        ResourceInfo $resource,
+        SamplerInterface $sampler,
+        SpanProcessorInterface $spanProcessor,
+        IdGeneratorInterface $idGenerator,
+        AttributesBuilderInterface $attributes,
+        AttributesFactoryInterface $linkAttributes,
+        AttributesFactoryInterface $eventAttributes,
```
lalex, line 55: It's not clear when Attributes are the Builder and when they are Factory. Probably the proper naming of the arguments and properties could help with it.

src/SDK/Trace/SpanLimits.php
```diff
@@ -25,6 +27,30 @@ final class SpanLimits
 
     private int $attributePerLinkCountLimit;
 
+    public function getSpanAttributes(): AttributesFactoryInterface
```
lalex, line 30: Is it correct to have a creational methods in the value-class?

src/SDK/AttributesBuilder.php
brettmc: Also, needs to be compatible with ArrayAccess signature for 8.1+, see https://github.com/open-telemetry/opentelemetry-php/runs/4729027088?check_suite_focus=true#step:13:45

src/SDK/Trace/Event.php
```diff
@@ -14,11 +13,11 @@ final class Event implements EventInterface
     private int $timestamp;
     private AttributesInterface $attributes;
 
-    public function __construct(string $name, int $timestamp, AttributesInterface $attributes = null)
+    public function __construct(string $name, int $timestamp, AttributesInterface $attributes)
```
brettmc, line 16: The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#add-events) says "zero or more attributes" for an event, so I think they should remain optional?

src/SDK/Trace/Sampler/TraceIdRatioBasedSampler.php
```diff
@@ -60,7 +60,7 @@ public function shouldSample(
         $traceIdCondition = $lowerOrderBytes < round($this->probability * $traceIdLimit);
         $decision = $traceIdCondition ? SamplingResult::RECORD_AND_SAMPLE : SamplingResult::DROP;
 
-        return new SamplingResult($decision, $attributes, $traceState);
+        return new SamplingResult($decision, [], $traceState);
```
brettmc, line 63: why not passing `$attributes` through?

src/SDK/Trace/Span.php
```diff
@@ -66,10 +63,10 @@ final class Span extends API\AbstractSpan implements ReadWriteSpanInterface
     /** @var list<EventInterface> */
     private array $events = [];
 
-    private ?AttributesInterface $attributes;
-    private int $totalRecordedEvents = 0;
+    private AttributesBuilderInterface $attributes;
```
brettmc, line 66: I don't get this, the variable might be better named `$attributesBuilder` ?

src/SDK/Trace/Sampler/TraceIdRatioBasedSampler.php
Nevay: The returned attributes are additional attributes, ["A set of span Attributes that will also be added to the `Span`"](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/trace/sdk.md#shouldsample).

src/SDK/Trace/Event.php
Nevay: Callers can provide an empty `AttributesInterface` instance; allowing `null` to represent empty attributes felt unecessary as the only caller (`Span`) always passes non-null attributes.

bobstrecansky:
Hey @Nevay -  are you still planning on updating this PR?

src/SDK/Trace/Span.php
Nevay: Should I revert this change or is adding a comment to the `droppedLinksCount` and `droppedEventsCount` properties sufficient?

bobstrecansky:
Closing this due to inactivity; please feel free to reopen if you'd like to!

PR #545, opened 2022-01-09
Revamped PR for Jaeger Agent Support

Grunet:
This is @saktib 's #337 PR but updated to work with all the changes to the repo over the past half year, plus a couple of starter tests, as well as fixes for the 3 static analysis tools.

I have NOT yet functional tested this against a real Jaeger agent.

Some other questions of mine

- Is CodecUtility.php still needed? It's unused and I finally sorta see what it was getting at, but I'm not sure it's needed anymore.
- Are the 2 areas of commented out code in the SpanConverter file still needed? The first was there previously, but the 2nd I had to get rid of as being unreachable with the types in play
- Is limiting the entire repo to be uninstallable on non 64-bit systems kosher? Or should this be a check in code so it only applies to those trying to use it with Jaeger?

I'll add the areas of code I think are still untested in a comment below. If there are any that you think need or have to have more test coverage prior to functional testing (which as of rn I'm planning to do after adding thrift/http support in future PRs), lmk!


Grunet:
Here are those remaining uncovered areas, mostly around input validation and error handling from what I could tell:

## AgentExporter

- Input validation in the constructor (the 2 exception throwing lines)

## JaegerTransport

- Line 70
- Line 93 - 96 - error handling

## ThriftUdpTransport

- Lines 28 - 33 - error handling
- Unused methods that need to be defined for the interface (isOpen, open, read)
- Line 63 - error handling
- Lines 82-84 - error handling

composer.json
```diff
@@ -7,7 +7,9 @@
     "license": "Apache-2.0",
     "require": {
         "php": "^7.4 || ^8.0",
+        "php-64bit": "^7.4",
```
brettmc, line 10: I think this replaces the `php` entry above

Grunet: Interesting, kinda weird it let me do that...got rid of that line and updated the existing one to say php-64bit

src/Contrib/Jaeger/CodecUtility.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\Jaeger;
+
+use Throwable;
+
+class CodecUtility
```
brettmc, line 9: A lot of this code exists already (or is very similar) in `SDK\Trace\RandomIdGenerator`, so I think we should use that, and add in any extra functionality from this class.

Grunet: Yeah I couldn't actually find any usages of this class. I think I sorta see what the first method was doing but yeah not so sure about this trace Id generation part

src/Contrib/Jaeger/Transport.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
brettmc, line 1: rename to `TransportInterface`

src/Contrib/Jaeger/JaegerTransport.php
```diff
@@ -0,0 +1,111 @@
...
+    private $process = null;
+    private $maxBufferSize = 0;
+
+    public function __construct($address = '127.0.0.1', $port = 6831, $maxBufferSize = 0)
+    {
+        $this->transport = new ThriftUdpTransport($address, $port);
+        $p = new TCompactProtocol($this->transport);
+        $this->client = new AgentClient($p, $p);
+
+        $this->maxBufferSize = ($maxBufferSize > 0 ? $maxBufferSize : self::DEFAULT_BUFFER_SIZE);
```
brettmc, line 34: this could be eliminated by using this const as the constructor parameter default

Grunet:
Weird that ```make psalm``` seems to be working fine for me locally. And the tests seem to find this OpenTelemetry\Tests\SDK\Util\SpanData class just fine too

src/Contrib/Jaeger/JaegerTransport.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+    * Flush submits the internal buffer to the remote server. It returns the
+    * number of spans flushed.
+    *
+    * @param bool $force - force a flush, even on a partial buffer
+    */
+    public function flush($force = false)
+    {
+        $spans = count($this->buffer);
```
brettmc, line 66: type-hint the return type (int)... (if `flush()` does indeed need to return this data, it's not used, so maybe it should return `void` instead)

src/Contrib/Jaeger/JaegerTransport.php
```diff
@@ -0,0 +1,111 @@
...
+            // flush the UDP data
+            $this->transport->flush();
+
+            // reset the internal buffer
+            $this->buffer = [];
+
+            // reset the process tag
+            $this->process = null;
+        } catch (TTransportException $e) {
+            error_log('jaeger: transport failure: ' . $e->getMessage());
```
brettmc, line 94: there is a trait now to facilitate logging... `LogsMessagesTrait`

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -0,0 +1,168 @@
...
+
+        foreach ($span->getAttributes() as $k => $v) {
+            $tags[$k] = $this->sanitiseTagValue($v->getValue());
+        }
+        $tags = $this->buildTags($tags);
+
+        [
+            'traceIdLow' => $traceIdLow,
+            'traceIdHigh' => $traceIdHigh
+        ] = SpanConverter::convertOtlpToJaegerTraceIds($span->getContext()->getTraceID());
```
brettmc, line 38: can you make this `self::`, or even make it a non-static function and use `$this`

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -0,0 +1,168 @@
...
+        $spanId = intval($span->getContext()->getSpanID(), 16);
+        $parentSpanId = intval($span->getParentSpanId(), 16);
+
+        //TODO - determine if any of the below commented out code is still needed
```
brettmc, line 42: I would think at least events should be, but some real-world testing and comparing with jaeger+zipkin might be a good place to start

tests/Contrib/Unit/JaegerSpanConverterTest.php
```diff
@@ -0,0 +1,61 @@
...
+namespace OpenTelemetry\Tests\Contrib\Unit;
+
+use OpenTelemetry\API\Trace\StatusCode;
+use OpenTelemetry\Contrib\Jaeger\SpanConverter;
+use OpenTelemetry\SDK\Trace\StatusData;
+use OpenTelemetry\Tests\SDK\Util\SpanData;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversDefaultClass SpanConverter
```
brettmc, line 14: a good starting point would be to just use `@covers` here, meaning coverage will only be generated for the target class (without you having to annotate every method)

brettmc (commented):
Looking good, I've added a few comments. Could you also please type-hint return types in methods where you can? (void, int, etc)?

tidal:
@Grunet 
Thank you. Having a Thrift Exporter will also be helpful for the abstraction of all exporter classes.

I don't have time for a in detail review atm, but can you please add `ext-sockets` in the Contrib `composer.json` and as a `suggest` in the main one?

brettmc:
> Weird that `make psalm` seems to be working fine for me locally. And the tests seem to find this OpenTelemetry\Tests\SDK\Util\SpanData class just fine too
> 
> EDIT: Not sure what to make of the psalm specific error either. Is something up with psalm at the moment perhaps if it's getting pulled in on-demand via an action or similar?

I think there is something up with psalm, too. But wrt SpanData complaints, I think they're legit - in any case SpanData is no longer a thing in `main`, so you might need to merge in main again.

composer.json
```diff
@@ -6,8 +6,9 @@
     "readme": "./README.md",
     "license": "Apache-2.0",
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php-64bit": "^7.4 || ^8.0",
```
Grunet, line 9: Per the Slack thread it sounds like this should be a runtime check

Grunet:
> > Weird that `make psalm` seems to be working fine for me locally. And the tests seem to find this OpenTelemetry\Tests\SDK\Util\SpanData class just fine too
> > EDIT: Not sure what to make of the psalm specific error either. Is something up with psalm at the moment perhaps if it's getting pulled in on-demand via an action or similar?
> 
> I think there is something up with psalm, too. But wrt SpanData complaints, I think they're legit - in any case SpanData is no longer a thing in `main`, so you might need to merge in main again.

Yup think I was running stuff locally without pulling those changes. I think I've gotten things fixed up now where the build seems to pass at least

Grunet:
> @Grunet Thank you. Having a Thrift Exporter will also be helpful for the abstraction of all exporter classes.
> 
> I don't have time for a in detail review atm, but can you please add `ext-sockets` in the Contrib `composer.json` and as a `suggest` in the main one?

Done

tests/Contrib/Unit/JaegerSpanConverterTest.php
Grunet: Think I incidentally did this following the warnings when trying to make the build pass

src/Contrib/composer.json
```diff
@@ -13,6 +13,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
+        "ext-sockets": "*",
```
brettmc, line 16: I think it should only be a suggest...if you're not using thrift, you don't need it. However, it could be added to our dev Dockerfile

src/Contrib/Jaeger/SpanConverter.php
Grunet: Yeah this PR isn't meant to acheive spec compliance for the conversion, so maybe I should just take it out for now? It'll be available to reference in other spots as needed later anyways

src/Contrib/Jaeger/SpanConverter.php
Grunet: Huh, did not know about self, neat! Much more readable

src/Contrib/Jaeger/Transport.php
Grunet: Done

src/Contrib/Jaeger/JaegerTransport.php
Grunet: I think I see what you mean. Would that account for negative inputs too? I'm not sure if this class is supposed to just be an implementation detail or consumer facing, but if it's just the former that seems fine (doesn't even appear like that parameter is being used now that I look at it)

src/Contrib/Jaeger/SpanConverter.php
brettmc: scanning over the [jaeger thrift IDL](https://github.com/jaegertracing/jaeger-idl/blob/main/thrift/jaeger.thrift), the commented code looks like it would work, so I think either uncomment or leave it there and we can get it working in a later PR

src/Contrib/Jaeger/JaegerTransport.php
Grunet: Is there a way for one of the static analysis or code styling tools to let me know all the places that are missing return types? I'll try to go through and get them all just want to make sure I don't miss something

composer.json
brettmc: yeah, that makes sense - 32bit is probably more prevalent than I suspected

src/Contrib/Jaeger/JaegerTransport.php
Grunet: Updated

composer.json
Grunet: Updated to throw an exception here. I believe this will be unhandled rn, but if it's preferable to just log here after stopping to do things upstream I can add that in too

src/Contrib/Jaeger/JaegerTransport.php
Grunet: Updated as many as I noticed were missing (left the ones extending the abstract class from the thrift generated files as is)

src/Contrib/composer.json
Grunet: Moved it to a suggest here too. And I think Timo added it to the dev image for me a while back so we should be good there

Grunet:
Ok I think I've responded to everyone's initial comments. Lmk if I missed any and I'll loop back on them

composer.json
brettmc: Unhandled is fine. We can later make some of the factories deal with handling that exception.

src/Contrib/Jaeger/JaegerTransport.php
brettmc: is a negative buffer size valid? It doesn't sound like it would or should be

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -0,0 +1,189 @@
...
+    const STATUS_CODE_TAG_KEY = 'op.status_code';
+    const STATUS_DESCRIPTION_TAG_KEY = 'op.status_description';
+
+    /**
+    * Convert span to Jaeger Thrift Span format
+    */
+    public function convert(SpanDataInterface $span): JTSpan
+    {
+        $references = $tags = $logs = [];
+        $startTime = (int) ($span->getStartEpochNanos() / 1e3); // microseconds
```
brettmc, line 24: `AbstractClock::nanosToMicro()` does this

tests/Unit/Contrib/AgentExporterTest.php
```diff
@@ -0,0 +1,32 @@
...
+ * @covers OpenTelemetry\Contrib\Jaeger\AgentExporter
+ * @covers OpenTelemetry\Contrib\Jaeger\JaegerTransport
+ * @covers OpenTelemetry\Contrib\Jaeger\ThriftUdpTransport
+ */
+class AgentExporterTest extends TestCase
+{
+    public function test_happy_path()
+    {
+        $exporter = AgentExporter::fromConnectionString(
+            'http://127.0.0.1:80', //This isn't realistic I imagine
```
brettmc, line 22: that's fine for a unit test, and you can make it more obviously fake like some other tests do... `localhost:1234` etc

src/Contrib/Jaeger/JaegerTransport.php
Grunet: No clue. But if this is meant to be an internal-only class I can probably just get rid of that parameter and all the associated stuff. Not sure if that's the case though

src/Contrib/Jaeger/SpanConverter.php
Grunet: Makes sense to me, will leave it as is for now

tests/Unit/Contrib/AgentExporterTest.php
Grunet: Ah forgot about those comments! (removed them now) I think I looked super briefly to try and figure out what they're supposed to be in reality but couldn't find anything quickly and gave up.

Might be something to loop back on post functional testing!

src/Contrib/Jaeger/SpanConverter.php
Grunet: Switched!

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -0,0 +1,190 @@
...
+
+        // throw new \Exception('unsupported tag type');
+    }
+
+    private static function convertOtlpToJaegerIds(SpanDataInterface $span): array
+    {
+        if (PHP_INT_SIZE < 8) {
+            $humanReadableIntSize = PHP_INT_SIZE*8;
+
+            throw new RuntimeException("Integrating with Jaeger requires usage of 64 bit integers, but your current platform is $humanReadableIntSize bit. See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids for more information.");
```
brettmc, line 161: I think this belongs in a constructor (which you'll need to add) - better to fail fast if we can.

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -0,0 +1,190 @@
...
+
+        return [
+            'traceIdLow' => $traceIdLow,
+            'traceIdHigh' => $traceIdHigh,
+            'spanId' => $spanId,
+            'parentSpanId' => $parentSpanId,
+        ];
+    }
+
+    private static function convertOtlpToJaegerTraceIds(string $traceId): array
```
brettmc, line 180: Is there any benefit to this method being static?

src/Contrib/Jaeger/SpanConverter.php
Grunet: Sounds good to me, moved it into one

src/Contrib/Jaeger/SpanConverter.php
Grunet: I'm not sure. I don't really know what the idiomatic way of handling small side-effect-free helper functions like this is in PHP, but whatever that is I'd like to learn it (just been using static as the closest thing I'm aware of)

Overall I don't have a strong preference any way, so I can update it however folks think is best

tidal: @Grunet 
I use static methods for functions like that as well, if they do not touch any property. 
In that way it's clear that it's a helper function which could potentially be extracted.

tidal (approved):

Grunet:
> Some other questions of mine
> 
>     * Is CodecUtility.php still needed? It's unused and I finally sorta see what it was getting at, but I'm not sure it's needed anymore.
> 
>     * Are the 2 areas of commented out code in the SpanConverter file still needed? The first was there previously, but the 2nd I had to get rid of as being unreachable with the types in play
> 

Maybe @saktib if you see this soon, but I think my latest understanding of these is now

- CodecUtility was in-progress work trying to deal with OTLP spans keeping ids as hexadecimal strings, while Jaeger kept them as 64- or 128-bit integers
- The type-by-type creation of Jaeger tags is needed (see [this spec ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#attributes)section that alludes to it) but after the sanitiseTagValue method was copied from the Zipkin span converter, it made it impossible to actually use the needed code (?)

Based on that I think what seems appropriate is

- Delete CodecUtility
- Leave the 2nd commented out block, for reference for the next PR(s) that will be making Jaeger/SpanConverter.php spec compliant

@brettmc @tidal @bobstrecansky does that feel appropriate? (those were my last concerns with the PR. I can also bring them up at the SIG meeting if I'm able to make it)

bobstrecansky:
Sounds like a plan to me @Grunet !

Grunet:
Ok @brettmc @bobstrecansky the PR should be good to go from my end. Lmk if there's anything else I should do for it before merging

brettmc (approved):

saktib:
Thanks @Grunet to take this long-pending [#337](https://github.com/open-telemetry/opentelemetry-php/pull/337) PR to closure.

PR #547, opened 2022-01-17
Satisfy base spec for Jaeger Thrift span conversion

Grunet:
The[ Jaeger spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#opentelemetry-to-jaeger-transformation) mentions near the top that the "[generic spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/non-otlp.md#dropped-links-count)" for all exporters also applies, so this implements those requirements (except for the "dropped" ones. I opened an issue in the [spec repo](https://github.com/open-telemetry/opentelemetry-specification/issues/2268) about whether those should be followed)

I drew inspiration from these other language libraries' implementations while putting this together

[Java](https://github.com/open-telemetry/opentelemetry-java/blob/e0854d6ac11cfe4c3794255d365b2272939674a4/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java)
[Go](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/jaeger/jaeger.go)
[JS](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-exporter-jaeger/src/transform.ts)

*The new comment is a reminder for implementation of the main Jaeger spec later

### DIFF
--- a/src/Contrib/Jaeger/SpanConverter.php
+++ b/src/Contrib/Jaeger/SpanConverter.php
@@ -15,6 +15,8 @@ class SpanConverter
 {
     const STATUS_CODE_TAG_KEY = 'op.status_code';
     const STATUS_DESCRIPTION_TAG_KEY = 'op.status_description';
+    const KEY_INSTRUMENTATION_LIBRARY_NAME = 'otel.library.name';
+    const KEY_INSTRUMENTATION_LIBRARY_VERSION = 'otel.library.version';
 
     public function __construct()
     {
@@ -33,6 +35,14 @@ class SpanConverter
             self::STATUS_CODE_TAG_KEY => $span->getStatus()->getCode(),
             self::STATUS_DESCRIPTION_TAG_KEY => $span->getStatus()->getDescription(),
         ];
+
+        if (!empty($span->getInstrumentationLibrary()->getName())) {
+            $tags[SpanConverter::KEY_INSTRUMENTATION_LIBRARY_NAME] = $span->getInstrumentationLibrary()->getName();
+        }
+
+        if ($span->getInstrumentationLibrary()->getVersion() !== null) {
+            $tags[SpanConverter::KEY_INSTRUMENTATION_LIBRARY_VERSION] = $span->getInstrumentationLibrary()->getVersion();
+        }
 
         foreach ($span->getAttributes() as $k => $v) {
             $tags[$k] = $this->sanitiseTagValue($v);
--- a/src/Contrib/Jaeger/SpanConverter.php
+++ b/src/Contrib/Jaeger/SpanConverter.php
@@ -7,14 +7,15 @@ namespace OpenTelemetry\Contrib\Jaeger;
 use Jaeger\Thrift\Span as JTSpan;
 use Jaeger\Thrift\Tag;
 use Jaeger\Thrift\TagType;
+use OpenTelemetry\API\Trace\StatusCode;
 use OpenTelemetry\SDK\AbstractClock;
 use OpenTelemetry\SDK\Trace\SpanDataInterface;
 use RuntimeException;
 
 class SpanConverter
 {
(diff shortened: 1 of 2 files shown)


Grunet:
Not entirely sure what's going on with the failing phpt tests, other than it seems like deprecation warnings are flooding into the assertions somehow

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -29,10 +32,34 @@ public function convert(SpanDataInterface $span): JTSpan
...
-            self::STATUS_CODE_TAG_KEY => $span->getStatus()->getCode(),
-            self::STATUS_DESCRIPTION_TAG_KEY => $span->getStatus()->getDescription(),
-        ];
+
+        $tags = [];
+
+        if ($span->getStatus()->getCode() !== StatusCode::STATUS_UNSET) {
+            switch ($span->getStatus()->getCode()) {
+                case StatusCode::STATUS_OK:
+                    $tags[self::STATUS_CODE_TAG_KEY] = 'OK';
```
brettmc, line 41: these should be constants... eg STATUS_CODE_OK and STATUS_CODE_ERROR

brettmc (approved):
LGTM, just that one comment to be looked at.

brettmc:
re: the failing tests. Those are for php8.1 with ext-protobuf, and are known to fail. You can ignore them.

src/Contrib/Jaeger/SpanConverter.php
Grunet: This is a rather strange one. Java and JS are both using enums that are using full caps and so are able to refer directly back to the name for this, whereas Go is using normal case like PHP is and so has to do this too.

I tried to allude to that in the naming of the constants, but if more detailed comments around that are needed lmk

Grunet: *I'm unable to tell from the specs which is correct, as this one I followed for this PR suggests that it's the all caps version, but then the one linked to in the PHP class's comment seems to say go with the normal case version

Grunet: FWIW the spec suggesting all caps appears to be around 6 months more recently edited than the the one linked in the StatusCode class file, so maybe there was some changes around this that weren't fully updated? Just a guess really

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -38,12 +40,12 @@ public function convert(SpanDataInterface $span): JTSpan
         if ($span->getStatus()->getCode() !== StatusCode::STATUS_UNSET) {
             switch ($span->getStatus()->getCode()) {
                 case StatusCode::STATUS_OK:
-                    $tags[self::STATUS_CODE_TAG_KEY] = 'OK';
+                    $tags[self::STATUS_CODE_TAG_KEY] = self::STATUS_OK_WITH_ALL_UPPERCASE_LETTERS;
```
brettmc, line 43: I'd just name it `STATUS_OK` - if we learn later that it should be lower/mixed case, we can change the constant's value.

src/Contrib/Jaeger/SpanConverter.php
brettmc: a real-world test will get to the bottom of it, and we're still pre-alpha so there's plenty of opportunity for people to test it out. I suspect that the constants in `StatusCode` are correct for otel, but doesn't mean that Jaeger would use those internally, so let's stick with uppercase.

src/Contrib/Jaeger/SpanConverter.php
Grunet: Sure makes sense to me, updated

Issue #296, opened 2021-03-31
Implement Jaeger exporter


jodeev:
We currently use Zipkin exporter as Jaeger exporter, but the latter has diverged and should have its own implementation. 

See [Jaeger spec](https://github.com/open-telemetry/opentelemetry-specification/blob/418016dc4014f0c64acea74881774d34178bf15a/specification/trace/sdk_exporters/jaeger.md) .

Grunet:
I'll go ahead and plan on tackling this.

1st thing is that the above Jaeger spec link is stale. [Here's one pointing to the latest version](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md).

From a quick skim of it it seems like there may be 2 or 3 sections shared with the Zipkin exporter, but it may make sense to just start from scratch and then see what shakes out/makes sense to refactor at the end. (I also don't recall any of the Java, Python, Go, or JS ones for Zipkin I looked at seeming like they were sharing code with a Jaeger implementation, but I might be wrong)

Per the [language SDK compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md), it seems like Go, Java, Python, Ruby, and Rust have the best implementations of the spec so far.

Based on that, I think I'll take a look at/take notes on how Go, Java, and Python do their implementations for starters and use them to better understand the spec (with an ultimate aim of writing a PHP implementation based on theirs)

Grunet:
In-flight PRs relevant to this to keep in mind

- #337 
- #482 

Interesting [issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1973) from Java that has some interesting info/context around this.  From there I found the [SDK env variables for Jaeger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#jaeger-exporter), which also help to clarify what's needed in terms of serialization+communication protocols a little.

Jaeger docs

- Some details on [the Jaeger agent](https://www.jaegertracing.io/docs/1.29/deployment/#agent)
- Details on the [3 allowed serialization+communication protocols](https://www.jaegertracing.io/docs/1.23/apis/#thrift-over-udp-stable)

Grunet:
[...]

PR #548, opened 2022-01-17
Replace apache/thrift with packaged/thrift

tidal:
As discussesdin https://github.com/open-telemetry/opentelemetry-php/issues/296#issuecomment-987684152 , the original `apache/thrift`composer package includes the code of all language implementations, which makes the package rather huge (~15mb).
 The package `packaged/thrift` only includes the thrift PHP code extracted from the original repo as a drop-in replacement. According to [packagist](https://packagist.org/?query=thrift), this package is even more widely used than the original one.

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -361,7 +361,7 @@ return [
         'src',
         'proto',
         'thrift',
-        'vendor/apache/thrift',
+        'vendor/packaged/thrift',
         'vendor/composer/xdebug-handler/src',
         'vendor/guzzlehttp',
         'vendor/psr',
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "ext-json": "*",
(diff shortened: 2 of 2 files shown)


tidal:
Phan seems to choke on this and still expects `apache/thrift` to be present. Needs further investigation.

brettmc:
@tidal another one-line change, then phan is happy:
```
diff --git a/.phan/config.php b/.phan/config.php
index e28a22e..3f50e1f 100644
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -361,7 +361,7 @@ return [
         'src',
         'proto',
         'thrift',
-        'vendor/apache/thrift',
+        'vendor/packaged/thrift',
         'vendor/composer/xdebug-handler/src',
         'vendor/guzzlehttp',
         'vendor/psr',
```

tidal:
@brettmc 
Thanks for investigating. I did not really think of phan's config.

brettmc (approved):

PR #556, opened 2022-01-26
Bring Jaeger Thrift Span Converter up to Spec

Grunet:
Implements these 3 sections of the spec, which appear to be the last ones affecting the thrift span converter

- [Error flag](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#resource)
- Allowing for more than just string-valued tags (per the [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#attributes) section)
- [Links](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#links)
    -[ Java](https://github.com/open-telemetry/opentelemetry-java/blob/e0854d6ac11cfe4c3794255d365b2272939674a4/exporters/jaeger-thrift/src/main/java/io/opentelemetry/exporter/jaeger/thrift/Adapter.java#L235) and [Python](https://github.com/open-telemetry/opentelemetry-python/blob/main/exporter/opentelemetry-exporter-jaeger-thrift/src/opentelemetry/exporter/jaeger/thrift/translate/__init__.py#L255) implementations for reference
    
    (The first commit below that wasn't from the previous PR is the one with the message "Uncommenting the handling for tags of non-string types" with short sha fdcfe8b)


src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -174,28 +173,88 @@ private static function convertOtelSpanDataToJaegerTags(SpanDataInterface $span)
...
+        }
+
+        return $tags;
+    }
+
+    private static function buildTag(string $key, $value): Tag
+    {
+        $valueToCreateTagFrom = self::convertValueToTypeJaegerTagsSupport($value);
+
+        return self::createJaegerTagInstance($key, $valueToCreateTagFrom);
```
tidal, line 197: Can you make this one call please? 
```php 
        return self::createJaegerTagInstance(
           $key, 
           self::convertValueToTypeJaegerTagsSupport($value)
         );
```

src/Contrib/Jaeger/SpanConverter.php
```diff
@@ -174,28 +173,88 @@ private static function convertOtelSpanDataToJaegerTags(SpanDataInterface $span)
...
+                'key' => $key,
+                'vType' => TagType::LONG,
+                'vLong' => $value,
+            ]);
+        } elseif (is_numeric($value)) {
+            return new Tag([
+                'key' => $key,
+                'vType' => TagType::DOUBLE,
+                'vDouble' => $value,
+            ]);
```
tidal, line 228: The Else-Police says: In this case else is really not needed, since you are returning values.
```php
        if (is_bool($value)) {
            return new Tag([
                'key' => $key,
                'vType' => TagType::BOOL,
                'vBool' => $value,
            ]);
        } 
        if (is_integer($value)) {
            return new Tag([
                'key' => $key,
                'vType' => TagType::LONG,
                'vLong' => $value,
            ]);
        } 
        if (is_numeric($value)) {
            return new Tag([
                'key' => $key,
                'vType' => TagType::DOUBLE,
                'vDouble' => $value,
            ]);
        }
```

tidal (approved):

tidal:
@Grunet 
Thank you for your work.
Just two nit-picky comments from my side.

Grunet:
I realized there was a line of code (the bool to string conversion in recursivelySerializeArray) that was previously covered that wasn't anymore after the latest changes, so I updated one of the tests to bring that coverage back.

Not sure what happened to the standalone psalm job when the checks re-ran, intermittent issue with the action?

src/Contrib/Jaeger/SpanConverter.php
Grunet: Very much agreed, updated!

src/Contrib/Jaeger/SpanConverter.php
Grunet: Yup np, updated!

tidal (approved):

tidal:
> Not sure what happened to the standalone psalm job when the checks re-ran, intermittent issue with the action? (seems fine when I run it locally at least)

That seems to be an error in psalm itself. However i ran the job again, and it still failed.
It's an error caused by an `assert`, so it's likely, that you don't have `assert` activated locally or this is an issue
with a specific psalm version.

```php 
Uncaught RuntimeException: PHP Error: assert(): assert($parent_const_storage !== null) failed in ...
 ```
 
 However, if you are fine with the PR, we can merge it.

Grunet:
> > Not sure what happened to the standalone psalm job when the checks re-ran, intermittent issue with the action? (seems fine when I run it locally at least)
> 
> That seems to be an error in psalm itself. However i ran the job again, and it still failed. It's an error caused by an `assert`, so it's likely, that you don't have `assert` activated locally or this is an issue with a specific psalm version.
> 
> ```
> Uncaught RuntimeException: PHP Error: assert(): assert($parent_const_storage !== null) failed in ...
> ```
> 
> However, if you are fine with the PR, we can merge it.

Yeah I think I found [the commit](https://github.com/vimeo/psalm/commit/1a6f968e157cadfe073e4992ff7dd062d67286ad) where they fixed it (or at least removed the line of code it's crashing on here) in psalm's master branch (which I guess the [action uses](https://github.com/psalm/psalm-github-actions/blob/master/Dockerfile#L25)?) around 2 hours ago. 

So I'm guessing that something with the action's use of the psalm master branch is maybe being cached (per the code comment in the above link. Not quite sure how tbh...) so we continue to get a more stale version that still has this problem.

The psalm checks in the PHP Composer/build jobs all still work though so, @tidal I think this is safe to merge

PR #559, opened 2022-01-28
Make it possible to pass endpoint url and headers to constructor of o…

markvdlaan93:
…tlp http exporter

### DIFF
--- a/src/Contrib/OtlpHttp/Exporter.php
+++ b/src/Contrib/OtlpHttp/Exporter.php
@@ -52,21 +52,16 @@ class Exporter implements SpanExporterInterface
         RequestFactoryInterface $requestFactory,
         StreamFactoryInterface $streamFactory,
         SpanConverter $spanConverter = null,
-        string $endpointUrl = null,
+        string $endpointUrl = 'https://localhost:4318/v1/traces',
         string $headers = null
     ) {
-        if(!$endpointUrl){
-            $endpointUrl = $this->getStringFromEnvironment('OTEL_EXPORTER_OTLP_ENDPOINT', 'https://localhost:4318/v1/traces');
-        }
+        $endpointUrl = $this->getStringFromEnvironment('OTEL_EXPORTER_OTLP_ENDPOINT', $endpointUrl);
 
         $this->setEndpointUrl($this->validateEndpoint($endpointUrl));
         // @todo: Please, check if this code is needed. It creates an error in phpstan, since it's not used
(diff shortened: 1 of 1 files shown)


src/Contrib/OtlpHttp/Exporter.php
```diff
@@ -51,17 +51,23 @@ public function __construct(
         ClientInterface $client,
         RequestFactoryInterface $requestFactory,
         StreamFactoryInterface $streamFactory,
-        SpanConverter $spanConverter = null
+        SpanConverter $spanConverter = null,
+        string $endpointUrl = null,
```
brettmc, line 55: If you use the default value from below here (localhost:4318), the if below is not required

src/Contrib/OtlpHttp/Exporter.php
```diff
@@ -51,17 +51,23 @@ public function __construct(
...
-        );
+        if(!$endpointUrl){
+            $endpointUrl = $this->getStringFromEnvironment('OTEL_EXPORTER_OTLP_ENDPOINT', 'https://localhost:4318/v1/traces');
+        }
+
+        $this->setEndpointUrl($this->validateEndpoint($endpointUrl));
         // @todo: Please, check if this code is needed. It creates an error in phpstan, since it's not used
         // $this->certificateFile = getenv('OTEL_EXPORTER_OTLP_CERTIFICATE') ?: 'none';
-        $this->headers = $this->processHeaders($this->getStringFromEnvironment('OTEL_EXPORTER_OTLP_HEADERS', ''));
+        if($headers){
```
brettmc, line 65: as above, use a default constructor arg and then the if is not required

brettmc:
Thanks for your contribution - can you please sign the CLA so that we can accept the code. Please also run the tests, checks etc locally ("make all"), there should be a couple of tests that need to be updated.

brettmc:
@markvdlaan93 a fix is coming for the psalm error that's failing the tests, it's not related to your changes. In the meantime, you should be able to run all of the tests and checks locally (there are instructions in the readme). Also please look at whether any examples can be updated/improved based on your changes.

tidal:
Thank you for your contribution.

I would prefer any direct configuration of an exporter to be the first arguments in the constructor.
This is more consistent with the other exporters and the (direct) dependency on the HTTP client etc, is subject to change anyway. This is a WIP of me (though I got distracted by some other stuff.).
You can take a look at this issues to get an idea: https://github.com/open-telemetry/opentelemetry-php/issues/451

tidal:
Closing this PR, since there has been no activity for over a month.
Fell free to reopen.

PR #562, opened 2022-02-02
Adding event and listener

kpratyus:
This is the first draft for https://github.com/open-telemetry/opentelemetry-php/issues/298. In this PR I have implemented an API for event and listener and how it can be used.

- SubscribedEvent Class would be the exposed hook through which would be extracting data from the library.
- Listeners would be the callback(maybe user defined or automatic) which would make changes according to type of even
- Finer implementation details like  how the listeners are called and how many events have to be created need to be 
   discussed with the community after there is some agreement about the API.

### DIFF
--- a/examples/SubscribedListenerLaravel.php
+++ b/examples/SubscribedListenerLaravel.php
@@ -1,0 +1,104 @@
+<?php
+
+use Illuminate\Contracts\Http\Kernel;
+use Illuminate\Http\Request;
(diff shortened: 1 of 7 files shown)


bobstrecansky:
Thank you for contributing to opentelemetry-php.
Since we are a [CNCF](https://www.cncf.io/) project, we need contributors to sign a Contributor License Agreement ([see above](https://github.com/open-telemetry/opentelemetry-php/pull/562#issuecomment-1027947318))

I've approved and run your Github Actions for CI.

PR #563, opened 2022-02-02
Adding event and listener

kpratyus:
This is the first draft for https://github.com/open-telemetry/opentelemetry-php/issues/298 . In this PR I have implemented an API for event and listener and how it can be used.

- SubscribedEvent Class would be the exposed hook through which would be extracting data from the library.

- Listeners would be the callback(maybe user defined or automatic) which would make changes according to type of even

- Finer implementation details like how the listeners are called and how many events have to be created need to be

discussed with the community after there is some agreement about the API.

### DIFF
--- a/examples/SubscribedListenerLaravel.php
+++ b/examples/SubscribedListenerLaravel.php
@@ -1,0 +1,104 @@
+<?php
+
+use Illuminate\Contracts\Http\Kernel;
+use Illuminate\Http\Request;
+use OpenTelemetry\Contrib\Jaeger\Exporter as JaegerExporter;
+use OpenTelemetry\Contrib\Zipkin\Exporter as ZipkinExporter;
+use OpenTelemetry\SDK\SystemClock as SystemClock;
+use OpenTelemetry\Context\Context;
+use OpenTelemetry\SDK\Trace\Sampler\AlwaysOnSampler as AlwaysOnSampler;
+# use OpenTelemetry\Sdk\Trace\Sampler as AlwaysOnSampler;
+use OpenTelemetry\SDK\Trace\SamplingResult;
+use OpenTelemetry\SDK\Trace\SpanProcessor\BatchSpanProcessor;
+use OpenTelemetry\SDK\Trace\TracerProvider;
+use OpenTelemetry\API\Trace as API;
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\HttpFactory;
+define('LARAVEL_START', microtime(true));
+
+/*
+|--------------------------------------------------------------------------
+| Check If The Application Is Under Maintenance
+|--------------------------------------------------------------------------
+|
+| If the application is in maintenance / demo mode via the "down" command
+| we will load this file so that any pre-rendered content can be shown
+| instead of starting the framework, which could cause an exception.
+|
+*/
+
+if (file_exists(__DIR__.'/../storage/framework/maintenance.php')) {
+    require __DIR__.'/../storage/framework/maintenance.php';
+}
+
+/*
+|--------------------------------------------------------------------------
+| Register The Auto Loader
+|--------------------------------------------------------------------------
+|
+| Composer provides a convenient, automatically generated class loader for
+| this application. We just need to utilize it! We'll simply require it
+| into the script here so we don't need to manually load our classes.
+|
+*/
+
+require __DIR__.'/../vendor/autoload.php';
+
+/*
+|--------------------------------------------------------------------------
+| Run The Application
+|--------------------------------------------------------------------------
+|
+| Once we have the application, we can handle the incoming request using
+| the application's HTTP kernel. Then, we will send the response back
+| to this client's browser, allowing them to enjoy our application.
+|
+*/
+$sampler = new AlwaysOnSampler();
+$samplingResult = $sampler->shouldSample(
+    new Context(),
+    md5((string) microtime(true)),
+    'io.opentelemetry.example',
+    API\SpanKind::KIND_INTERNAL
+);
+$zipkinExporter = new ZipkinExporter(
+    'Hello World Web Server Zipkin',
+    'http://docker.for.mac.localhost:9411/api/v2/spans',
+    new Client(),
+    new HttpFactory(),
+    new HttpFactory()
+);
+$jaegerExporter = new JaegerExporter(
+    'Hello World Web Server Jaeger',
+    'http://localhost:9412/api/v2/spans',
+    new Client(),
+    new HttpFactory(),
+    new HttpFactory()
+);
+if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
+
+    $jaegerTracer = (new TracerProvider(new BatchSpanProcessor($jaegerExporter, SystemClock::getinstance()), $sampler))
+        ->getTracer('io.opentelemetry.contrib.php');
+
+    $zipkinTracer = (new TracerProvider(new BatchSpanProcessor($zipkinExporter, SystemClock::getinstance()), $sampler))
+        ->getTracer('io.opentelemetry.contrib.php');
+
+    $request = Request::createFromGlobals();
+    
+    $jaegerSpan = $jaegerTracer->spanBuilder($request->getUri())->addListener('SubscribedEvent','StartSpanListener')->startSpan();
+    $zipkinSpan = $zipkinTracer->spanBuilder($request->getUri())->addListener('SubscribedEvent','StartSpanListener')->startSpan();
+}
+
+$app = require_once __DIR__.'/../bootstrap/app.php';
+
+$kernel = $app->make(Kernel::class);
+
+$response = $kernel->handle(
+    $request = Request::capture()
+)->send();
+
+$kernel->terminate($request, $response);
+if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
+    $zipkinSpan->end();
+    $jaegerSpan->end();
+}
--- a/src/SDK/Subscriber/Listener/StartSpanListener.php
+++ b/src/SDK/Subscriber/Listener/StartSpanListener.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber\Listener;
+
+use function count;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\AttributesInterface;
+use OpenTelemetry\SDK\Trace\Span;
+use OpenTelemetry\SDK\Subscriber\ListenerInterface;
+
+class StartSpanListener implements ListenerInterface
+{
+
+	public function action(Span $span):void
+	{
+		$span->setAttribute('Listener','Does_some_action');
+
+	}
+}
--- a/src/SDK/Subscriber/ListenerInterface.php
+++ b/src/SDK/Subscriber/ListenerInterface.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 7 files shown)


prondubuisi:
Hello @kpratyus , you need to follow this link https://contributor.easycla.lfx.linuxfoundation.org/#/cla/project/d8cead54-92b7-48c5-a2c8-b1e295e8f7f1/user/d5022c4b-1186-4037-a564-17de52c88e4f?redirect=https:%2F%2Fgithub.com%2Fopen-telemetry%2Fopentelemetry-php%2Fpull%2F563 , then click proceed as individual contributor to make EasyCLA happy

kpratyus:
> Hello @kpratyus , you need to follow this link https://contributor.easycla.lfx.linuxfoundation.org/#/cla/project/d8cead54-92b7-48c5-a2c8-b1e295e8f7f1/user/d5022c4b-1186-4037-a564-17de52c88e4f?redirect=https:%2F%2Fgithub.com%2Fopen-telemetry%2Fopentelemetry-php%2Fpull%2F563 , then click proceed as individual contributor to make EasyCLA happy

I am a corporate contributor and for some reason I am facing some issues. I am working with the folks inside my company and will complete the EasyCLA.

examples/SubscribedListenerLaravel.php
```diff
@@ -0,0 +1,104 @@
...
+|--------------------------------------------------------------------------
+| Run The Application
+|--------------------------------------------------------------------------
+|
+| Once we have the application, we can handle the incoming request using
+| the application's HTTP kernel. Then, we will send the response back
+| to this client's browser, allowing them to enjoy our application.
+|
+*/
+$sampler = new AlwaysOnSampler();
```
brettmc, line 57: this example is based on old code (I think around `0.3`), but won't work with `main`

brettmc:
At a high level, I think this should be using [psr-14](https://www.php-fig.org/psr/psr-14), and allow users to provide their own psr-14 implementation. I'm unsure whether not having a psr-14 implementation should be optional? I believe this will allow better integration with different frameworks.

examples/SubscribedListenerLaravel.php
```diff
@@ -0,0 +1,104 @@
+<?php
+
```
tidal, line 2: I don't really understand the purpose of this file.
It looks like a front controller, but a lot of the dependencies are missing.
Also cross cutting concerns should generally not go into a front controller.

examples/SubscribedListenerLaravel.php
```diff
@@ -0,0 +1,104 @@
...
+| Run The Application
+|--------------------------------------------------------------------------
+|
+| Once we have the application, we can handle the incoming request using
+| the application's HTTP kernel. Then, we will send the response back
+| to this client's browser, allowing them to enjoy our application.
+|
+*/
+$sampler = new AlwaysOnSampler();
+$samplingResult = $sampler->shouldSample(
```
tidal, line 58: This should not be used anymore.

examples/SubscribedListenerLaravel.php
```diff
@@ -0,0 +1,104 @@
...
+    new HttpFactory()
+);
+$jaegerExporter = new JaegerExporter(
+    'Hello World Web Server Jaeger',
+    'http://localhost:9412/api/v2/spans',
+    new Client(),
+    new HttpFactory(),
+    new HttpFactory()
+);
+if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
```
tidal, line 78: This should not be used anymore.

examples/SubscribedListenerLaravel.php
```diff
@@ -0,0 +1,104 @@
...
+$app = require_once __DIR__.'/../bootstrap/app.php';
+
+$kernel = $app->make(Kernel::class);
+
+$response = $kernel->handle(
+    $request = Request::capture()
+)->send();
+
+$kernel->terminate($request, $response);
+if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
```
tidal, line 101: This should not be used anymore.

src/SDK/Subscriber/Listener/StartSpanListener.php
```diff
@@ -0,0 +1,21 @@
...
+use function count;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\AttributesInterface;
+use OpenTelemetry\SDK\Trace\Span;
+use OpenTelemetry\SDK\Subscriber\ListenerInterface;
+
+class StartSpanListener implements ListenerInterface
+{
+
+	public function action(Span $span):void
```
tidal, line 16: A method name should be a verb.
A listener/subscriber should depend on events or callbacks not on concrete objects, otherwise it would be an observer.

src/SDK/Subscriber/ListenerInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber;
+
+use function count;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\AttributesInterface;
```
tidal, line 9: Where are these imports used?

src/SDK/Subscriber/ListenerInterface.php
```diff
@@ -0,0 +1,16 @@
...
+use OpenTelemetry\SDK\Trace\Span;
+
+interface ListenerInterface
+{
+
+	public function action(Span $span):void;
```
tidal, line 15: see above

src/SDK/Subscriber/SubscribedEvent.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace OpenTelemetry\SDK\Subscriber;
+
+use function count;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\AttributesInterface;
+use OpenTelemetry\SDK\Trace;
+use OpenTelemetry\SDK\Trace\Span;
+
+final class SubscribedEvent implements SubscribedEventInterface
```
tidal, line 13: What does `SubscribedEvent` mean?

src/SDK/Subscriber/SubscribedEvent.php
```diff
@@ -0,0 +1,38 @@
...
+	private Span $span;
+
+	public function __construct(Span $span)
+	{
+		$this->span = $span;
+	}
+
+	public function dispatch():void
+	{
+		$classname = get_called_class();
```
tidal, line 25: What's the purpose of calling `get_called_class` , when it's a final class?

src/SDK/Subscriber/SubscribedEvent.php
```diff
@@ -0,0 +1,38 @@
...
+		$classname = str_replace('OpenTelemetry\\SDK\\Subscriber\\','',$classname);
+		$listenerClassName = $this->span->getListener($classname);
+		$listenerClassName = "OpenTelemetry\\SDK\\Subscriber\\Listener\\".$listenerClassName;
```
tidal, line 28: This looks unreliable.
How do you want to test this?

src/SDK/Subscriber/SubscribedEventInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber;
+
+use function count;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\AttributesInterface;
+use OpenTelemetry\SDK\ListenerInterface as ListenerInterface;
```
tidal, line 10: Where are these imports used?

src/SDK/Trace/Span.php
```diff
@@ -149,12 +155,17 @@ public static function startSpan(
...
+            $startEpochNanos,
+            $listener
         );
 
         // Call onStart here to ensure the span is fully initialized.
         $spanProcessor->onStart($span, $parentContext);
 
+        // Create a event and dispatch
+        $subscribedevent = new SubscribedEvent($span);
+        $subscribedevent->dispatch();
```
tidal, line 167: An event should never disptach itself.

src/SDK/Trace/SpanBuilder.php
```diff
@@ -103,6 +106,18 @@ public function addLink(API\SpanContextInterface $context, iterable $attributes
         return $this;
     }
 
+    public function addListener(string $subscribedEvent, string $listener): self
+    {
+        if(count($this->listener)==0){
```
tidal, line 111: Please, don't use weak comparisons.

src/SDK/Trace/SpanBuilder.php
```diff
@@ -103,6 +106,18 @@ public function addLink(API\SpanContextInterface $context, iterable $attributes
...
+            $this->listener = [$subscribedEvent => $listener];
+        }
+        else
```
tidal, line 114: This `else` is unnecessary. (Like 99.9% of all elses)

tidal (changes requested):

tidal:
Thank you for your PR.

In general I agree with @brettmc on PSR14. though there is not necessarily a need use 3rd party implementations, as these are failry trivial interfaces to implement.

I added a few comments, however here are some general aspects:

The purpose of event driven approaches is to decouple behavior. In other words, an object/componet exposing events should not be coupled to whoever/whatever is listening/subscribing to it. Your implementation is kind of tightly coupled.
An event should not disptach itself, nor should there only one possible listener/subscriber/observer.

Having object based events doesn't really make sense in our case. It's useful in GUIs, for example event listeners for DOM events in a HTML page, but in backend applications usually live-cycle events are more useful, as there usually is an interest in events of all spans, and not a specific single one. (Enteties might be an exception, but that's not what we are dealing with).

`SpanProcessors` would be a a better place to implement such life-cycle-events, as there are already callbacks like `onStart`, `onEnd`. So there could be a SpanProcessor which simply handles exposing events for a span.

kpratyus:
Thanks @tidal and @brettmc for your reviews and comments. I will incorporate your changes and raise a PR again. Thanks!!

src/SDK/Subscriber/Listener/StartSpanListener.php
kpratyus: In the Otel-php module case isn't the observer mechanism suited better with the user registering the observer with the Events to be listened?

src/SDK/Subscriber/SubscribedEvent.php
kpratyus: It's the nomenclature I have used for Event class

src/SDK/Subscriber/SubscribedEvent.php
kpratyus: In the implementation the process the user could use to register the Listener with Event is as follows-

- During the instantiation of Span builder class the user will provide the The event class name with which he/she wants to register the listener along-with the Listener class name both in string format.  So to correlate the Listener class with Event class there names have to be known for which "get_called_class" is used.

src/SDK/Subscriber/SubscribedEvent.php
kpratyus: In the implementation the process the user could use to register the Listener with Event is as commented above. Now to register the Listener class the class file should be present in a fixed folder that's why I am using a static path.

bobstrecansky:
@kpratyus - are you still planning on updating this PR?

kpratyus:
> @kpratyus - are you still planning on updating this PR?

Yes, I am about to have discussions regarding this PR in SIG today. Then I will update it.

kpratyus:
Raising a new PR

PR #576, opened 2022-02-07
Scaffold Jaeger Thrift Http Exporter

Grunet:
I started out by adapting what the jaeger-client-php library had [already done](https://github.com/jonahgeorge/jaeger-client-php/blob/e2610656c30f461926bd55a2b726c23835d98b90/src/Jaeger/ReporterFactory/JaegerHttpReporterFactory.php), but ended up removing most of its relevant code while trying to integrate it into the existing patterns here.

I incidentally noticed that the SpanConverter for Jaeger Thrift wasn't implementing the same interface the other span converters were, so I also went ahead and fixed that up.

I haven't functional tested yet against the all-in-one container (was planning to try and do so after resolving that commented out TODO block, which I think might correspond to the last piece of the thrift-parts of the spec that needs work)

And here are some questions that came up as I was putting this together:
- Does this codebase already have a way for tests to test "through http clients"? (or in this case fopen?)
    - The smoke test I added is currently actually making a network request (which was helpful though as it helped me spot 2 bugs) so I was curious if other tests have already handled this
     - I'd also be fine switching it to not make network requests (since at this point I think I have somewhat of an understanding of what THttpClient::flush is up to now)
- There are now 3 Exporters in the Contrib/Jaeger namespace, the oldest of which (providing support through Jaeger's compatibility with Zipkin spans as I gather) is just called "Exporter"
    - I was thinking of renaming it but thought that might be overly disruptive and decided against it. (I guess this is less a question and more just a note...)
- AgentExporter now has a little duplication in constructor validation logic with this new exporter, but I wasn't sure if that needed to be addressed rn or what the preferred way would be (e.g. superclass, trait, helper class) so I left it as is for the time being


tidal:
> Does this codebase already have a way for tests to test "through http clients"? (or in this case fopen?)
> * The smoke test I added is currently actually making a network request (which was helpful though as it helped me spot 2 bugs) so I was curious if other tests have already handled this
> * I'd also be fine switching it to not make network requests (since at this point I think I have somewhat of an understanding of what THttpClient::flush is up to now)

No there is no way to test network requests yet, as this is usually done in functional, e2e or acceptance tests. And I am still working on the infrastructure to create those automatically.

However, we should really not use Thrift's [THttpClient](https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/THttpClient.php).
- It uses file operations as network requests, which are not guaranteed to work, as this can be turned on/off via php.ini.
- It [supresses error messages](https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/THttpClient.php#L172), which is generally not a good idea.
- It reinvents a wheel, which already has been invented many times, without making it any better.
- I personally do not trust the code of people [who blame their lack of knowledge of a technology on the technology itself](https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/THttpClient.php#L211).

So it would be nice, if you make this work with a PSR Client like the other exporters. It should not really be complicated, otherwise we can help.



> There are now 3 Exporters in the Contrib/Jaeger namespace, the oldest of which (providing support through Jaeger's compatibility with Zipkin spans as I gather) is just called "Exporter"
> * I was thinking of renaming it but thought that might be overly disruptive and decided against it. (I guess this is less a question and more just a note...)

No need to overthink this for now. I will most probably refactor a lot of stuff anyway, when you are done with the exporters, and I can finish up the exporter rework
(No pressure though, it's not like I don't have anything to do)



> * AgentExporter now has a little duplication in constructor validation logic with this new exporter, but I wasn't sure if that needed to be addressed rn or what the preferred way would be (e.g. superclass, trait, helper class) so I left it as is for the time being

I'd  usually go with a trait or helper class. I rarely see any use for superclasses, since traits exists, as they most of the time don't bring any benefit.

src/Contrib/Jaeger/HttpCollectorExporter.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    use UsesSpanConverterTrait;
+    use SpanExporterTrait;
+
+    private SpanConverter $spanConverter;
+
+    private ThriftHttpSender $sender;
+
+    public function __construct(
+        $name,
```
tidal, line 22: I guess `$name` is s `string`?

src/Contrib/Jaeger/HttpCollectorExporter.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if (!isset($parsedDsn['host']) || !isset($parsedDsn['port'])) {
+            throw new InvalidArgumentException('Endpoint should have host, port');
+        }
+
+        $this->sender = new ThriftHttpSender(
+            $name,
+            $parsedDsn['host'],
+            $parsedDsn['port'],
+            isset($parsedDsn['path']) ? $parsedDsn['path'] : '', //Matching THttpClient's default
```
tidal, line 39: You can use the short notation (also see below):
```php
$parsedDsn['path'] ?: ''
```

src/Contrib/Jaeger/HttpCollectorExporter.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * @psalm-return SpanExporterInterface::STATUS_*
+     */
+    public function doExport(iterable $spans): int
+    {
+        $thriftSpans = $this->spanConverter->convert($spans);
+
+        $this->sender->send($thriftSpans);
```
tidal, line 53: This can be one call:
```php
$this->sender->send(
   $this->spanConverter->convert($spans)
);
````

File: src/Contrib/Jaeger/ThriftHttpSender.php [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+    public function __construct(
+        string $serviceName,
+        string $host,
+        int $port,
+        string $path,
+        string $scheme
+    ) {
+        $this->serviceName = $serviceName;
+
+        $transport = new THttpClient(
```
tidal, line 32: Please don't use this client (see my comment)

tests/Unit/Contrib/JaegerHttpCollectorExporterTest.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @covers OpenTelemetry\Contrib\Jaeger\HttpCollectorExporter
+ * @covers OpenTelemetry\Contrib\Jaeger\ThriftHttpSender
+ */
+class JaegerHttpCollectorExporterTest extends TestCase
+{
+    public function test_happy_path()
+    {
+        $exporter = HttpCollectorExporter::fromConnectionString(
+            'https://httpbin.org:443/post', //FYI this will actually end up making a network request
```
tidal, line 21: See my comment. Unit tests should never do any I/O.

tests/Unit/Contrib/JaegerSpanConverterTest.php
```diff
@@ -26,7 +26,7 @@ public function test_should_convert_an_otel_span_to_a_jaeger_thrift_span()
         $span = (new SpanData())
                     ->setName('otelSpanName');
 
-        $jtSpan = (new SpanConverter())->convert($span);
+        [$jtSpan] = (new SpanConverter())->convert([$span]);
```
tidal, line 29: What does `jtSpan`mean?

src/Contrib/Jaeger/HttpCollectorExporter.php
```diff
@@ -0,0 +1,66 @@
...
+        string $endpointUrl
+    ) {
+        $parsedDsn = parse_url($endpointUrl);
+
+        if (!is_array($parsedDsn)) {
+            throw new InvalidArgumentException('Unable to parse provided DSN');
+        }
+
+        if (!isset($parsedDsn['host']) || !isset($parsedDsn['port'])) {
+            throw new InvalidArgumentException('Endpoint should have host, port');
```
tidal, line 32: Why does the endpoint need to have a port?

src/Contrib/Jaeger/AgentExporter.php
```diff
@@ -53,7 +53,7 @@ public function doExport(iterable $spans): int
     {
         // UDP Transport begins here after converting to thrift format span
         foreach ($spans as $span) {
-            $cSpan = $this->spanConverter->convert($span);
+            [$cSpan] = $this->spanConverter->convert([$span]);
```
tidal, line 56: I thought cspan is a TV station, no?
Anyway, this could be one call:
```php
$this->jaegerTransport->append(
   $this->spanConverter->convert([$span]), 
   $this->serviceName
);
````

**Review by tidal — CHANGES_REQUESTED** [review]

File: src/Contrib/Jaeger/ThriftHttpSender.php (hunk 4 above)
Comment (Grunet) [reply src/Contrib/Jaeger/ThriftHttpSender.php:32]: Yeah I agree current state is not gonna work having stepped back from it a bit now. I'm thinking maybe I'll try 2 things along the lines of your direction

1) Rewrite THttpClient::flush to use a PSR-18 HTTP client injected by consumers (like I gather the other exporters are already doing from checking a few)
2) Try and see if apache/thrift would be interested in moving this change upstream (or if there's maybe some v subtle reason fopen has to be used there?) since it doesn't look like that code has been touched much since 3-4 years ago

I'll go ahead and pop open an issue in their repo for 2) to gauge interest, then start on giving 1) a shot

Comment (tidal) [reply src/Contrib/Jaeger/ThriftHttpSender.php:32]: Thank you.
I believe the initial code-base is older than PSR-18.
FWIW there is also a [TCurlClient](https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/TCurlClient.php), however there are no tests for both clients.

Comment (Grunet) [reply src/Contrib/Jaeger/ThriftHttpSender.php:32]: Took a stab at it. You can see my current progress in this draft PR I made inside my own fork - https://github.com/Grunet/opentelemetry-php/pull/2/files Still some cleanup work to do but figured it was shareable

I wanted to start with a normal composed adapter for THttpClient but realized eventually all of its state was in protected variables, so I switched to just a subclass (since it seemed like I'd need one anyways) and injected the extras it would need (not sure this is the best approach, just the first that came to mind after the pivot)

I also cut out a lot of the aspects of the "flush" method that otel-php doesn't need to use, so probably no simple path to upstreaming this at the moment.

Would appreciate any pointers or comments on things that are obv off (this is the first time I think I've interacted with the PSR-18/PSR-7 interfaces directly, and it was a neat experience!)

Comment (tidal) [reply src/Contrib/Jaeger/ThriftHttpSender.php:32]: @Grunet 
Sorry for the late reply, I was a bit busy.
Let's talk about this in tomorrows meeting, if you still have questions.

File: src/Contrib/Jaeger/HttpCollectorExporter.php (hunk 1 above)
Comment (Grunet) [reply src/Contrib/Jaeger/HttpCollectorExporter.php]: Yup, updated

File: src/Contrib/Jaeger/HttpCollectorExporter.php (hunk 3 above)
Comment (Grunet) [reply src/Contrib/Jaeger/HttpCollectorExporter.php]: Yup, updated

File: tests/Unit/Contrib/JaegerSpanConverterTest.php (hunk 6 above)
Comment (Grunet) [reply tests/Unit/Contrib/JaegerSpanConverterTest.php:29]: I think the agent exporter was using it as shorthand for "Jaeger Thift Span" iirc, also in part because the thrift library's exported type was also called "Span", so I think in the converter file this was renamed to JTSpan

Definitely open to switching it up though however is more understandable

File: src/Contrib/Jaeger/AgentExporter.php (hunk 8 above)
Comment (Grunet) [reply src/Contrib/Jaeger/AgentExporter.php]: Yeah I didn't quite understand the naming either...but in any case updated to one call instead

File: src/Contrib/Jaeger/HttpCollectorExporter.php (hunk 2 above)
Comment (Grunet) [reply src/Contrib/Jaeger/HttpCollectorExporter.php:39]: Ah actually it seems like Phan is concerned about the key not existing (which may've been why I used the isset and the ternary before now that I try to remember back). 

Is there a way to check for existence but used the condensed form still?

Here's the Phan error output

src/Contrib/Jaeger/HttpCollectorExporter.php:39 PhanTypePossiblyInvalidDimOffset Possibly invalid offset 'path' of $parsedDsn of array type array{port:int,host:string,scheme?:string,user?:string,pass?:string,path?:string,query?:string,fragment?:string}
src/Contrib/Jaeger/HttpCollectorExporter.php:40 PhanTypePossiblyInvalidDimOffset Possibly invalid offset 'scheme' of $parsedDsn of array type array{port:int,host:string,scheme?:string,user?:string,pass?:string,path?:string,query?:string,fragment?:string}

File: src/Contrib/Jaeger/HttpCollectorExporter.php (hunk 7 above)
Comment (Grunet) [reply src/Contrib/Jaeger/HttpCollectorExporter.php]: It looks like THttpClient uses it in the exception messages it's "read" method can throw, and then also in its "flush" method (here's a [link to that file](https://github.com/apache/thrift/blob/master/lib/php/lib/Transport/THttpClient.php#L178))

I've currently got that in-progress work to override the "flush" method and replace it entirely with the PSR-18/-7 implementation, which doesn't need to use it, so it's only the exceptions using it now.

I did notice that [Evernote's SDK's usage](https://github.com/evernote/evernote-sdk-php/blob/master/lib/Evernote/Client.php#L163) of THTTPClient went ahead and just inferred the ports from whether the scheme was http or https (so 80 or 443),.

I went ahead just now and did the same here since that seemed reasonable to me rather than requiring callers to specify it

File: src/Contrib/Jaeger/ThriftHttpSender.php (hunk 4 above)
Comment (Grunet) [reply src/Contrib/Jaeger/ThriftHttpSender.php]: Went ahead and merged what I had into this branch. Lmk if the approach needs any adjustment (it's light on automated test coverage/assertions for the moment. Probably going to wait to see what doing the final adapting work for that commented out block of code in the next PR will do before adding some more)

**Comment by Grunet** [comment]
Ok got the build to be green again (had to edit the psalm config to do so though oddly. Couldn't find a better way to have it stop complaining about THttpClient, which is from a dependency...)

@tidal lmk if I missed anything about your earlier comments. Looking forward to your review on the new changes as well!

File: src/Contrib/Jaeger/CustomizedTHttpClient.php [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+namespace OpenTelemetry\Contrib\Jaeger;
+
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Thrift\Exception\TTransportException;
+use Thrift\Factory\TStringFuncFactory;
+use Thrift\Transport\THttpClient;
+
+class CustomizedTHttpClient extends THttpClient
```
brettmc, line 14: I think this could just be named `HttpClient`, and you could add some docblock comments about why psr-18 was added to THttpClient...

src/Contrib/Jaeger/CustomizedTHttpClient.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $this->endpointUrl = $endpointUrl;
+
+        return $this;
+    }
+
+    /**
+     * Opens and sends the actual request over the HTTP connection
+     *
+     * @throws TTransportException if a writing error occurs
```
brettmc, line 55: I don't see any indication that an exception is thrown (and a psr-18 client would not throw a TTransportException)

src/Contrib/Jaeger/HttpCollectorExporter.php
tidal: @Grunet 
Yes, there is (and that' s what I actually meant to write above):
```php
$parsedDsn['path'] ?? ''
```
The ?? short hand acts like using an `isset` on the left hand operand.

tidal:
@Grunet 
Thanks for your work again.
Please, take a look at @brettmc 's commets, and then I think we are ready to merge this PR.
Otherwise we can still go back and make adjustments later.

tests/Unit/Contrib/JaegerSpanConverterTest.php
tidal: @grunet
My comment was more about readabilty. I believe it should be instantly understandable what a variable (or anything else)
means and that's often hard enough with English being such an ambiguous language (at least compared to German).
The point is, code is more often read than written, so anything that helps to easily understand the code is useful imo, even if it's often cumbersome to write something like `$value` instead of just `$v`

tidal (approved):

src/Contrib/Jaeger/HttpCollectorExporter.php
Grunet: Ahh neat! Updated. And remembering this one for sure now :)

src/Contrib/Jaeger/ThriftHttpSender.php
Grunet: Ah yeah I see what you mean. It looks like the $batch->write($this->protocol) line only ends up calling into the "write" method of THttpClient, so I think I should probably be able to copy that and give stub implementations for the other TTransport interface methods I think.

Ok went ahead and gave this a try, seems like it went fine

tests/Unit/Contrib/JaegerSpanConverterTest.php
Grunet: Yeah for sure. "convertedSpan" looked good to me so I went ahead and switched that in

src/Contrib/Jaeger/CustomizedTHttpClient.php
Grunet: Yeah for sure, I went ahead and named it similarly to ThriftUdpTransport (called it THriftHttpTransport) since both files are doing the same sort of thing now (implementing TTransport).

I also renamed ThriftHttpSender to just HttpSender so it doesn't seem like it's related to those 2 files directly since that was throwing me off a little

src/Contrib/Jaeger/CustomizedTHttpClient.php
Grunet: Yup, copy-pasta leftover lol. Removed

Grunet:
> @Grunet Thanks for your work again. Please, take a look at @brettmc 's comments, and then I think we are ready to merge this PR. Otherwise we can still go back and make adjustments later.

Ok think I got the build green again. Phan didn't seem to be able to trace that the host wasn't null once it got to the flush method, so I had to factor that out into the ParsedEndpointUrl helper (and also went ahead and refactored the UDP exporter to also use it, consolidating the validation logic a tad)

Otherwise I think I touched on everyone's comments. Lmk if I missed anything!

Issue #225, opened 2020-11-24
 User Research Feedback Improvements - Auto-detecting resource attributes


bobstrecansky:
We should autodetect resource attributes.

This ticket comes from [this comment](https://docs.google.com/document/d/1n43oTVsnydmbpzvhN-5dgPAvRWKZCiugmmR_wg_8ziQ/edit?disco=AAAAHTU6d48)

arminru:
Note that in addition to the user feedback you got having such a resource detector mechanism is also part of the spec:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#detecting-resource-information-from-the-environment

kishannsangani:
Hey @bobstrecansky, can you please assign this issue to me?

Fydon:
Hi. This issue has a wont fix label and the [compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md) shows this as not supported, linking to this issue. I take it from the pull request linked to this that it has been implemented?

kishannsangani:
Hey @Fydon, yes this has been implemented.

PR #582, opened 2022-02-16
Adding ResourceInfo attributes to Span attributes

kishannsangani:
PR for issue #225. Adding ResourceInfo attributes to Span attributes that can be exported to backends. Currently, there are a handful of ResourceInfo attributes that the Resource Detectors detect. Adding ResourceInfo attributes to Span attributes consumes the AttributeLimit. Adding more detectors or setting less AttributeLimit by the user could result In user attributes not getting set and reported.

### DIFF
--- a/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -72,6 +72,7 @@ class BatchSpanProcessor implements SpanProcessorInterface
      */
     public function onStart(ReadWriteSpanInterface $span, ?Context $parentContext = null): void
     {
+        $span->addResourceToAttributes();
     }
 
     /**
--- a/src/SDK/Trace/Span.php
+++ b/src/SDK/Trace/Span.php
@@ -412,6 +412,11 @@ final class Span extends API\AbstractSpan implements ReadWriteSpanInterface
         return $this->resource;
     }
 
+    public function addResourceToAttributes(): void
+    {
+        $this->setAttributes($this->resource->getAttributes());
+    }
+
     private function getImmutableAttributes(): AttributesInterface
     {
         if (null === $this->attributes) {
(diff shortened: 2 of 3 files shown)


brettmc:
@kishannsangani can you please look into failed build (static analysis is failing, they're likely all failing on the same issue). You can run this locally via `make all`.

kishannsangani:
@brettmc sure.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -72,6 +72,7 @@ public function __construct(
      */
     public function onStart(ReadWriteSpanInterface $span, ?Context $parentContext = null): void
     {
+        $span->addResourceToAttributes();
```
brettmc, line 75: I think that the span converters take care of this (a couple that I checked, in `src/contrib` do, anyway). So it looks to me like you just need to make sure that the detected resources (via `Composite` detector?) are added to the tracer provider, and as far as I can see they're passed on to a Tracer, then SpanBuilder, and finally to a Span.

kishannsangani: Detected resources are already present in tracer provider, but only as ResourceInfo. When we export on span end, we report only attributes of span, not the resource info. With this function addResourceToAttributes, I am adding attributes present in ResourceInfo to span attributes and calling it in onStart because ResourceInfo will be initialised with tracer.

brettmc: hmm, I thought that is what this line was doing? https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Otlp/SpanConverter.php#L176 - iterating over a span's resources (`ResourceInfo`), and adding them as attributes to a span.

kishannsangani: Yes, you are correct. This line does exactly that. But apart from Otlp SpanConverter, does this.

brettmc: It seems like different exporters have different requirements for how resources are represented, eg https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#resource - if that's the case, it probably needs to be managed in the individual span converters in a similar way to how OTLP does it.

PR #595, opened 2022-02-26
Add support for OTLP signal specific environment overrides [#590]

### LINKED ISSUE #590: Support Per-Signal Environment Variables in OTLP Exporters.
At the moment the OTLP [gRPC](https://github.com/open-telemetry/opentelemetry-php/blob/d2a90236325ea8ff6da5bf90d16e05e36ad0b8de/src/Contrib/OtlpGrpc/Exporter.php) and [HTTP](https://github.com/open-telemetry/opentelemetry-php/blob/d2a90236325ea8ff6da5bf90d16e05e36ad0b8de/src/Contrib/OtlpHttp/Exporter.php) Trace Exporters only allow for configuration via generic environment variables. 
(e.g. `OTEL_EXPORTER_OTLP_ENDPOINT`).
According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) : "Each configuration option MUST be overridable by a signal specific option." 
(e.g. `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`).

Configuration should first check for the existence of a signal (traces) specific environment variable, and in absence fall back to the respective generic environment variable.

Special attention must be given to the [modification rules for endpoint URLs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp).

yuktea:
Fixes #590 
Modified the OTLP gPRC and HTTP trace exporters to allow for configuration options overrides by signal specific options to comply with OpenTelemetry specs.

### DIFF
--- a/src/Contrib/OtlpGrpc/Exporter.php
+++ b/src/Contrib/OtlpGrpc/Exporter.php
@@ -11,7 +11,6 @@ use OpenTelemetry\Contrib\Otlp\SpanConverter;
 use Opentelemetry\Proto\Collector\Trace\V1\ExportTraceServiceRequest;
 use Opentelemetry\Proto\Collector\Trace\V1\TraceServiceClient;
 use OpenTelemetry\SDK\Common\Environment\KnownValues as Values;
-use OpenTelemetry\SDK\Common\Environment\Resolver as EnvResolver;
 use OpenTelemetry\SDK\Common\Environment\Variables as Env;
 use OpenTelemetry\SDK\Trace\Behavior\SpanExporterTrait;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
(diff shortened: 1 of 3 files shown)


tidal:
Thank you @yuktea!

LGTM.

Only one suggestion and this does not have any technical reason, but is about redabilty even more.
You could create a method `hasEnvironmentVariable` in the [EnvironmentVariablesTrait](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Common/Environment/EnvironmentVariablesTrait.php), which simply
calls said method.
The only reason is, that usually it's easier for the eye, if methods which are used in the same place look kind of similar.

That being said, the `EnvironmentVariablesTrait` may (or may not) be subject to change or deprecated, when Brett and me
pick up working on the general configuration again. The reason being, that knowledge about Env Vars or the existence
of those should not really be part of any component implementation. It goes a against Single Responsibilty Principle where
a class should only have one reason to change (and you can find more places in the library, where this could be improved)

So I leave it up to you, if you want to make that change. I would merge the PR as it is. 
Just tell us what you prefer. :)

brettmc (approved):

yuktea:
> 

Yes, I'll look into why they're failing

src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -48,24 +48,39 @@ public function __construct(
...
-            Env::OTEL_EXPORTER_OTLP_COMPRESSION,
-            $compression ? Values::VALUE_GZIP : Values::VALUE_NONE
-        );
-        $this->timeout = $this->getIntFromEnvironment(Env::OTEL_EXPORTER_OTLP_TIMEOUT, $timeout);
+
+        $this->insecure = $this->hasEnvironmentVariable(Env::OTEL_EXPORTER_OTLP_INSECURE) ?
+            $this->getBooleanFromEnvironment(Env::OTEL_EXPORTER_OTLP_INSECURE, $insecure) :
+            $this->getBooleanFromEnvironment(Env::OTEL_EXPORTER_OTLP_INSECURE, $insecure);
+
+        if (!empty($certificateFile) || EnvResolver::hasVariable(Env::OTEL_EXPORTER_OTLP_CERTIFICATE))
```
tidal, line 56: Could you use `hasEnvironmentVariable`here as well?

tidal:
There is a style issue now.
This is likely due to the `Resolver` class not used any more, but still being imported.
You can easily fix the style issue by running `make style`

src/Contrib/OtlpGrpc/Exporter.php
yuktea: Yes, not sure how I missed that one. Will make the changes

PR #600, opened 2022-03-01
Added OTEL_PHP_DETECTORS environment variable

### LINKED ISSUE #584: Add OTEL_PHP_DETECTORS environment variable
Add OTel PHP specific OTEL_PHP_DETECTORS environment variable to set up a given comma-separated list of detectors.

kishannsangani:
Resource detectors can be configured by an environment variable named `OTEL_PHP_DETECTORS`. By default, all resource detectors will be used to get resource info. The type of variable is List and known values are "all, env, host, os, process, process_runtime, sdk, sdk_provided, none". The order of priority is as follows:

1. If "all" is present, it will return resource info from all detectors and won't iterate the list of values.
2. If "all" is not present, it will iterate the list and add new instances of detectors(could raise multiple instances of the same detector but merge will handle duplicate resource info)
3. For "none", specific code is not written. If the list doesn't have "all" or any of the detectors in the list, it will return the equivalent of an empty resource.

"all" > any of the detectors > "none"

Fixes #584.

### DIFF
--- a/src/SDK/Resource/ResourceInfo.php
+++ b/src/SDK/Resource/ResourceInfo.php
@@ -4,8 +4,11 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\SDK\Resource;
 
+use function in_array;
 use OpenTelemetry\SDK\Attributes;
 use OpenTelemetry\SDK\AttributesInterface;
+use OpenTelemetry\SDK\Common\Environment\EnvironmentVariablesTrait;
+use OpenTelemetry\SDK\Common\Environment\Variables as Env;
 
 /**
  * A Resource is an immutable representation of the entity producing telemetry. For example, a process producing telemetry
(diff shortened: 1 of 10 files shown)


brettmc:
This looks great, thankyou! The only improvement I can think of is if you could also add an example to the `examples` dir showing how it can be used.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -51,17 +56,57 @@ public static function merge(ResourceInfo ...$resources): self
 
     public static function defaultResource(): self
     {
-        return (new Detectors\Composite([
-            new Detectors\Environment(),
+        $detectors = (new ResourceInfo(new Attributes()))->getListFromEnvironment(Env::OTEL_PHP_DETECTORS);
```
tidal, line 59: Could you use `Accessor::getList(Env::OTEL_PHP_DETECTORS)` here?
This method and Env Vars in general are static anyway, so not creating objects would be a bit more performant.

tidal:
Thank you @kishannsangani .
Just two little nit-picks from my side.

kishannsangani:
> This looks great, thankyou! The only improvement I can think of is if you could also add an example to the `examples` dir showing how it can be used.

Added an example file ResourceDetectors.php.

PR #601, opened 2022-03-02
Refactoring to use known values

kpratyus:
Replaced hard coded values with known values in ExporterFactory, SamplerFactory, SpanProcessorFactory

### DIFF
--- a/src/SDK/Trace/ExporterFactory.php
+++ b/src/SDK/Trace/ExporterFactory.php
@@ -9,6 +9,7 @@ use Nyholm\Dsn\Configuration\Dsn;
 use Nyholm\Dsn\Configuration\Url;
 use Nyholm\Dsn\DsnParser;
 use OpenTelemetry\SDK\Common\Environment\EnvironmentVariablesTrait;
+use OpenTelemetry\SDK\Common\Environment\KnownValues as Values;
 use OpenTelemetry\SDK\Common\Environment\Variables as Env;
 
 class ExporterFactory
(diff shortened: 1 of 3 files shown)


src/SDK/Trace/ExporterFactory.php
```diff
@@ -83,24 +84,24 @@ public function fromEnvironment(): ?SpanExporterInterface
...
-            case 'none':
+            case Values::VALUE_NONE:
                 return null;
-            case 'jaeger':
-            case 'zipkin':
-            case 'newrelic':
+            case Values::VALUE_JAEGER:
+            case Values::VALUE_ZIPKIN:
+            case Values::VALUE_NEWRELIC:
             case 'zipkintonewrelic':
```
brettmc, line 92: is there a reason this one was left untouched?

tidal:
Thanks @kpratyus !

If you also change the one value, which Brett mentioned, which should be good to merge.

src/SDK/Trace/ExporterFactory.php
kpratyus: The refactoring has been done for those variable which are already defined in the knownvalues inteface. 'zipkintonewrelic' is not defined there. The knownvalues interface only has variables which are defined in the OTEL spec -  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#exporter-selection. That's why it has been ommitted.

tidal: @kpratyus 
Ok, got you.
I did not remember, that I didn't add those values.
In general our "known values" could have more values than the ones stated in the spec.
However the exporter factory is subject to change in this case anyway as we may have to deal
with more vendor specific exporters in general, so we must have a more dynamic way than hard coding
all the options. Also actually users should be able to use thier custom exporters.

@brettmc 
So, I think we should just merge this as is, and take care of tis later (I ill keep in mind)
Do you agree?

tidal (approved):

src/SDK/Trace/ExporterFactory.php
brettmc: Yes, I'm happy to merge as is

PR #603, opened 2022-03-03
Creating Events and Listener API

kpratyus:
In this PR Event and Listener API has been created.
Steps for Adding Listeners -

- Event-Listener string pair should be added to Dispatcher class(The dispatcher class is a singleton class over the whole library)
- The Event Class should be added in the location SDK/Subscriber/Event and it should inherit the EventInterface.
- The Listener Class should be added in the location SDK/Subscriber/Listener and it should inherit the ListenerInterface.
- The return type of Event classes and the parameter of Listener classes are of type Array. The Array could contain objects like Span, Tracer etc and is fully configurable.


tidal:
Thank you @kpratyus 
I will take a look later.
However some tests are failing. It seems like its the same cause for all of those.
Could you have a look, please?

kpratyus:
> Thank you @kpratyus
> I will take a look later.
> However some tests are failing. It seems like its the same cause for all of those.
> Could you have a look, please?

Hey @tidal seems like one of the Unit tests is failing. While you are looking into the PR, in the meantime I will rectify the error.

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+    public function setListener(string $event, string $listener):void
+    {
+        self::$Listener[$event] = $listener;
+    }
+
+    public function dispatch(EventInterface $event):void
+    {
+        $eventClassname = get_class($event);
+        $eventClassname = str_replace('OpenTelemetry\\SDK\\Subscriber\\Event\\', '', $eventClassname);
+        $listenerClassName = self::$Listener[$eventClassname];
```
brettmc, line 34: there may be no listeners for an event, in which case this method should exit quickly

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function setListener(string $event, string $listener):void
+    {
+        self::$Listener[$event] = $listener;
+    }
+
+    public function dispatch(EventInterface $event):void
+    {
+        $eventClassname = get_class($event);
+        $eventClassname = str_replace('OpenTelemetry\\SDK\\Subscriber\\Event\\', '', $eventClassname);
```
brettmc, line 33: I think it's better if we use `$event::class` as the key, so there can be no ambiguity between classes in different namespaces?

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+        $listenerClassName = self::$Listener[$eventClassname];
+        $listenerClassName = 'OpenTelemetry\\SDK\\Subscriber\\Listener\\' . $listenerClassName;
+        $listener = new $listenerClassName();
```
brettmc, line 36: I think the listener should be provided as an invokable or callable (it might be a class, or a function, or ...) rather than creating new instances on every event

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+        $listener->takeAction($event->getObject());
```
brettmc, line 37: `takeAction()` accepts an array, but is being given an object? I think a listener should receive the event itself, and be type-hinted as such. An array is ambiguous because it can hold pretty much anything.

src/SDK/Subscriber/ListenerInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber;
+
+interface ListenerInterface
+{
+    public function takeAction(array $array):void;
```
brettmc, line 9: `handle` might be a better method name

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+        if (self::$instance == null) {
+            self::$instance = new Dispatcher();
+        }
+
+        return self::$instance;
+    }
+
+    public function setListener(string $event, string $listener):void
+    {
+        self::$Listener[$event] = $listener;
```
brettmc, line 27: do we want to allow for multiple listeners for an event?

src/SDK/Subscriber/Listener/EndSpanListener.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
brettmc, line 1: these example listeners shouldn't be in the SDK (unless you plan to expand on them to provide built-in functionality).

src/SDK/Subscriber/Dispatcher.php
tidal: The `class` constant is not available on objects in php 7.4, so it should simply be `get_class` or the event class should have a `name` parameter/method.

src/SDK/Subscriber/Event/EndSpanEvent.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber\Event;
+
+use OpenTelemetry\SDK\Subscriber\EventInterface;
+
+class EndSpanEvent implements EventInterface
+{
+    private array $array ;
```
tidal, line 11: Please use a different name that is more descriptive.
Naming things in software development is hard, but naming things after their type doesn't tell what it's supposed to do.

src/SDK/Subscriber/Event/EndSpanEvent.php
```diff
@@ -0,0 +1,21 @@
...
+    public function __construct(array $array)
+    {
+        $this->array = $array;
+    }
+
+    public function getObject():array
```
tidal, line 17: This method is called `getObject` but is returning an array.

src/SDK/Subscriber/Event/StartSpanEvent.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Subscriber\Event;
+
+use OpenTelemetry\SDK\Subscriber\EventInterface;
+
+class StartSpanEvent implements EventInterface
+{
+    private array $array ;
```
tidal, line 11: Same as above

src/SDK/Subscriber/Event/StartSpanEvent.php
```diff
@@ -0,0 +1,21 @@
...
+    public function __construct(array $array)
+    {
+        $this->array = $array;
+    }
+
+    public function getObject():array
```
tidal, line 17: Same as above

src/SDK/Subscriber/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+    public static function getInstance()
+    {
+        if (self::$instance == null) {
+            self::$instance = new Dispatcher();
+        }
+
+        return self::$instance;
+    }
+
+    public function setListener(string $event, string $listener):void
```
tidal, line 25: `listen` would be a better name for the method than `setListener` imo.

src/SDK/Subscriber/Dispatcher.php
tidal: That's would be expected bahavior for event listeners, however technically you can chain a Listener, which in turn could have multiple Subscribers and fan out like that.

src/EventHandler/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+        if (self::$instance == null) {
+            self::$instance = new Dispatcher();
+        }
+
+        return self::$instance;
+    }
+
+    public function listen(string $event, callable $listener):void
+    {
+        self::$listeners[$event] = $listener;
```
brettmc, line 27: this will still only allow one listener per event... I think the method should be `addListener` in that case

src/EventHandler/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public function dispatch(EventInterface $event):void
+    {
+        $eventClassname = $event->getClassName();
+        $listenerCallable = self::$listeners[$eventClassname];
```
brettmc, line 33: with multiple listeners (or none) this should iterate over all of them

src/EventHandler/EventInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\EventHandler;
+
+interface EventInterface
+{
+    public function getArray():array;
+
+    public function getClassName():string;
```
brettmc, line 11: maybe `getDescription` ? What is it used for?

src/EventHandler/EventInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\EventHandler;
+
+interface EventInterface
+{
+    public function getArray():array;
```
brettmc, line 9: This is not descriptive enough. What is it actually getting?

src/EventHandler/Events.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\EventHandler;
+
+interface Events
+{
+    public const START = 'StartSpanEvent';
```
brettmc, line 9: needs to be more specific, probably `SPAN_START` and `SPAN_END` - other things can start and end, and we might want to have listeners on those events, too, in the future.

examples/EventListenerExample.php
```diff
@@ -0,0 +1,58 @@
...
+use GuzzleHttp\Psr7\HttpFactory;
+use OpenTelemetry\Contrib\Zipkin\Exporter as ZipkinExporter;
+use OpenTelemetry\EventHandler\Dispatcher;
+use OpenTelemetry\EventHandler\Events;
+use OpenTelemetry\SDK\Attributes;
+use OpenTelemetry\SDK\Trace\SpanProcessor\SimpleSpanProcessor;
+use OpenTelemetry\SDK\Trace\TracerProvider;
+
+//Register the event listener pair, The dispatcher here is a global Class
+Dispatcher::getinstance()->listen(Events::START, 'OpenTelemetry\EventHandler\Listener\StartSpanListener::handle');
```
brettmc, line 16: since those listeners shouldn't be in the SDK, perhaps a better example would be to just pass in a callable (an inline function) which shows it actually changing the span in some way.

examples/EventListenerExample.php
```diff
@@ -0,0 +1,58 @@
...
+    new HttpFactory()
+);
+$tracerProvider =  new TracerProvider(
+    new SimpleSpanProcessor(
+        $zipkinExporter
+    )
+);
+$tracer = $tracerProvider->getTracer();
+
+echo 'Starting AlwaysOnZipkinExample';
```
brettmc, line 33: I think the example can be much simpler. A single span, using the console exporter (more like the Getting Started example)

src/EventHandler/Listener/StartSpanListener.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
brettmc, line 1: This file must not be part of the final PR, it looks like testing/evaluation code. We 100% do not want this random attribute added to all spans

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -72,6 +75,8 @@ public function __construct(
      */
     public function onStart(ReadWriteSpanInterface $span, ?Context $parentContext = null): void
     {
+        $event = new StartSpanEvent([$span]);
```
brettmc, line 78: Why is the span in an array? Wouldn't it make sense for this event to accept a `SpanInterface` (or even `WritableSpanInterface`)? Then it's really clear what you're working with (and IDEs will be happy)

src/EventHandler/Dispatcher.php
kpratyus: I will add the functionality for multiple listeners. @tidal suggested the method name as "listen".

src/EventHandler/EventInterface.php
kpratyus: The event and it's corresponding listeners have been stored as map of Event name as string key and Listener method callable as value so the getClassname returns the event name whenever the listener value has to be chosen from map.

src/EventHandler/Events.php
kpratyus: Yes the SPAN_START makes sense. 'StartSpanEvent' is the value which is to be returned by getClassName

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
kpratyus: - All the Event classes should Ideally Inherit an EventInterface so the getArray(getObject getSpan or whichever name we wanna use for the method) should have the same return type which can be inherited by all the Events. It might be that a particular Event wants to return 'Tracer' then it won't be able be able to implement the interface.
- In some case there may be a requirement of returning more than one object
Considing these 2 cases I thought it was best to return Array. What is your opinion?

src/EventHandler/Listener/StartSpanListener.php
kpratyus: Yeah, makes sense. Where do we include the listeners as an example?

examples/EventListenerExample.php
kpratyus: Okay

src/EventHandler/Listener/StartSpanListener.php
brettmc: Just create some inline functions in the example php file - I gave an example of doing that in a comment the other day.

src/EventHandler/EventInterface.php
brettmc: I think the map only needs to be the event name, not the class name? If each event accepts a callable, then it can be just called (eg by `call_user_func()`)

kpratyus: So you mean that the callable is stored in the event class only?

examples/EventListenerExample.php
kpratyus: But, it would lead to coupling of Listener with Dispatcher right? Isn't the better way to implement is keeping listener logic separate?

src/EventHandler/Listener/StartSpanListener.php
kpratyus: As, I mentioned above wouldn't it couple listener with dispatcher?

examples/EventListenerExample.php
brettmc: I don't think it does lead to any coupling. The dispatcher's listen event accepts a callable, and in this example, a callable is passed which demonstrates some dummy functionality. Nobody would use this example as-is, it's just illustrating how you can plug in functionality via an event.

src/EventHandler/EventInterface.php
brettmc: No, the callable would be stored in the dispatcher, against the event type. By default, the dispatcher has no listeners registered against any events (but maybe some default ones in future). Then, users can register one or more listeners for each event, providing the event handling logic as they do. Remember that one of the main benefits of events is that is allows for extensibility - we are giving users the ability to attach their own custom functionality.

src/EventHandler/Listener/StartSpanListener.php
brettmc: Given that the dispatcher will accept any callable as an event handler, I think that is very loose coupling. Are we talking about the same thing?

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
brettmc: Have you looked at what other event-based implementations do? Most of my experience with events is with zf2/laminas - they use `target` to refer to the thing the event is operating on (and the type would probably be `mixed`).
It might help you to come up with a solid solution if you can come up with a concrete use-case for having events, and ensure that it can be supported. Once events are implemented, what problem would you like to use events to solve?

examples/EventListenerExample.php
tidal: I agree with @brettmc .
A [Callable](https://www.php.net/manual/en/language.types.callable.php) would not lead to any coupling on the contrary. It would allow for a Listener to be a class or a simple lambda function, whatever suits the needs of the users best.

src/EventHandler/EventInterface.php
tidal: I agree. An array is a type and this is just not describing the kind of data it accesses.
So this could for example  be called `getMetaData` or `getHeaders` depending on the actual usage.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -94,6 +99,8 @@ public function onEnd(ReadableSpanInterface $span): void
         if ($this->bufferReachedExportLimit() || $this->enoughTimeHasPassed()) {
             $this->forceFlush();
         }
+        $event = new EndSpanEvent([$span]);
+        Dispatcher::getinstance()->dispatch($event);
```
tidal, line 103: This can be one call, as there is no need for a temporary variable:
```php
Dispatcher::getinstance()->dispatch(new EndSpanEvent([$span]));
```

src/SDK/Trace/SpanProcessor/SimpleSpanProcessor.php
```diff
@@ -38,6 +43,9 @@ public function onEnd(ReadableSpanInterface $span): void
         if (null !== $this->exporter) {
             $this->exporter->export([$span->toSpanData()]);
         }
+
+        $event = new EndSpanEvent([$span]);
+        Dispatcher::getinstance()->dispatch($event);
```
tidal, line 48: see above

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
brettmc: Good point. In that case, what about type-hinting only in phpdoc, and not adding a php type-hint (which we tidy up in the future when 7.4 support is dropped)

brettmc: I think that `object` is the right type to type-hint to. `mixed` can include things like resources, integers, booleans etc which I wouldn't expect to be the targets of an event

kpratyus:
@tidal @brettmc I have made the changes you suggested(Except the part where more than 1 listener is supported, I will add it in another commit). Could you review it once more?

examples/EventListenerExample.php
brettmc: this still needs to be actioned (but, the listener bits are looking good)

brettmc: still needs to be actioned... doesn't need zipkin, update message

examples/EventListenerExample.php
```diff
@@ -0,0 +1,64 @@
...
+
+$root = $span = $tracer->spanBuilder('root')->startSpan();
+$span->activate();
+
+for ($i = 0; $i < 3; $i++) {
```
brettmc, line 44: not required for this example

src/EventHandler/Dispatcher.php
brettmc, line 27: only supports one listener per event, needs to support multiple

src/EventHandler/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $eventName = $event->getEventName();
+        $listenerCallable = self::$listeners[$eventName];
+        if ($listenerCallable == null) {
+            return;
+        }
+        call_user_func($listenerCallable, $event);
+    }
+
+    public function getListener(string $event):string
```
brettmc, line 40: is this method needed?

src/EventHandler/EventInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\EventHandler;
+
+interface EventInterface
+{
+    public function getTarget():object;
+
+    public function getEventName():string;
```
brettmc, line 11: this should use the constants from Events class

src/EventHandler/Dispatcher.php
```diff
@@ -0,0 +1,44 @@
...
+
+    public function listen(string $event, callable $listener):void
+    {
+        self::$listeners[$event] = $listener;
+    }
+
+    public function dispatch(EventInterface $event):void
+    {
+        $eventName = $event->getEventName();
+        $listenerCallable = self::$listeners[$eventName];
```
brettmc, line 33: iterate over multiple listeners

bobstrecansky:
@kishannsangani - do you have everything you need for this PR?

kishannsangani:
> @kishannsangani - do you have everything you need for this PR?

Hey @bobstrecansky this PR belongs to @kpratyus and not me 😅

bobstrecansky:
Oh my goodness sorry @kishannsangani !  I was making this comment during our SIG meeting and my autocomplete must have filled your name in incorrectly.  Sorry about that!

PR #626, opened 2022-03-21
Resource Detector Unit Tests

### LINKED ISSUE #622: Add tests for Resource Detectors
At the moment there are no Unit Tests for the [Resource Detectors](https://github.com/open-telemetry/opentelemetry-php/tree/main/src/SDK/Resource/Detectors).

kishannsangani:
Closes #622.

Added separate unit tests files for Resource Detectors. These tests highly overlap with tests in ResourceTest.php but they were not included in the coverage.

### DIFF
--- a/tests/Unit/SDK/Resource/Detectors/ConstantTest.php
+++ b/tests/Unit/SDK/Resource/Detectors/ConstantTest.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tests\Unit\SDK\Resource\Detectors;
+
+use OpenTelemetry\SDK\Common\Attribute\Attributes;
+use OpenTelemetry\SDK\Resource\Detectors;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers OpenTelemetry\SDK\Resource\Detectors\Constant
+ */
+class ConstantTest extends TestCase
+{
+    public function test_constant_get_resource_with_empty_resource(): void
+    {
+        $resouceDetector = new Detectors\Constant(ResourceInfo::create(new Attributes()));
+        $resource = $resouceDetector->getResource();
+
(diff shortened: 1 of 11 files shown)


tidal:
@kishannsangani 
Thank you, I will take a look.



> These tests highly overlap with tests in ResourceTest.php but they were not included in the coverage.

This is due to the fact that ResourceTest.php is not a Unit Test for ResourceInfo but a Integration Test and should not be placed under Unit Tests in the first place. As we dicussed on slack, I will create a separate issue for this.

tests/Unit/SDK/Resource/Detectors/ConstantTest.php
```diff
@@ -0,0 +1,36 @@
...
+ * @covers OpenTelemetry\SDK\Resource\Detectors\Constant
+ */
+class ConstantTest extends TestCase
+{
+    public function test_constant_get_resource_with_empty_resource(): void
+    {
+        $resouceDetector = new Detectors\Constant(ResourceInfo::create(new Attributes()));
+        $resource = $resouceDetector->getResource();
+
+        $this->assertNull($resource->getSchemaUrl());
```
tidal, line 22: See above regarding schema url, but again this doesn't need to be addressed in this PR, as I'm not sure about the purpose of this detector in the first place.

tests/Unit/SDK/Resource/Detectors/EnvironmentTest.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public function test_environment_get_resource_with_service_name(): void
+    {
+        $this->setEnvironmentVariable('OTEL_SERVICE_NAME', 'test-service');
+
+        $resouceDetector = new Detectors\Environment();
+        $resource = $resouceDetector->getResource();
+
+        $this->assertSame(ResourceAttributes::SCHEMA_URL, $resource->getSchemaUrl());
```
tidal, line 54: This is redundant, as you are testing for it in above test.

tidal:
Just some nit-picks and comments from my side.

Issue #445, opened 2021-10-09
Support `schemaUrl`


Blacksmoke16:
See https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/trace/api.md#get-a-tracer and https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md.

_Originally posted by @Blacksmoke16 in https://github.com/open-telemetry/opentelemetry-php/pull/439#discussion_r723821913_

tidal:
What kind of schema is this actually referring to? I could not find any answer to that in the spec or docs (or I'm blind).

Blacksmoke16:
@tidal Mm yea, the spec doesn't make it super obvious what it is :/ I updated the OP to include some more context.

tidal:
Since semantic conventions landed in https://github.com/open-telemetry/opentelemetry-php/pull/502, the respective constants can be used to implement this feature.

brettmc:
It looks like this is partially addressed in #537

tidal:
@stale
真的吗？

yuktea:
#537, #607 and #627 have addressed parts of this PR. Can someone confirm if the issue should be closed or give some context if we still have something to cover?

Issue #621, opened 2022-03-19
Implement factory for ResourceInfo


tidal:
At the moment most of the methods in [ResourceInfo](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Resource/ResourceInfo.php) are factory methods. To better comply with [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) those methods should be extracted into a factory class.

kishannsangani:
@tidal I would like to work on this. Can you help me?

tidal:
@kishannsangani 
Sure, we'll talk on slack.

PR #627, opened 2022-03-21
Resolve schemaUrl when merging ResourceInfo

lalex:
Continue with the #445. This PR to resolve a resulting `schemaUrl` when merging `ResourceInfo`s. Reference to [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/441bafa8eda1697c3f0b975250e421e25da496b8/specification/resource/sdk.md#merge).

### DIFF
--- a/src/SDK/Resource/ResourceInfoFactory.php
+++ b/src/SDK/Resource/ResourceInfoFactory.php
@@ -24,12 +24,12 @@ class ResourceInfoFactory
     public static function merge(ResourceInfo ...$resources): ResourceInfo
     {
         $attributes = [];
-        $schemaUrl = null;
 
         foreach ($resources as $resource) {
             $attributes += $resource->getAttributes()->toArray();
-            $schemaUrl ??= $resource->getSchemaUrl();
         }
+
+        $schemaUrl = self::mergeSchemaUrl(...$resources);
 
         return ResourceInfo::create(new Attributes($attributes), $schemaUrl);
     }
@@ -92,5 +92,19 @@ class ResourceInfoFactory
     public static function emptyResource(): ResourceInfo
     {
         return ResourceInfo::create(new Attributes());
+    }
+
+    private static function mergeSchemaUrl(ResourceInfo ...$resources): ?string
+    {
+        $schemaUrl = null;
+        foreach ($resources as $resource) {
+            if ($schemaUrl !== null && $resource->getSchemaUrl() !== null && $schemaUrl !== $resource->getSchemaUrl()) {
+                // stop the merging if non-empty conflicting schemas are detected
+                return null;
+            }
+            $schemaUrl ??= $resource->getSchemaUrl();
+        }
+
+        return $schemaUrl;
     }
 }
--- a/tests/Unit/SDK/Resource/ResourceInfoTest.php
+++ b/tests/Unit/SDK/Resource/ResourceInfoTest.php
@@ -242,6 +242,29 @@ class ResourceInfoTest extends TestCase
         $this->assertEquals('', $empty);
     }
 
+    /**
+     * @dataProvider schemaUrlsToMergeProvider
+     */
+    public function test_merge_schema_url(array $schemaUrlsToMerge, ?string $expectedSchemaUrl): void
+    {
+        $resourcesToMerge = [];
+        foreach ($schemaUrlsToMerge as $schemaUrl) {
+            $resourcesToMerge[] = ResourceInfo::create(new Attributes([]), $schemaUrl);
+        }
+        $result = ResourceInfoFactory::merge(...$resourcesToMerge);
+
+        $this->assertEquals($expectedSchemaUrl, $result->getSchemaUrl());
+    }
+
+    public function schemaUrlsToMergeProvider()
+    {
+        yield 'Should keep old schemaUrl when the updating one is empty' => [['http://url', null], 'http://url'];
+        yield 'Should override empty old schemaUrl with non-empty updating one' => [[null, 'http://url'], 'http://url'];
+        yield 'Should keep matching schemaUrls' => [['http://url', 'http://url'], 'http://url'];
+        yield 'Should resolve an empty schemaUrl when the old and the updating are not equal' => [['http://url-1', 'http://url-2'], null];
+        yield 'Should keep empty schemaUrl when not equal schemas have been merged before' => [['http://url-1', 'http://url-2', 'http://url-2'], null];
+    }
+
     public function test_immutable_create(): void
     {
         $attributes = new Attributes();


src/SDK/Resource/ResourceInfo.php
```diff
@@ -42,15 +45,15 @@ public static function create(AttributesInterface $attributes, ?string $schemaUr
      */
     public static function merge(ResourceInfo ...$resources): self
```
tidal, line 46: Please note that this method is about to be extracted out.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -42,15 +45,15 @@ public static function create(AttributesInterface $attributes, ?string $schemaUr
...
-            $schemaUrl ??= $resource->getSchemaUrl();
+        if (empty($resources)) {
+            return new ResourceInfo(new Attributes([]), null);
         }
 
-        return new ResourceInfo(new Attributes($attributes), $schemaUrl);
+        $initialResource = array_shift($resources);
+
+        return array_reduce($resources, function (ResourceInfo $old, ResourceInfo $updating): ResourceInfo {
+            return $old->mergeWith($updating);
```
tidal, line 55: Calling the `mergeWith` will result in unnecessary object creation when used in `defaultResource`, however this could be optomized later.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -122,4 +125,25 @@ public function getSchemaUrl(): ?string
     {
         return $this->schemaUrl;
     }
+
+    private function mergeWith(ResourceInfo $resource): ResourceInfo
```
tidal, line 129: Please make this method public and add it to the interface, since static `merge`is subject to be moved.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -122,4 +125,25 @@ public function getSchemaUrl(): ?string
...
+    {
+        $attributes = $this->attributes->toArray() + $resource->getAttributes()->toArray();
+
+        $schemaUrl = null;
+        if (null === $this->schemaUrl) {
```
tidal, line 134: Please, do not use yoda style.
Please, use the accessor for this property here, since the accessor is used in the updatig resource and this will make it easier to read.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -122,4 +125,25 @@ public function getSchemaUrl(): ?string
...
+            $schemaUrl = $resource->getSchemaUrl();
+        } elseif (null === $resource->getSchemaUrl()) {
```
tidal, line 136: Please, do not use yoda style.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -122,4 +125,25 @@ public function getSchemaUrl(): ?string
...
+            $schemaUrl = $this->schemaUrl;
+        } elseif ($this->schemaUrl === $resource->getSchemaUrl()) {
+            $schemaUrl = $this->schemaUrl;
+        } else {
```
tidal, line 140: Please refactor these conditions in a method which simply resolves the schemaURL and remove else/elseif.

src/SDK/Resource/ResourceInfo.php
```diff
@@ -122,4 +125,25 @@ public function getSchemaUrl(): ?string
...
+    {
+        $attributes = $this->attributes->toArray() + $resource->getAttributes()->toArray();
+
+        $schemaUrl = null;
+        if (null === $this->schemaUrl) {
+            $schemaUrl = $resource->getSchemaUrl();
+        } elseif (null === $resource->getSchemaUrl()) {
+            $schemaUrl = $this->schemaUrl;
+        } elseif ($this->schemaUrl === $resource->getSchemaUrl()) {
+            $schemaUrl = $this->schemaUrl;
```
tidal, line 139: See above.

tests/Unit/SDK/Resource/ResourceTest.php
```diff
@@ -241,6 +243,32 @@ public function test_merge(): void
         $this->assertEquals('', $empty);
     }
 
+    /**
+     * @dataProvider schemaUrlsToMergeProvider
+     */
+    public function test_merge_schema_url($oldScheme, $updatingScheme, $expectedScheme, $shouldLog): void
+    {
+        if ($shouldLog) {
```
tidal, line 251: Please, do not use conditions in tests. This can simply be a separate test.

tests/Unit/SDK/Resource/ResourceTest.php
```diff
@@ -241,6 +243,32 @@ public function test_merge(): void
...
+            $loggerMock = $this->createMock(LoggerInterface::class);
+            $loggerMock->expects($this->once())->method('log');
+            LoggerHolder::set($loggerMock);
+        }
+
+        $primary = ResourceInfo::create(new Attributes([]), $oldScheme);
+        $secondary = ResourceInfo::create(new Attributes([]), $updatingScheme);
+        $result = ResourceInfo::merge($primary, $secondary);
```
tidal, line 259: Please test against the `mergeWith`method here. (See above)

tidal:
Thank you!

tidal (changes requested):

src/SDK/Resource/ResourceInfo.php
lalex: I'm not sure that `php-cs-fixer` is configured properly to avoid yoda style.

src/SDK/Resource/ResourceInfo.php
lalex: This piece of code is self-describing and follows [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/441bafa8eda1697c3f0b975250e421e25da496b8/specification/resource/sdk.md#merge). A simplified version could hide an original logic.

tests/Unit/SDK/Resource/ResourceTest.php
lalex: Good point, thanks!

lalex:
Let's postpone this PR until the `ResourceInfo` refactoring is finished in #621.

src/SDK/Resource/ResourceInfo.php
tidal: No it's not, as there is a lot of it in the code atm, which would needed to be changed (not sure if the fixer can handle this correctly).
However we have agreed in a meetig, that we'd rather want to avoid it in the future.

src/SDK/Resource/ResourceInfo.php
tidal: I wouldn't go for a simplified version, but rather avoiding the "IFs" with return statements.
I know it's to be found like that in the spec, which doesn't really make the spec better in my eyes.
However avoiding "IFs" is my personal issue, so I'd at least ask for it. Feel free to ignore my request. :)

tidal:
> Let's postpone this PR until the `ResourceInfo` refactoring is finished in #621.


> Let's postpone this PR until the `ResourceInfo` refactoring is finished in #621.

Ok.

kishannsangani:
> Let's postpone this PR until the `ResourceInfo` refactoring is finished in #621.

@lalex ResourceInfo refactoring is done in #628 and has been merged in the main branch.

bobstrecansky:
I think you should be able to continue work on this when you'd like to @lalex !

lalex:
I've decided to not to go too deep into the process of schema merging. The only significant thing is done is the setting an empty schemaUrl in case of conflict detected.
I think the PR is ready to be merged.

PS. I didn't find a proper way to write a log message from static context. @tidal, could you get a hint?

brettmc:
> PS. I didn't find a proper way to write a log message from static context.

Hi @lalex - I don't think logging messages from static methods has come up yet, but looking at `LogsMessagesTrait`, I suspect it can be changed to use static functions so that it works for both static and non-static use cases. I will have time to play around with that later today.

tidal (approved):

tidal:
@lalex  Thanks for your work!

> PS. I didn't find a proper way to write a log message from static context. @tidal, could you get a hint?

You could access the [LoggerHolder's get](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Common/Log/LoggerHolder.php#L25) method to get the logger as it's static anyway 
(The constructor is just there to help with dependendy injection in frameworks)

Anyway,I will merge the PR for now.

PR #634, opened 2022-03-26
Add compliance tests for environment variables

yuktea:
Added integration tests to verify compliance with the specification matrix.
Covers:
`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`
`OTEL_SPAN_EVENT_COUNT_LIMIT`

### DIFF
--- a/tests/Integration/SDK/SpanLimitsBuilderTest.php
+++ b/tests/Integration/SDK/SpanLimitsBuilderTest.php
@@ -2,18 +2,20 @@
 
 declare(strict_types=1);
 
-namespace OpenTelemetry\Tests\Integration\SDK\Trace;
+namespace OpenTelemetry\Tests\Integration\SDK;
 
 use AssertWell\PHPUnitGlobalState\EnvironmentVariables;
 use OpenTelemetry\SDK\Trace\SpanLimitsBuilder;
 use PHPUnit\Framework\TestCase;
 
-/**
- * @covers OpenTelemetry\SDK\Trace\SpanLimitsBuilder
- */
 class SpanLimitsBuilderTest extends TestCase
 {
     use EnvironmentVariables;
+
+    public function tearDown(): void
+    {
+        $this->restoreEnvironmentVariables();
+    }
 
     /**
      * @group trace-compliance


tidal:
Thank you @yuktea !
I will take a look.

tests/Integration/SDK/SpanLimitsBuilderTest.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tests\Integration\SDK\Trace;
```
tidal, line 5: The namespace is `OpenTelemetry\Tests\Integration\SDK\Trace`, but the directory is `tests/Integration/SDK/`
Please, change one or the other.

(This is not super important as the tests are not loaded/resolved via the namespace like library classes, but it makes for better consitency. It's also not super important which option you choose, as the integration tests do not necessarily have to match library classes)

tests/Integration/SDK/SpanLimitsBuilderTest.php
```diff
@@ -0,0 +1,63 @@
...
+use AssertWell\PHPUnitGlobalState\EnvironmentVariables;
+use OpenTelemetry\SDK\Trace\SpanLimitsBuilder;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers OpenTelemetry\SDK\Trace\SpanLimitsBuilder
+ */
+class SpanLimitsBuilderTest extends TestCase
+{
+    use EnvironmentVariables;
```
tidal, line 16: if you use this trait (or test for anything global like env vars) you should restore the state the state the process was, either before or after any test in that class.

PHPunit test are usually executed in one process (unless configured differently), however for every test in a class a new instance of this test class is created. 
The method `setUp` is executed before every test, the method `tearDown` after every test.
You can see how to reset the env vars using this trait here:
https://github.com/tidal/opentelemetry-php/blob/main/tests/Unit/Contrib/OTLPGrpcExporterTest.php#L30

tidal (changes requested):

tidal:
@yuktea 
LGTM, only a few nit-picks from my side.
(I have added some background info. If you already know this stuff, please just ignore it. I just think those things are not necessarily obvious or intuitive)

yuktea:
> @yuktea LGTM, only a few nit-picks from my side. (I have added some background info. If you already know this stuff, please just ignore it. I just think those things are not necessarily obvious or intuitive)

Thank you so much for the feedback @tidal, I'll take care of these things in the future.

tests/Integration/SDK/SpanLimitsBuilderTest.php
```diff
@@ -0,0 +1,63 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Tests\Integration\SDK\Trace;
+
+use AssertWell\PHPUnitGlobalState\EnvironmentVariables;
+use OpenTelemetry\SDK\Trace\SpanLimitsBuilder;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers OpenTelemetry\SDK\Trace\SpanLimitsBuilder
```
tidal, line 12: The `covers` annotation is also not really needed in the integration tests (and often the point of integration tests is to cover more than one class in conjunction)
We mainly use it in the unit tests to not have to create mocks for everything, because otherwise the code coverage report would make every code which is executed green, whether is expöicetly tested or not

PR #635, opened 2022-03-26
Fix span and trace ids to be the correct unsigned ints they need to be when converted for Thrift

### LINKED ISSUE #631: Jaeger thrift exporter exports spanIds as "7fffffff" for some IDs
The Jaeger `SpanConverter` converts the hexadecimal `spanId` to an integer [here](https://github.com/open-telemetry/opentelemetry-php/blob/2b72c92fcb30ed241c64c2bea613e7d3bb6ac7cb/src/Contrib/Jaeger/SpanConverter.php#L103), which is then, according to the [thrift span specification](https://github.com/open-telemetry/opentelemetry-php/blob/2b72c92fcb30ed241c64c2bea613e7d3bb6ac7cb/thrift/jaeger/SpanRef.php#L42), converted to a 64-bit **signed** integer.

The problem is that intval will interpret the hexadecimal `spanId` as an **unsigned** integer. Thus any ID that starts with `8` or greater will get capped at the maximum 64-bit signed integer value of `7FFFFFFF`.

I guess this also affects `traceId`s.

Grunet:
Resolves #631 or at least attempts to...

FYI I still haven't functionally tested any of the Jaeger changes yet, so this is mostly going off of the issue discussion + a re-reading of the spec section (mentioned in a code comment) and re-visiting my understanding of how Go and Java handles this


Grunet:
Can't quite tell what coverage drop codecov is pointing to in the SpanConverter

src/Contrib/Jaeger/IdConverter.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return self::convertHexStringToSignedInt($spanId);
+    }
+
+    private static function convertHexStringToSignedInt(string $hexString): int
+    {
+        //This is using high precision arithmetic because PHP by default won't do subtraction correctly for large integers (e.g. (2^63 - 1) - (2^63))
+        $hexStringAsInteger = BigInteger::fromBase($hexString, 16);
+
+        $shiftedInteger = $hexStringAsInteger->minus(BigInteger::fromBase('8000000000000000', 16)); //The number being subtracted is 2^63
```
brettmc, line 36: could it be written as `pow(2, 63)` for ease of reading?

Grunet: I updated the comments around this to try and clarify, but I don't think integers in PHP can support 2^63, so pow(2,63) probably won't work directly.

There may be other ways to represent it using the brick\math library though but nothing is catching my eye from another look at their READme - https://github.com/brick/math

DrLuke: Maybe defining a constant would be appropriate here?

Grunet: Yeah that seems fine to me, updated.

Not in love with my initial naming so all suggestions are welcome

brettmc: (I'm learning about big numbers as I go, here)... So, it's really `PHP_INT_MAX+1`, right?
```php
php > var_dump(dechex(PHP_INT_MAX+1));
string(16) "8000000000000000"
```
Is it reliable/efficient to just calculate it, as above? As to constant name, `2_POW_63_HEX` ? Too brief?

Grunet: Yup about PHP_INT_MAX+1 (for 64-bit which we're requiring elsewhere iirc)

And whaaaa about that example and the [docs for dechex](https://www.php.net/manual/en/function.dechex.php) that say "The largest number that can be converted is PHP_INT_MAX * 2 + 1"!?! What does that even mean, like is PHP not representing PHP_INT_MAX * 2 + 1 as an integer or something? I wonder if it's getting converted to a float and then dechex is somehow still handling that passably 

(EDIT: yeah when I print PHP_INT_MAX in an online REPL I get 9223372036854775807 but when I print PHP_INT_MAX+1 I get 9.2233720368548E+18. And yeah when I print gettype(PHP_INT_MAX+1) it outputs "double", so that might be what's going on

So I guess it could be computed with pow(2,63), but that'll end up as a double and then we probably can't subtract properly between brick\math's BigInteger and BigDecimal representations (the latter which this would probably have to use if I had to guess) is what I would expect at least.

Also no preference on the constant name. Just lmk and I can switch whatever in.

src/Contrib/Jaeger/IdConverter.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Brick\Math\BigInteger;
+
+/**
+ * An implementation of the section of the spec around converting OTEL's span/trace id representations to Thrift's format
+ * https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids
+ */
+class IdConverter
+{
+    const HEX_STR_FOR_2_TO_THE_POWER_63 = '8000000000000000';
```
kishannsangani, line 15: How about the name `HEX_STR_FOR_2_POWER_63`?

kishannsangani (approved):

tests/Unit/Contrib/JaegerSpanConverterTest.php
```diff
@@ -29,10 +30,10 @@ public function test_should_convert_an_otel_span_to_a_jaeger_thrift_span()
 
         [$convertedSpan] = (new SpanConverter())->convert([$span]);
 
-        $this->assertSame(0, $convertedSpan->traceIdLow);
-        $this->assertSame(0, $convertedSpan->traceIdHigh);
-        $this->assertSame(0, $convertedSpan->spanId);
-        $this->assertSame(0, $convertedSpan->parentSpanId);
+        $this->assertSame(PHP_INT_MIN, $convertedSpan->traceIdLow); //For some reason hardcoding -9223372036854775808 (-2^63) instead of PHP_INT_MIN isn't liked by PHPUnit. Might be a high precision arithmetic issue?
```
brettmc, line 33: I don't think the comments need to stay. Using PHP_INT_MIN is more readable and less mysterious anyway.

kishannsangani: Found a [bug](https://bugs.php.net/bug.php?id=76385) reported in PHP. Quoting one comment from the bug:

`This bug is hilarious:
The problem is that PHP parses "-9223372036854775808" not as a negative number but as a "-" followed by "9223372036854775808". The latter value is, of course, just barely too large for an integer so it overflows to a float.`

bobstrecansky (approved):

bobstrecansky:
One merge conflict but I think we're off to the races after that's resolved.  Nice work!

tests/Unit/Contrib/JaegerSpanConverterTest.php
Grunet: Wow that's so interesting. Great find @kishannsangani !

And went ahead and removed the comments

src/Contrib/Jaeger/IdConverter.php
Grunet: Yeah no preferences here. If anyone has a strong preference before this is merged I'll swap it in

Grunet:
@bobstrecansky yup np. Think I was eventually able to take care of the merge conflict so should be good now

Grunet:
Oh lol I may've just reset everyone's reviews, sorry about that

bobstrecansky:
No problem!

PR #642, opened 2022-03-30
Fix #641

### LINKED ISSUE #641: TraceId mismatches between otel library & jager collector when request processed through JaegerAgentExporter Thrift using UDP
**Describe your environment** 
This issue persists irrespective of the php -v and environment. Verified using main branch as well.

**Steps to reproduce**
By using jaegertracing/all-in-one image we can spin up jager agent & collector locally. 
Then use the otel library to create a span that will have traceId and SpanId with 32 char hex string. 
While using JaegerAgentExporter to send the traceId and spanId along with parent spanId to jager agent container by converting the traceId of 36 char hex string to int64 (as Jager Thrift accepts int64 only) the conversion here (open-telemetry/opentelemetry/src/Contrib/Jaeger/IdConverter.php) doesn't go well and end up sending a wrongly converted int64 to jaeger agent which eventually gets converted to hex while getting stored in jager collector with a different 32 char hex which has no reference of the otel generated traceId. 

Ideally, otel lib generated traceId, spanId & parentSpanId should match with the  traceId, spanId & parentSpanId stored in jager collector while getting passed through Jager Agent UDP exporter like it is happening for the HTTP exporter.

otel lib generated traceId, spanId  doesn't match with the jager collector's traceId and spanId

**Additional context**
I believe we need to work on convertHexStringToSignedInt method on open-telemetry/opentelemetry/src/Contrib/Jaeger/IdConverter.php to pass correctly  converted int64 so when jager agent convert this to hex and stores in jager collector it would be the same which was generated by the otel library.

I tried a small example where I converted a 16 char hex to int64 and convert it back to hex string and it matches. (This is a potential solution to convert 16 char hex spanIds to int64 while passing to thrift, however this doesn't work with the 36 char hex traceId )

`<?php`
`// 16 char hex`
`$hex = bin2hex(random_bytes(8));`
 `echo $hex."\n";`

`$hi = intval(substr($hex, -16, -8), 16);`
`$lo = intval(substr($hex, -8, 8), 16);`

`$traceId = $hi << 32 | $lo;`

`// Converted to int 64`
`echo $traceId."\n";`

`// Converted back to same 16 char hex`
`echo dechex($traceId);`

saktib:
#641 
Now the traceId and spanId are matching, and the same can be used to search the traces from Jaeger UI. [Functional Test]

**traceId & spanId using otel lib** 
image

**traceId & spanId from Jaeger Collector** 
image

**Able to search with the traceId from Jaeger UI**
image


bobstrecansky:
I see a couple unit tests failing.

saktib:
> I see a couple unit tests failing.

@bobstrecansky I've fixed those now.

PR #645, opened 2022-03-31
Add test based on the Jaeger spec

Grunet:
Adds a specific test based on the spec example, since I tried at least twice at the logic and missed both times until @saktib fixed it all up...

Also fixes up some flakiness I noticed after running the tests a few times, like here https://github.com/open-telemetry/opentelemetry-php/runs/5763933166?check_suite_focus=true

### DIFF
--- a/tests/Unit/Contrib/IdConverterTest.php
+++ b/tests/Unit/Contrib/IdConverterTest.php
@@ -20,7 +20,7 @@ class IdConverterTest extends TestCase
         $this->assertEquals(-72057594037927936, IdConverter::convertOtelToJaegerSpanId($hex));
     }
 
-    public function test_correctly_converted_span_id()
+    public function test_correctly_converted_random_span_id()
     {
         // 16 char hex string
         $hex = bin2hex(random_bytes(8));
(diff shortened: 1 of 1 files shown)


tests/Unit/Contrib/IdConverterTest.php
```diff
@@ -12,6 +12,14 @@
  */
 class IdConverterTest extends TestCase
 {
+    //Based on this section of the Jaeger spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids
+    public function test_correctly_converts_example_from_spec()
+    {
+        $hex = 'FF00000000000000';
+
+        $this->assertEquals(-72057594037927936, IdConverter::convertOtelToJaegerSpanId($hex));
```
brettmc, line 20: do any of the other implementations have a set of well-known or edge inputs and their correct conversion? This feels like a good place to implement a phpunit `@dataProvider`

Grunet: I added the 1 (non-edge-case) test case from Go in. Let me take a look at some of the others to see if they have edge cases already.

Grunet: Didn't see any tests for this in Java or Python (maybe I wasn't searching in the right spots though). Found 1 case in JS but it's not an edge case - https://github.com/open-telemetry/opentelemetry-js/blob/a824578c981e02a762561db0b6bb44d24af1229d/packages/opentelemetry-exporter-jaeger/test/transform.test.ts#L180

brettmc (approved):

PR #655, opened 2022-04-03
Revert "replace E_USER_NOTICE with E_USER_DEPRECATED"

brettmc:
Reverts open-telemetry/opentelemetry-php#647
We were discussing this PR internally, and think that the original behaviour was correct and intentional - the BC layer is for the protection of code running in production mode, but not development mode.


brettmc:
@ivoba - you can catch one of us on slack if you'd like more details :)

bobstrecansky (approved):

PR #657, opened 2022-04-04
make Error level of BC messages configurable and add info to README.

tidal:


tests/Unit/SDK/Common/Dev/Compatibility/UtilTest.php
```diff
@@ -12,24 +13,86 @@
...
+        $this->assertSame(
+            $level,
+            Util::getErrorLevel()
+        );
+    }
+
+    /**
+     * @dataProvider errorLevelProvider
+     */
+    public function test_set_error_level_throws_exception_on_in_correct_level(int $level): void
```
brettmc, line 37: this doesn't use the data provider

brettmc (approved):
approved with one comment to look at

tests/Unit/SDK/Common/Dev/Compatibility/UtilTest.php
tidal: oops, you are right. That was a late and sloppy addition :)

PR #662, opened 2022-04-14
Add integration tests for tracing compliance

yuktea:
Some additional tests for the Span Builder SDK that verify:
1) Attributes are merged correctly
2) Setting attributes after calling `startSpan()` should not work
plus some additional tests for setting null and empty string values to attributes.

UPDATE: _I commented out one of the tests that I was not totally sure of. 
(unsure if `setAttribute()` should work with empty string `''` values.)_

### DIFF
--- a/tests/Integration/SDK/SpanBuilderTest.php
+++ b/tests/Integration/SDK/SpanBuilderTest.php
@@ -127,7 +127,7 @@ class SpanBuilderTest extends MockeryTestCase
 
         $span->end();
     }
-
+    
     public function test_add_link_truncate_link_attribute_value(): void
     {
         $maxLength = 25;
@@ -164,7 +164,10 @@ class SpanBuilderTest extends MockeryTestCase
 
         $span->end();
     }
-
+    
+    /**
+     * @group trace-compliance
+     */
     public function test_add_link_no_effect_after_start_span(): void
     {
         $spanBuilder = $this->tracer->spanBuilder(self::SPAN_NAME);
@@ -218,6 +221,9 @@ class SpanBuilderTest extends MockeryTestCase
         $span->end();
     }
 
+    /**
+     * @group trace-compliance
+     */
     public function test_set_attribute_after_end(): void
     {
         /** @var Span $span */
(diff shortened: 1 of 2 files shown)


bobstrecansky (approved):

tests/Integration/SDK/SpanBuilderTest.php
```diff
@@ -230,6 +236,62 @@ public function test_set_attribute_after_end(): void
...
     }
 
+    public function test_set_attribute_null_string_value(): void
+    {
+        /** @var Span $span */
+        $span = $this
+            ->tracer
+            ->spanBuilder(self::SPAN_NAME)
+            ->setAttribute('nil', null)
+            ->setAttribute('empty-string', '')
```
bobstrecansky, line 246: We were discussing whether or not an empty string should be an allowed attribute.  Not necessarily part of this PR, just something to raise.

tests/Integration/SDK/SpanBuilderTest.php
```diff
@@ -230,6 +236,62 @@ public function test_set_attribute_after_end(): void
...
+            ->startSpan();
+
+        $attributes = $span->toSpanData()->getAttributes();
+        $this->assertSame(1, $attributes->count());
+        $this->assertSame('', $attributes->get('empty-string'));
+        $this->assertNull($attributes->get('nil'));
```
brettmc, line 252: this is a little confusing - did it get the nil attribute, or was null returned as a default for an undefined attribute?

tests/Integration/SDK/SpanBuilderTest.php
```diff
@@ -230,6 +236,62 @@ public function test_set_attribute_after_end(): void
...
+
+        $span->end();
+    }
+
+    public function test_set_attribute_only_null_string_value(): void
```
brettmc, line 257: this test could have a more descriptive name (they can be really long, that doesn't matter). test_set_attribute_with_only_null_string_should_<what should it do?>

tests/Integration/SDK/SpanBuilderTest.php
```diff
@@ -303,6 +365,31 @@ public function test_set_attributes(): void
         $span->end();
     }
 
+    public function test_set_attributes_merge_attributes(): void
```
brettmc, line 368: can it have a more descriptive name?

brettmc:
any tests which are verifying behaviour from the spec compliance matrix need a group annotation, please

tests/Integration/SDK/SpanBuilderTest.php
yuktea: > this is a little confusing - did it get the nil attribute, or was null returned as a default for an undefined attribute?

It did not accept the nil attribute

tests/Integration/SDK/SpanBuilderTest.php
```diff
@@ -230,6 +242,71 @@ public function test_set_attribute_after_end(): void
...
+            ->startSpan();
+
+        $attributes = $span->toSpanData()->getAttributes();
+        $this->assertSame(2, $attributes->count());
+
+        $spanBuilder
+            ->setAttribute('bar1', 77);
+
+        $attributes = $span->toSpanData()->getAttributes();
+        $this->assertSame(2, $attributes->count());
```
tidal, line 305: See my comment in slack. This is not really testing if adding an attribute is not possible, but if it changes the number of attributes.
It would be nice, if you could change it, and also the other tests, which only test for attribute count.

tidal (approved):

tidal:
@yuktea
Thx! LGTM

Issue #619, opened 2022-03-19
Move InstrumentationLibrary key generation out of TracerProvider


tidal:
The generation of a key for a InstrumentationLibrary should be moved out of the [TracerProvider class](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Trace/TracerProvider.php#L69) into a utility class in `SDK/Common/Util`,  since this funtionality is not only required in the `Trace` component/signal.

- Move key generation logic in `TracerProvider::getTracer`  into a dedicated Util class

ditsuke:
@tidal I'd like to be assigned.

PR #663, opened 2022-04-14
Move InstrumentationLibrary key generation logic out of TracerProvider/SpanConverter [#619]

ditsuke:
PR for Issue #619.

Factors key generation logic for InstrumentationLibrary instances out to a helper method in a utils class, replacing generation in `TracerProvider` and `SpanConverter` with calls to the new method.

### DIFF
--- a/tests/Unit/SDK/Common/Instrumentation/KeyGeneratorTest.php
+++ b/tests/Unit/SDK/Common/Instrumentation/KeyGeneratorTest.php
@@ -12,7 +12,7 @@ use PHPUnit\Framework\TestCase;
  */
 class KeyGeneratorTest extends TestCase
 {
-    public function test_non_empty(): void
+    public function test_non_null(): void
     {
         $name = 'foo';
         $version = 'bar';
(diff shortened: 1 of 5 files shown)


ditsuke:
Could use some feedback on the new `Helpers` class, can we do a better name?

brettmc:
@ditsuke welcome aboard! re: class naming, I think `SDK\Common\Instrumentation` is a better location. As to class name, `KeyGenerator`? It will also need unit tests.

ditsuke:
> @ditsuke welcome aboard! re: class naming, I think `SDK\Common\Instrumentation` is a better location. As to class name, `KeyGenerator`? It will also need unit tests.

@brettmc thanks! I've moved and renamed the generator and added unit tests, PTAL

tests/Unit/SDK/Common/Instrumentation/KeyGeneratorTest.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public function test_null_version_schema(): void
+    {
+        $name = 'foo';
+
+        $key = KeyGenerator::generateInstanceKey($name, null, null);
+
+        $this->assertSame(sprintf('%s@unknown ', $name), $key);
+    }
```
tidal, line 33: Could you please also add a test for providing a `name`, `version` and no `schemaURL`?
This is only to assert, that key are different for all of these three variants.

tidal:
@ditsuke 
Thank you. LGTM
I have added one little comment for a test you could add, just to make sure.
However other than his, I think we can merge the PR, if you are ok with it.

tidal (approved):

ditsuke:
> @ditsuke Thank you. LGTM I have added one little comment for a test you could add, just to make sure. However other than t, I think we can merge the PR, if you are ok with it.

@tidal thanks for reviewing. I've added the third test with b84e3bb

ditsuke:
Thanks!

Issue #660, opened 2022-04-13
protobuf generation broken


brettmc:
Recently, our protobuf generation process (`make protobuf`) has stopped working, due to the image we use (socialpoint/protobuf-tools) no longer being available on docker hub.
It looks like https://docs.buf.build/introduction is the new way to generate APIs, so this could be investigated as a possible solution.

**Steps to reproduce**
Run `make protobuf`

**What is the expected behavior?**
Regenerated protobuf API files are produced

**What is the actual behavior?**
Failed to pull image

brettmc:
It looks like the opentelemetry-proto repo is already set up to generate language-specific protobuf code... `make gen-php` looks to generate sensible output, and it might be as simple as swapping our current build process over to that.

PR #665, opened 2022-04-25
Fixing protobuf generation and InstrumentationScope

brettmc:
switching over to using opentelemetry-proto's own mechanism for generating protobuf interfaces, and
updating interfaces to latest version which works with our grpc exporter (v0.14.0). A separate issue + PR will
be required to work out why grpc exporting doesn't work with v0.15.0+

fixes: #660


script/proto_gen.sh
```diff
@@ -1,31 +1,34 @@
...
+DESTINATION_DIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")")/../proto/otel
+GPBMETA_DIR="GPBMetadata"
+OTEL_DIR="Opentelemetry"
 REPO_DIR=opentelemetry-proto
 
-mkdir -p $DESTINATION_DIR
-rm -R ./$REPO_DIR
-git clone https://github.com/open-telemetry/$REPO_DIR
+cd "${DESTINATION_DIR}" || exit
+rm -rf ./${REPO_DIR}
```
bobstrecansky, line 8: probably want a check for something here.  If $`{REPO_DIR}` ends up being empty or being inadvertently set some other way......

script/proto_gen.sh
```diff
@@ -1,31 +1,34 @@
...
+      git checkout "${TAG}"
 
-    git checkout "${TAG}"
+      echo "${TAG}"
+  )
 
-    echo "$TAG"
+  echo "Generating protobuf files for version ${TAG} ..."
+  make gen-php
+  rm -rf ${GPBMETA_DIR} ${OTEL_DIR}
```
bobstrecansky, line 24: Same in re: `rm -rf`

script/proto_gen.sh
```diff
@@ -1,31 +1,34 @@
...
-protoc --proto_path=$REPO_DIR/ --php_out=$DESTINATION_DIR --grpc_out=$DESTINATION_DIR \
- --plugin=protoc-gen-grpc=usr/local/bin/grpc_php_plugin $(find $REPO_DIR/opentelemetry/proto -iname "*.proto")
-
-echo "$TAG" > $DESTINATION_DIR/VERSION
+echo "Copying generated source..."
+cp -r ${REPO_DIR}/gen/php/${GPBMETA_DIR} .
+cp -r ${REPO_DIR}/gen/php/${OTEL_DIR} .
 
+echo "Cleaning up..."
+rm -rf ./${REPO_DIR}
```
bobstrecansky, line 33: Ditto in re: `rm -rf`

script/proto_gen.sh
brettmc: It's set a couple of lines up, so I _think_ this is safe. I've confirmed that it's portable enough to be run from / or /scripts and still work...

brettmc:
setting this to draft, since I now think I can get it working with 0.16 - the issue looks to be the renaming of InstrumentationLibrary to Scope

bobstrecansky (approved):

Issue #671, opened 2022-05-04
tracing with async php runtimes


brettmc:
**Describe your environment**
Using an async (not-shared-nothing) PHP runtime (eg roadrunner, swoole, react) where a long-lived application is used to service multiple requests.
Since we use a single shared `ContextStorage`, a call to `$span->activate()` will set the active span in that single context storage. Then, any other requests executed in the same process/worker will be parented to that span, and attempts to set an active span will clobber any other request's active span.

**Steps to reproduce**
Any simple otel-instrumented PHP application which creates a span, when run in parallel in the one runtime.

**What is the expected behavior?**
Traces and spans should be correctly nested, as if they were run in an isolated process in a shared-nothing runtime (apache, fpm)

**What is the actual behavior?**
Generated traces and spans are randomly parented to each other in the order they were executed.

**Additional context**
Our usage of `Context::getCurrent()` looks to be the core of the problem, since it always operates on a single, static storage.

I've spent some time looking at this, and I think the path forward is to:
- allow the creation of individual storage (eg one per request/operation)
- default to using a static storage (as we do now, which is fine for apache/fpm)
- update classes which use `Context::getCurrent()` to instead use `$storage->current()`, and allow those classes/methods to accept storage as a class variable (or static method parameter), so that "get the current context" becomes "get the current context from this storage"

tidal:
That sounds like a good idea.
I would like if we also can make use of the strength of roadrunner, swoole, react, etc.
So what about we create adapter for the different runtimes?
For the runtime detectors, we just need to implement them once for the attributes,
and we could reuse the detectors.
Also we may introduce a `ContextProvider` so we have a consistent pattern across all the components.

brettmc:
It's an interesting idea to be able to detect a runtime, and ContextProvider might be a good abstraction to hide storage complexities, but I do think that `ContextStorage` is the key.
I've got some changes in a branch which work (can have multiple traces being created at the same time), but it's not the most elegant solution - I will submit as a draft PR to seek feedback. The idea is that a static default storage is used if you don't do anything (as happens now, so the API is unchanged for apache/fpm/etc).
If you do require concurrency, then you are responsible for managing that and keeping track of each request's context/storage/tracer, for example by:
- creating per-request storage, keeping track of it (eg in a request attribute), passing it to `spanBuilder` (which is what I have now); or perhaps
- creating per-request tracer with individual storage attached (not to spec?), keep track of it as above

[...]

PR #674, opened 2022-05-09
Concurrent traces

brettmc:
To allow concurrent traces to be generated without clobbering each other, introduce the concept of "default storage", and allow the span builder to accept a storage via `setStorage()`.
If users wish to use non-default storage, they will be responsible for creating that storage, and keeping a reference to it (for example, in a psr-7 request attribute), and then passing that storage on when creating future spans.

Fixes: #671


bobstrecansky (approved):

brettmc:
Converting back to draft while Nevay explores a more elegant solution which might replace this PR.

brettmc:
closing, replaced by #675

Issue #328, opened 2021-05-05
OTLP/GRPC exporter implementation - Missing retries on Exporter errors


bobstrecansky:
We're missing retries on Exporter errors: https://github.com/open-telemetry/opentelemetry-specification/blob/318745aaf6fb173eb5cbc887ab76796bd817b3d6/specification/protocol/exporter.md#retry

Grunet:
Here are some notes from my quick look at the current state of other client SDK implementations around this:

- [Go](https://github.com/open-telemetry/opentelemetry-go/pull/2095) is the only one I see evidence of having tried implementing this, but possibly only partially (i.e. maybe just for gRPC/http and traces?)
- [Java](https://github.com/open-telemetry/opentelemetry-java/issues/3271) is waiting on further clarification [around the spec](https://github.com/open-telemetry/opentelemetry-specification/issues/1742) before starting, in particular around how the backoff parameters should be made adjustable by consumers
    - Side note: per that spec thread, gRPC maybe has [built-in capacities for backoff](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/retry#BackoffExponentialWithJitter), but I'm not sure if Go or any others are currently relying on that 
- [.NET](https://github.com/open-telemetry/opentelemetry-dotnet/issues/1779)'s issue for this only seems to discuss gRPC support

And for PHP, I'm not sure how backoff strategies typically get implemented (in particular for a library that needs to be very generically reusable in various different frameworks, e.g. the reactive ones I've heard about) so if anyone happens to know of a really good example of this from somewhere else on the internet, feel free to link to it here

tidal:
This is not feasible as a generic feature of the exporter, as it pretty much depends on the [runtime](https://github.com/php-runtime/runtime), if this will create problems or not. So if anything this feature has to be turned on explicitly by the user/instrumentation.
[...]

PR #677, opened 2022-05-19
OTLP/GRPC exporter implementation - Missing Reries on Exporter errors

sanketmehta28:
**Is your feature request related to a problem? Please describe.**
The OTLP exporter specification says that [transient errors MUST be handled with a retry strategy](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry), where the retriable gRPC error codes are defined in the [OTLP spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-response).

Fixes: #328 

I think we should implement generic retry policy which can be utilized by all exporters

**Additional context**
In grpc, retry is supported out of the box using service config but in this PR I have tried to provide the generic solution which can be used by all the exporters irrespective of the transport method.


tidal:
@sanketmehta28 
Hi, thanks for your contribution and sorry for the late response.
Can you please fix the CS issues, so we can review the  PR?
(The easiest way is just to call `make style` in the repo root dir)

bobstrecansky:
There are some psalm errors appearing in CI

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +54,65 @@ public function export(iterable $spans): int
...
+        if ($this->retryPolicy === null) {
+            return $this->doExport($spans);
+        }
+
+        while ($shouldRetry) {
+            // find the delay time before retry (it is 0 for the first attempt)
+            $delay = $this->retryPolicy->getDelay($attempt++);
+            if ($attempt > 0) {
+                self::logDebug('Waiting for ' . $delay . ' seconds before retrying export');
+                sleep($delay);
```
brettmc, line 67: I don't think sleeping here is a good idea, because it will cause the application being instrumented to wait. A better way might be to keep track of spans to retry (eg in internal storage), and check for spans to retry as a side-effect of doing something else in the exporter. We could do a final check as part of exporter shutdown to see if any pending spans need to be exported

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,187 @@
...
+        $this->setJitter($jitter);
+        $this->retryableStatusCodes = $retryableStatusCodes;
+    }
+
+    public function shouldRetry(
+        int $attempt,
+        int $status,
+        ?RetryableExportException $exception = null
+    ): bool {
+        if ($attempt >= $this->maxAttempts && null !== $exception) {
```
brettmc, line 41: I think this method should either return true or false, not true/throw

sanketmehta28:
Hi, Can anyone help me resolve this code coverage error?

brettmc:
> Hi, Can anyone help me resolve this code coverage error?

It's complaining about a decrease in code coverage, specifically in `SpanExporterTrait` - you can `make test-coverage` locally, you will need to add/change some unit tests to ensure code coverage is not reduced by this PR.

sanketmehta28:
Just an explanation for not including fiber as of now:
currently we are supporting php 7.4 and 8.0 which does not support fiber out of the box. so to provide retry mechanism in those php versions we have to provide an implementation without fiber and separate implementation for php 8.1 onwards with fiber. so as of now it does not make sense to provide two separate implementation.

Also I have tried the default retry mechanism of GRPC which also blocks the exporter until retry times out or successful. .net otel module is using this same behaviour([PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2766)) so it will not make much difference if this generic retry mechanism blocks the exporter.

src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -69,6 +70,18 @@ public function __construct(
             $this->getIntFromEnvironment(Env::OTEL_EXPORTER_OTLP_TIMEOUT, $timeout);
 
         $this->setSpanConverter(new SpanConverter());
+        $this->setRetryPolicy(ExponentialWithJitterRetryPolicy::getDefault());
```
brettmc, line 73: the default retry policy has 5 max attempts - didn't we discuss in SIG that by default there would not be a retry policy (or a retry policy that didn't retry) ?

brettmc:
I think this needs an example of how to use and configure a retry policy, so that users can see it in action and understand how to configure (and disable/turn off)

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,184 @@
...
+namespace OpenTelemetry\SDK\Common\Retry;
+
+use InvalidArgumentException;
+use OpenTelemetry\SDK\Metrics\Exceptions\RetryableExportException;
+use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+
+class ExponentialWithJitterRetryPolicy implements RetryPolicyInterface
+{
+    private $maxAttempts;
+    private $initial_backoff;
```
brettmc, line 14:
```suggestion
    private $initialBackoff;
```

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        return $this->maxAttempts;
+    }
+
+    public function setMaxAttempts(
+        int $max_attempts = self::DEFAULT_MAX_ATTEMPTS
+    ) {
+        $this->checkArgument(
+            $max_attempts > 0,
+            sprintf('Max attempts value must be > 0 and of int type. "%s" is given.', $max_attempts)
```
brettmc, line 81: it can't not be `int` because of the type-hint on the parameter

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function setInitialBackoff(float $initial_backoff);
+
+    public function getMaxBackoff(): int;
+
+    public function setMaxBackoff(int $max_backoff);
+
+    public function getBackoffMultiplier(): float;
+
+    public function setBackoffMultipler(float $backoff_multiplier);
```
Nevay, line 57: Should remove (most) of the getters/setters as these values are specific to the exponential retry policy and not relevant to retry strategies in general.

sanketmehta28:
> 

Sure I will create one and add it

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: I have put this check for negative value actually. Not for double.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +55,73 @@ public function export(iterable $spans): int
...
+        if ($this->retryPolicy === null) {
+            return $this->doExport($spans); /** @phpstan-ignore-line */
+        }
+
+        while ($shouldRetry) {
+            // find the delay time before retry (it is 0 for the first attempt)
+            $delay = $this->retryPolicy->getDelay($attempt);
+            if ($attempt > 0) {
+                self::logDebug('Waiting for ' . $delay . ' seconds before retrying export');
+                usleep($delay * 1000);
```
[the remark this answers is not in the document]
sanketmehta28: Can you explain this in more details?

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: As per the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#retry), it is mandatory to implement exponential retry with jitter in otel exporter. That is the reason I have put all these values in RetryPolicyInterface.php and not in ExponentialRetry.php

sanketmehta28:
> I think this needs an example of how to use and configure a retry policy, so that users can see it in action and understand how to configure (and disable/turn off)

Hi @brettmc Sure I will create one and let you know

src/SDK/Trace/Behavior/SpanExporterTrait.php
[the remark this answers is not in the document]
sanketmehta28: Hi @Nevay : I have added the blocking scheduler as of now. We will add non blocking scheduler later on.

src/SDK/Common/Retry/RetryPolicyInterface.php
Nevay: I'm still in favor of removing the getters/setters to have less mutable state.
Additionally other exponential retry policies might not support all of these configurations.

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+     * @return bool
+     */
+    public function shouldRetry(
+        int $attempt,
+        int $status,
+        ?RetryableExportException $exception
+    ): ?bool;
+
+    /**
+     * Returns the time to wait in seconds.
```
Nevay, line 95: Should be milliseconds.

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+
+    public function setJitter(float $jitter);
+
+    public function setRetryableStatusCodes(array $statusCodes);
+
+    public function getRetryableStatusCodes(): array;
+
+    public function setDelayScheduler(SchedulerInterface $scheduler);
+
+    public function getDelayScheduler(): SchedulerInterface;
```
Nevay, line 130: Shouldn't be part of the retry policy, unless the retry policy performs the waiting instead of returning the delay.

src/SDK/Common/Time/BlockingScheduler.php
```diff
@@ -0,0 +1,27 @@
...
+namespace OpenTelemetry\SDK\Common\Time;
+
+use OpenTelemetry\SDK\Behavior\LogsMessagesTrait;
+
+final class BlockingScheduler implements SchedulerInterface
+{
+    use LogsMessagesTrait;
+    public function __construct()
+    {
+        self::logInfo('setting Blocking Delay scheduler');
```
Nevay, line 14: Should be removed.

src/SDK/Resource/Detectors/Sdk.php
```diff
@@ -33,7 +33,7 @@ public function getResource(): ResourceInfo
                 if (!InstalledVersions::isInstalled($package)) {
                     continue;
                 }
-                if (($version = InstalledVersions::getPrettyVersion($package)) === null) {
+                if (($version = InstalledVersions::getVersion($package)) === null) {
```
Nevay, line 36: Is this change intended?

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
             return SpanExporterInterface::STATUS_SUCCESS;
         }
 
-        return $this->doExport($spans); /** @phpstan-ignore-line */
+        // If retryPolicy is not set then just doExport once and return the status
+        if ($this->retryPolicy === null) {
+            self::logInfo('retry polict is null. Exporting the spans without retry' . PHP_EOL);
```
Nevay, line 61: Having an info log message for this on every export seems superfluous.

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,199 @@
...
+    public function getBackoffMultiplier(): float
+    {
+        return $this->backoffMultiplier;
+    }
+
+    public function setBackoffMultipler(
+        float $backoff_multiplier = self::DEFAULT_BACKOFF_MULTIPLIER
+    ) {
+        $this->checkArgument(
+            $backoff_multiplier > 0,
```
Nevay, line 135: `> 1`

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,199 @@
...
+
+namespace OpenTelemetry\SDK\Common\Retry;
+
+use InvalidArgumentException;
+use OpenTelemetry\SDK\Common\Time\BlockingScheduler;
+use OpenTelemetry\SDK\Common\Time\SchedulerInterface;
+use OpenTelemetry\SDK\Metrics\Exceptions\RetryableExportException;
+use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+
+class ExponentialWithJitterRetryPolicy implements RetryPolicyInterface
```
Nevay, line 13: `final`

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,199 @@
...
+        // Dont wait for the first attempt
+        if ($attempt == 0) {
+            return 0;
+        }
+        // Initial exponential backoff in mili seconds
+        $delay = ($this->backoffMultiplier ** $attempt) * $this->initialBackoff * 1000;
+        // Adding jitter to exponential backoff
+        if ($this->jitter > 0) {
+            $randomness = (int) ($delay * $this->jitter);
+            $delay = $delay + random_int(-$randomness, +$randomness);
```
Nevay, line 65: `rand()`

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,199 @@
...
+{
+    private $maxAttempts;
+    private $initialBackoff;
+    private $maxBackoff;
+    private $backoffMultiplier;
+    private $jitter;
+    public $retryableStatusCodes;
+    public $scheduler;
+
+    private function __construct(
```
Nevay, line 23: `public` and remove `::getDefault()`?

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+    {
+        return $this->retryPolicy;
+    }
+
+    /**
+     * Set the retryable status code. All the status codes apart from
+     * this list will not be retried
+     * @param array $statusCode: array of retryable status code
+     */
+    public function setRetryableStatusCodes(array $statusCode)
```
tidal, line 124: Why is this method needed? There are accessors for the `retryPolicy`property.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+        if ($this->retryPolicy) {
+            $this->retryPolicy->setRetryableStatusCodes($statusCode);
+        }
+    }
+
+    /**
+     * Set the delay scheduler which will decide whether
+     * @param SchedulerInterface $scheduler: schedular object which implements delay method
+     */
+    public function setDelayScheduler(SchedulerInterface $scheduler)
```
tidal, line 135: Why is this method needed? There are accessors for the `retryPolicy`property.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+    }
+
+    /**
+     * Set the retryable status code. All the status codes apart from
+     * this list will not be retried
+     * @param array $statusCode: array of retryable status code
+     */
+    public function setRetryableStatusCodes(array $statusCode)
+    {
+        if ($this->retryPolicy) {
```
tidal, line 126: This should check for `instanceof` not a truthy value.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+        }
+    }
+
+    /**
+     * Set the delay scheduler which will decide whether
+     * @param SchedulerInterface $scheduler: schedular object which implements delay method
+     */
+    public function setDelayScheduler(SchedulerInterface $scheduler)
+    {
+        if ($this->retryPolicy) {
```
tidal, line 137: This should check for `instanceof` not a truthy value.

src/SDK/Common/Time/BlockingScheduler.php
```diff
@@ -0,0 +1,27 @@
...
+    }
+    /**
+     * delays the current thread by $timeout seconds
+     *
+     * @param int $timeout - no of mili seconds to delay the current execution
```
tidal, line 19: Is it seconds or milliseconds?

> delays the current thread by $timeout seconds
> ... mili seconds to delay ...

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
 
-        return $this->doExport($spans); /** @phpstan-ignore-line */
+        // If retryPolicy is not set then just doExport once and return the status
+        if ($this->retryPolicy === null) {
+            self::logInfo('retry polict is null. Exporting the spans without retry' . PHP_EOL);
+
+            return $this->doExport($spans); /** @phpstan-ignore-line */
+        }
+
+        while ($shouldRetry) {
```
tidal, line 66: This adds a lot to the cyclomatic complexity of this method.
Could you extract the logic into a separate method to make it easier to follow?

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: Yes it is actually in millisecond. Its a typo in comment. Corrected

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: I have kept it here with the idea that each policy must have a delayschedular (blocking delay scheduler is the default type). so its better to keep in the policyInterface. we get the delayscheduler instance in exporter class using retryPolicy instance and use its delay function in exporter itself. So that way retryPolicy does perform the delay

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: ok I will change it to debug level

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: Done

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: Done

src/SDK/Behavior/LogsMessagesTrait.php
```diff
@@ -9,34 +9,34 @@
 
 trait LogsMessagesTrait
 {
-    private static function doLog(string $level, string $message, array $context): void
+    private static function writeLog(string $level, string $message, array $context): void
```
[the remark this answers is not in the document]
sanketmehta28: sure.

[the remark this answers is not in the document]
sanketmehta28: It was not intentional. I was trying it locally. 
I will take care of it

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+class ExponentialWithJitterRetryPolicyTest extends TestCase
+{
+    private OtlpExporter $exporter;
+    // private ClientInterface $client;
+
+    public function setUp(): void
+    {
+        $request = $this->createMock(RequestInterface::class);
+        // $this->client = $this->createMock(ClientInterface::class);
+        $this->exporter = $this->createExporter();
```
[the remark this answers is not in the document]
sanketmehta28: sure

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private function createExporter(): OtlpExporter
```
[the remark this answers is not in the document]
sanketmehta28: sure

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return new class() extends OtlpExporter {
+        };
+    }
+
+    public function test_retry_policy_set_properly()
+    {
+        $this->exporter->setRetryPolicy(ExponentialWithJitterRetryPolicy::getDefault());
```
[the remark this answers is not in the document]
sanketmehta28: sure

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: I have removed all setter getters from RetryPolicyInterface and moved them in actual retry policy

src/SDK/Resource/Detectors/Sdk.php
sanketmehta28: Reverted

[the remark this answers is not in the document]
sanketmehta28: Reverted

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * @covers \OpenTelemetry\SDK\Common\Retry\ExponentialWithJitterRetryPolicy
+ */
+class ExponentialWithJitterRetryPolicyTest extends TestCase
+{
+    private OtlpExporter $exporter;
+    // private ClientInterface $client;
```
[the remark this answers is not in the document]
sanketmehta28: Done

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+        );
+        $this->assertInstanceOf(
+            BlockingScheduler::class,
+            $this->exporter->getRetryPolicy()->getDelayScheduler()
+        );
+    }
+
+    public function test_delay_is_less_or_equal_to_max_backoff()
+    {
+        $this->exporter->setRetryPolicy(ExponentialWithJitterRetryPolicy::getDefault());
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+                self::logDebug('Retrying span export for ' . $attempt . ' time');
+            }
+            $attempt++;
+
+            try {
+                $status = $this->doExport($spans);
+            } catch (Throwable $e) {
+                if ($e instanceof \Error) {
+                    self::logError('Exception occured while retrying export
+                        span: ' . $e->getMessage() . '\n');
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -4,12 +4,19 @@
...
+use OpenTelemetry\SDK\Metrics\Exceptions\RetryableExportException;
 use OpenTelemetry\SDK\Trace\SpanDataInterface;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+use Throwable;
 
 trait SpanExporterTrait
 {
+    use LogsMessagesTrait;
     private bool $running = true;
+    protected ?RetryPolicyInterface $retryPolicy = null;
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -32,10 +39,15 @@ abstract public static function fromConnectionString(string $endpointUrl, string
...
      * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.7.0/specification/trace/sdk.md#exportbatch
      *
+     * @psalm-suppress ArgumentTypeCoercion
      * @psalm-return SpanExporterInterface::STATUS_*
      */
     public function export(iterable $spans): int
     {
+        $shouldRetry = true;
+        $status = SpanExporterInterface::STATUS_SUCCESS;
+        $attempt = 0;
```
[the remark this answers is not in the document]
sanketmehta28: This particular variable is for the current value of the attempt. That is why it is singular.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+     */
+    public function setRetryPolicy(RetryPolicyInterface $retryPolicy)
+    {
+        $this->retryPolicy = $retryPolicy;
+    }
+
+    /**
+     * Returns the retry policy set by the user
+     */
+    public function getRetryPolicy()
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +56,87 @@ public function export(iterable $spans): int
...
+            );
+        }
+
+        return $status; /** @phpstan-ignore-line */
+    }
+
+    /**
+     * Use to define the retry policy for the exporter.
+     * Exporter needs to call it with specific retryPolicy
+     * param: RetryPolicy <RetryPolicyInterace>
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+     * Returns whether the request should be retried.
+     *
+     * @param int $attempt - current number of retry attempt
+     * @param int $status - SpanExporterStatusInterface
+     * @param RetryableExportException|null $exception: exception caught while retrying
+     *
+     * @return bool
+     */
+    public function shouldRetry(
+        int $attempt,
```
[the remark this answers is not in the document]
sanketmehta28: as I have mentioned before, this is for the current value of attempt. That is why it is singular.

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+        int $status,
+        ?RetryableExportException $exception
+    ): ?bool;
+
+    /**
+     * Returns the time to wait in seconds.
+     * @param int $attempt: current number of retry attempt
+     *
+     * @return int
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public function getDelay(int $attempt): int;
+
+    public static function getDefault(): RetryPolicyInterface;
+
+    public function getMaxAttempts(): int;
+
+    public function setMaxAttempts(int $max_attempts);
```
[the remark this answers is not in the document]
sanketmehta28: All getter setters are removed from interface. Although I have taken care of this at other places

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+
+    public function setInitialBackoff(float $initial_backoff);
+
+    public function getMaxBackoff(): int;
+
+    public function setMaxBackoff(int $max_backoff);
+
+    public function getBackoffMultiplier(): float;
+
+    public function setBackoffMultipler(float $backoff_multiplier);
```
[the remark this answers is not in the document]
sanketmehta28: Done as mentioned above

src/SDK/Common/Time/BlockingScheduler.php
```diff
@@ -0,0 +1,27 @@
...
+     * @return void
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: Removed

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: Removed

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: Removed method

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: Method removed

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: @tidal:
It is just a while loop for no of attempts allowed. 
 Can you give me some more idea about how I can improve it?

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,192 @@
...
+    private $initialBackoff;
+    private $maxBackoff;
+    private $backoffMultiplier;
+    private $jitter;
+    public $retryableStatusCodes;
+    public $scheduler;
+
+    public function __construct($arguments=[])
+    {
+        extract($arguments, EXTR_IF_EXISTS);
```
brettmc, line 25: this extracts the arguments into local variables, however the code just gets them from the array anyway?

src/SDK/Common/Time/BlockingScheduler.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    use LogsMessagesTrait;
+    public function __construct()
+    {
+        self::logInfo('setting Blocking Delay scheduler');
+    }
+
+    public function delay(int $timeout): void
+    {
+        self::logInfo("Delaying the execution by $timeout milliseconds");
```
brettmc, line 19:
```suggestion
        self::logDebug("Delaying the execution by $timeout milliseconds");
```

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -44,7 +55,58 @@ public function export(iterable $spans): int
...
+
+    public function setRetryPolicy(RetryPolicyInterface $retryPolicy)
+    {
+        $this->retryPolicy = $retryPolicy;
+    }
+
+    /**
+     * @return RetryPolicyInterface|null
+     */
+    public function getRetryPolicy()
```
brettmc, line 107:
```suggestion
    public function getRetryPolicy() ?RetryPolicyInterface
```

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,83 @@
...
+use OpenTelemetry\SDK\Common\Retry\ExponentialWithJitterRetryPolicy;
+use OpenTelemetry\SDK\Common\Time\BlockingScheduler;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \OpenTelemetry\SDK\Common\Retry\ExponentialWithJitterRetryPolicy
+ */
+class ExponentialWithJitterRetryPolicyTest extends TestCase
+{
+    public function test_retry_policy_set_properly()
```
brettmc, line 16: these methods all need a `void` return typehint

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @dataProvider provider
+     */
+    public function test_delay_is_less_or_equal_to_max_backoff($attempt, $delay, $maxBackoff)
+    {
+        $this->assertLessThanOrEqual($maxBackoff, $delay);
+    }
+
+    public function provider()
```
brettmc, line 66:
```suggestion
    public function provider(): array
```

src/SDK/Common/Retry/RetryPolicyInterface.php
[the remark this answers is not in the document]
tidal: So it means attemptNr?

src/SDK/Trace/Behavior/SpanExporterTrait.php
tidal: I did not mean to improve the loop itself, but maybe to simply extract it.
However it's not be super important and could be done in a refactoring later.

brettmc:
> failed because of doLog() function in LogsMessagetrait.php

I understand now... By adding `LogsMessagesTrait` to `SpanExporterTrait`, our `LoggerExporter` ends up with two methods named `doLog`. In that case, it does seem reasonable to rename one of them as part of this PR (I'd nominate the one in `LoggerExporter`).
There have been a couple of unintended code changes in this PR so we were probably over-zealous in checking for them :)

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: The build will fail for php 7.4 for now. We can refactor the same after we drop the support for php 7.4. I hope thats fine

src/SDK/Common/Time/BlockingScheduler.php
sanketmehta28: Done

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
sanketmehta28: Done

tests/Unit/SDK/Common/Retry/ExponentialWithJitterRetryPolicyTest.php
sanketmehta28: Done

src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -35,6 +36,17 @@ class Exporter implements SpanExporterInterface
 
     private TraceServiceClient $client;
 
+    private array $retryableStatusCodes = [
+            \Grpc\STATUS_CANCELLED,
+            \Grpc\STATUS_DEADLINE_EXCEEDED,
+            \Grpc\STATUS_PERMISSION_DENIED,
```
Nevay, line 42: Not directly related to this PR, but `STATUS_PERMISSION_DENIED` and `STATUS_PERMISSION_DENIED` are incorrect, see [OTLP/gRPC Response](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-response).

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
Nevay: I assume we will reach beta stability before EOL of 7.4, refactoring then would be a BC-break.
Adding the parameters shouldn't break 7.4, it just requires users to provide default values explicitly if they want to override later arguments.

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -45,7 +56,55 @@ public function export(iterable $spans): int
...
+                self::logDebug('Waiting for ' . $delay . ' mili seconds before retrying export');
+                $this->retryPolicy->delay($delay);
+                self::logDebug('Retrying span export for ' . $attempt . ' time');
+            }
+            $attempt++;
+
+            try {
+                $status = $this->doExport($spans);
+            } catch (Throwable $e) {
+                if ($e instanceof \Error) {
```
Nevay, line 79: Retrying on `Error` seems incorrect; can we remove the exception handling completely?
This does not set `$status` and will provide an incorrect value (the previous `$status`) to `RetryPolicy::shouldRetry()`.

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,85 @@
...
+ *    NonBlockingScheduler will be more useful when fiber is implemented.
+ *    By Default scheduler is of type BlockingScheduler.
+ */
+interface RetryPolicyInterface
+{
+    public const DEFAULT_MAX_ATTEMPTS = 5;
+    public const DEFAULT_INITIAL_BACKOFF = 1;
+    public const DEFAULT_MAX_BACKOFF = 10;
+    public const DEFAULT_BACKOFF_MULTIPLIER = 1.5;
+    public const DEFAULT_JITTER = 0.1;
```
Nevay, line 70: These constants should be in `ExponentialWithJitterRetryPolicy`.

src/Contrib/OtlpGrpc/Exporter.php
sanketmehta28: @nI can change it in my PR but it was there before I have raised my PR. 
Do you want me to remove in my PR itself?

src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -69,7 +81,10 @@ public function __construct(
             $this->getIntFromEnvironment(Env::OTEL_EXPORTER_OTLP_TIMEOUT, $timeout);
 
         $this->setSpanConverter(new SpanConverter());
-
+        $retryPolicy = new ExponentialWithJitterRetryPolicy([
+            'retryableStatusCodes' => $this->retryableStatusCodes,
```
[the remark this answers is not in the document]
sanketmehta28: @Nevay: For GRPC and HTTP different status codes are used which can be divided into retryable and non-retryable.
To make sure that we retry only for retryable codes, I have associated status codes with retry policy.

What other solution do you suggest?

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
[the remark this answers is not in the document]
sanketmehta28: @tidal: whats Nr stands for?

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Returns whether the request should be retried.
+     *
+     * @param int $attempt - current number of retry attempt
+     * @param int $status - SpanExporterStatusInterface
+     * @param RetryableExportException|null $exception: exception caught while retrying
+     *
+     * @return bool
```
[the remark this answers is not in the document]
sanketmehta28: Done

src/Contrib/OtlpGrpc/Exporter.php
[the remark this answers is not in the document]
Nevay: Retryable status codes are handled by the exporter, the status codes associated with the retry policy are not used; see [line 145](https://github.com/open-telemetry/opentelemetry-php/blob/b159577d793afbdf51785bf2f110d5e5331d9500/src/Contrib/OtlpGrpc/Exporter.php#L145):
```php
if ($this->retryPolicy && in_array($status->code, $this->retryableStatusCodes, false)) {
```

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,72 @@
...
+        int $status
+    ): ?bool;
+
+    public function getDelay(int $attempt): int;
+
+    /**
+     * delay the execution of the thread by $timeout milliseconds
+     * @param  int $timeout: milliseconds to delay the execution
+     */
+    public function delay(int $timeout): void;
```
Nevay, line 71: > [I have kept it here with the idea that each policy must have a delayschedular](https://github.com/open-telemetry/opentelemetry-php/pull/677#discussion_r919080161)

After working on an async exporter: this should not be part of the retry policy and should be moved to the exporter; providing a scheduler is only useful for exporters that are not explicitely async, async exporters already know the preferred way of delaying execution.

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: I have removed everything out of the retryPolicyInterface.php except then shouldRetry(), delay() and getDelay() function.

src/Contrib/OtlpGrpc/Exporter.php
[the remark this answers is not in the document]
Nevay: This is still open - it should be possible to simply remove `$retryableStatusCodes` from `ExponentialRetryPolicy`, the status codes are no longer part of the `RetryPolicyInterface` and not used by the exporter (exporter uses its own `$retryableStatusCodes` directly, see comment above).

src/Contrib/OtlpGrpc/Exporter.php
sanketmehta28: Done

src/Contrib/OtlpGrpc/Exporter.php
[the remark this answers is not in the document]
sanketmehta28: Sure. Removed

sanketmehta28:
@Nevay @tidal Please do review it and approve

src/Contrib/OtlpGrpc/Exporter.php
```diff
@@ -69,7 +80,15 @@ public function __construct(
             $this->getIntFromEnvironment(Env::OTEL_EXPORTER_OTLP_TIMEOUT, $timeout);
 
         $this->setSpanConverter(new SpanConverter());
-
+        $retryPolicy = new ExponentialWithJitterRetryPolicy(
+            $maxAttempts = ExponentialWithJitterRetryPolicy::DEFAULT_MAX_ATTEMPTS,
```
Nevay, line 84: Can remove arguments, `new ExponentialWithJitterRetryPolicy()`.

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,154 @@
...
+        // Adding jitter to exponential backoff
+        if ($this->jitter > 0) {
+            $randomness = (int) ($delay * $this->jitter);
+            $delay += rand(-$randomness, +$randomness);
+        }
+
+        return min((int) $delay, $this->maxBackoff * 1000);
+    }
+
+    public function getMaxAttempts(): int
```
Nevay, line 65: I'm in favor of removing the getters/setters, no reason to expose more than needed.

src/SDK/Common/Retry/RetryPolicyInterface.php
```diff
@@ -0,0 +1,54 @@
...
+ *    DelayScheduler is used to define the delay method for the exporter whether should it block the current execution
+ *    or not.
+ *    NonBlockingScheduler will be more useful when fiber is implemented.
+ *    By Default scheduler is of type BlockingScheduler.
+ */
+interface RetryPolicyInterface
+{
+    public function shouldRetry(
+        int $attempt,
+        int $status
```
Nevay, line 50: I don't think this parameter is useful, will this ever be used for something besides checking for `SpanExporterInterface::STATUS_FAILED_RETRYABLE`?

src/SDK/Trace/Behavior/SpanExporterTrait.php
```diff
@@ -33,15 +39,59 @@ abstract public static function fromConnectionString(string $endpointUrl, string
...
+        // If retryPolicy is not set then just doExport once and return the status
+        if ($this->retryPolicy === null) {
+            self::logDebug('retry polict is null. Exporting the spans without retry');
+
+            return $this->doExport($spans); /** @phpstan-ignore-line */
+        }
+
+        while ($shouldRetry) {
+            // find the delay time before retry (it is 0 for the first attempt)
+            $delay = $this->retryPolicy->getDelay($attempt);
```
Nevay, line 63: Having the retry logic in the trait instead of the exporter complicates handling `Retry-After` header & co.
[OTLP/HTTP Throttling](https://github.com/open-telemetry/opentelemetry-specification/blob/dbea54d2a416fe9e487d78dd33f8ab9251596bc2/specification/protocol/otlp.md#otlphttp-throttling)

src/Contrib/OtlpGrpc/Exporter.php
sanketmehta28: Done

sanketmehta28: Done

src/SDK/Common/Retry/RetryPolicyInterface.php
sanketmehta28: For current state, it is only being used for checking if the status is STATUS_FAILED_RETRYABLE. but this can be checked against a list of retryable codes for diff retry policies and exporters.

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: I have to put the retry logic in trait only because it is applicable to all span exporters and export() is actually implemented in that only.

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: I can remove the getters but setters are important as value validation is implemented in the same.

Nevay: Setters can be `private` if you want to keep the validation logic in them.

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        return $this->backoffMultiplier;
+    }
+
+    public function setBackoffMultipler(
+        float $backoff_multiplier = self::DEFAULT_BACKOFF_MULTIPLIER
+    ) {
+        $this->checkArgument(
+            $backoff_multiplier > 1,
+            sprintf('Backoff multiplier must be greater than 0: "%s" value provided', $backoff_multiplier)
```
Nevay, line 125: `must be greater than 1`

src/SDK/Common/Retry/ExponentialWithJitterRetryPolicy.php
sanketmehta28: Done

src/SDK/Trace/Behavior/SpanExporterTrait.php
sanketmehta28: @Nevay : Can you please give me an example where I can implement this?

Nevay: I think we can ignore this; exporters that have to handle `Retry-After` should implement their own retry logic instead of using this trait.

sanketmehta28: ok I will move the logic to exporter

PR #686, opened 2022-05-26
Fix W3C test service

yuktea:
Attempts to fix the W3C test service by updating usage patterns to match the refactorings and functional changes that have been applied since it was last updated. The effort remains a work-in-progress with the tests failing due to a reason I've been unable to diagnose yet.

### DIFF
--- a/src/API/Trace/Propagation/TraceContextPropagator.php
+++ b/src/API/Trace/Propagation/TraceContextPropagator.php
@@ -32,7 +32,6 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
     public const TRACEPARENT = 'traceparent';
     public const TRACESTATE = 'tracestate';
     private const VERSION = '00'; // Currently, only '00' is supported
-    private const VERSION_REGEX = '/^(?!ff)[\da-f]{2}$/';
 
     private static ?self $instance = null;
 
@@ -104,22 +103,22 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
 
         [$version, $traceId, $spanId, $traceFlags] = $pieces;
 
-        $versionIsFuture = hexdec($version) > hexdec(self::VERSION);
-        $versionIsInvalid = !preg_match(self::VERSION_REGEX, $version)
-            || ($version !== self::VERSION && !$versionIsFuture);
-
         /**
-         * Restart trace if:
+         * Return invalid if:
          * - Version is invalid (not 2 char hex or 'ff')
-         * - Trace doesn't have 4 pieces and is not a future version
-         * - Trace ID, Span ID or Trace Flag are invalid
+         * - Trace version, trace ID, span ID or trace flag are invalid
          */
-        if ($versionIsInvalid
-            || (count($pieces) !== 4 && !$versionIsFuture)
+        if (!SpanContext::isValidTraceVersion($version)
             || !SpanContext::isValidTraceId($traceId)
             || !SpanContext::isValidSpanId($spanId)
             || !SpanContext::isValidTraceFlag($traceFlags)
         ) {
+            return SpanContext::getInvalid();
+        }
+
+        // Return invalid if the trace version is not a future version but still has > 4 pieces.
+        $versionIsFuture = hexdec($version) > hexdec(self::VERSION);
+        if (count($pieces) > 4 && !$versionIsFuture) {
             return SpanContext::getInvalid();
         }
 
--- a/src/API/Trace/Propagation/TraceContextPropagator.php
+++ b/src/API/Trace/Propagation/TraceContextPropagator.php
@@ -16,6 +16,7 @@ use OpenTelemetry\Context\Propagation\ArrayAccessGetterSetter;
 use OpenTelemetry\Context\Propagation\PropagationGetterInterface;
 use OpenTelemetry\Context\Propagation\PropagationSetterInterface;
 use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
+use function preg_replace;
 
 /**
  * TraceContext is a propagator that supports the W3C Trace Context format
@@ -117,7 +118,13 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
         // Tracestate = 'Vendor1=Value1,...,VendorN=ValueN'
         $rawTracestate = $getter->get($carrier, self::TRACESTATE);
         if ($rawTracestate !== null) {
-            $tracestate = new TraceState($rawTracestate);
+            /*
+             * Some servers concatenate multiple headers with ';' -- we need to replace these with ','
+             * This is still a workaround and doesn't get around the problem fully, specifically it doesn't
+             * handle edge cases where the header has a trailing ';' or an empty trace state.
+             */
+            $sanitizedTraceState = preg_replace('/;(?=[^,=;]*=|$)/', ',', $rawTracestate);
+            $tracestate = new TraceState($sanitizedTraceState);
 
             return SpanContext::createFromRemoteParent(
                 $traceId,
--- a/src/API/Trace/Propagation/TraceContextPropagator.php
+++ b/src/API/Trace/Propagation/TraceContextPropagator.php
@@ -32,6 +32,7 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
     public const TRACEPARENT = 'traceparent';
     public const TRACESTATE = 'tracestate';
     private const VERSION = '00'; // Currently, only '00' is supported
+    private const HEX_REGEX = '/^[\da-f_]+$/';
 
     private static ?self $instance = null;
 
@@ -93,20 +94,27 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
             return SpanContext::getInvalid();
         }
 
-        // Traceparent = {version}-{trace-id}-{parent-id}-{trace-flags}
+        // traceParent = {version}-{trace-id}-{parent-id}-{trace-flags}
         $pieces = explode('-', $traceparent);
-
-        // Unable to extract traceparent. Expected 4 values
-        if (count($pieces) !== 4) {
-            return SpanContext::getInvalid();
-        }
 
         [$version, $traceId, $spanId, $traceFlags] = $pieces;
 
-        // Validates the version, traceId, spanId and traceFlags
-        // Returns an invalid spanContext if any of the checks fail
-        if ($version !== self::VERSION || !SpanContext::isValidTraceId($traceId) ||
-            !SpanContext::isValidSpanId($spanId) || !SpanContext::isValidTraceFlag($traceFlags)) {
+        $versionIsFuture = hexdec($version) > hexdec(self::VERSION);
+        $versionIsInvalid = !preg_match(self::HEX_REGEX, $version)
+            || ($version !== self::VERSION && !$versionIsFuture);
+
+        /**
+         * Restart trace if:
+         * - Version is invalid (not hex)
+         * - Trace doesn't have 4 pieces and is not a future version
+         * - Trace ID, Span ID or Trace Flag are invalid
+         */
+        if ($versionIsInvalid
+            || (count($pieces) !== 4 && !$versionIsFuture)
+            || !SpanContext::isValidTraceId($traceId)
+            || !SpanContext::isValidSpanId($spanId)
+            || !SpanContext::isValidTraceFlag($traceFlags)
+        ) {
             return SpanContext::getInvalid();
         }
 
--- a/src/API/Trace/Propagation/TraceContextPropagator.php
+++ b/src/API/Trace/Propagation/TraceContextPropagator.php
@@ -16,7 +16,6 @@ use OpenTelemetry\Context\Propagation\ArrayAccessGetterSetter;
 use OpenTelemetry\Context\Propagation\PropagationGetterInterface;
 use OpenTelemetry\Context\Propagation\PropagationSetterInterface;
 use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
-use function preg_replace;
 
 /**
  * TraceContext is a propagator that supports the W3C Trace Context format
@@ -118,13 +117,7 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
         // Tracestate = 'Vendor1=Value1,...,VendorN=ValueN'
         $rawTracestate = $getter->get($carrier, self::TRACESTATE);
         if ($rawTracestate !== null) {
-            /*
-             * Some servers concatenate multiple headers with ';' -- we need to replace these with ','
-             * This is still a workaround and doesn't get around the problem fully, specifically it doesn't
-             * handle edge cases where the header has a trailing ';' or an empty trace state.
-             */
-            $sanitizedTraceState = preg_replace('/;(?=[^,=;]*=|$)/', ',', $rawTracestate);
-            $tracestate = new TraceState($sanitizedTraceState);
+            $tracestate = new TraceState($rawTracestate);
 
             return SpanContext::createFromRemoteParent(
                 $traceId,
--- a/src/API/Trace/Propagation/TraceContextPropagator.php
+++ b/src/API/Trace/Propagation/TraceContextPropagator.php
@@ -32,7 +32,7 @@ final class TraceContextPropagator implements TextMapPropagatorInterface
     public const TRACEPARENT = 'traceparent';
     public const TRACESTATE = 'tracestate';
     private const VERSION = '00'; // Currently, only '00' is supported
-    private const HEX_REGEX = '/^[\da-f_]+$/';
+    private const VERSION_REGEX = '/^(?!ff)[\da-f]{2}$/';
 
     private static ?self $instance = null;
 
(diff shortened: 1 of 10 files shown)


yuktea:
@bobstrecansky

> You'll need to run a `make style` to format your code and get to the next step of CI.
> 
> When I pulled down your branch and ran `make test` to test your code, I was able to see two relatively simple errors. Is that what you see as well?

I think you'll need to run `make w3c-test-service` to get the test service to run. What I see on my end: first all the tests from the w3c test.py script are failing.
```
----------------------------------------------------------------------
Ran 40 tests in 24.305s

FAILED (failures=40)
ERROR: 40
make: *** [Makefile:64: w3c-test-service] Error 40
```

On investigating the `curl` outputs from the symfony app, I see an error about an invalid argument to `foreach` (1) and (2):
```
<!-- Attempted to load class &quot;AlwaysOnSampler&quot; from namespace &quot;OpenTelemetry\SDK\Trace\Sampler&quot;.
Did you forget a &quot;use&quot; statement for another namespace? (500 Internal Server Error) -->
```

I haven't figured out why (1) is happening because I didn't change the (I think) relevant part in `W3CTestService/TestController.php`.

Any clue what could be causing (2)?

yuktea:
At this point, the test service is running as expected. However all 40 w3c tests are failing, output for all failures from the w3c trace-context test script looks like:
```
Harness trying to send the following request to your service 127.0.0.1:8001/test

POST 127.0.0.1:8001/test HTTP/1.1
traceparent: 00-12345678901234567890123456789012-1234567890123456-00
tracestate: foo=bar=baz

[{'arguments': [],
  'url': 'http://127.0.0.1:7777/c29f95d4316d471d9e980c09c29bced1.0'}]

Harness got an exception InvalidURL

127.0.0.1:8001/test
```

@bobstrecansky any ideas to what might be causing this?

yuktea:
I've ironed out issues in the service controller. All tests still fail, I suspect we're doing something wrong within the controller still but I'm not sure. Could use help in diagnosis.

bobstrecansky:
I'm trying to replicate this locally.  What commands are you running to see your failures?

yuktea:
> I'm trying to replicate this locally. What commands are you running to see your failures?

`make w3c-test-service`

tests/TraceContext/W3CTestService/index.php
```diff
@@ -25,25 +30,33 @@
...
+    'http://localhost:9412/api/v2/spans',
+    $httpClient,
+    $httpClient,
+    $httpClient,
 );
 
-if (SamplingResult::RECORD_AND_SAMPLE === $samplingResult->getDecision()) {
-    $tracer = (new TracerProvider())
-        ->addSpanProcessor(new BatchSpanProcessor($exporter, ClockFactory::getDefault()))
+$shouldRecordAndSampleSpan = $samplingResult->getDecision() === SamplingResult::RECORD_AND_SAMPLE;
```
brettmc, line 43: The sampling decision is part of SpanBuilder now, and is taken care of automatically.

yuktea: Thanks. I'll change this just as we're documenting in #689.

yuktea:
At of [dff26c7](https://github.com/open-telemetry/opentelemetry-php/pull/686/commits/dff26c7119e4285b0b2ff99c6267645b90cdb853) the w3c tests are still failing. I'll try to document my interpretation of what's going wrong with these tests in the hope it helps us figure out what it's lacking.

- All 40 w3c tests fail.
- Failure diagnosis output looks like:
  ```
  ======================================================================
  FAIL: test_tracestate_value_illegal_characters (__main__.TraceContextTest)
  harness sends a request with an invalid tracestate header with illegal value format
  ----------------------------------------------------------------------
  Traceback (most recent call last):
  File "/usr/src/myapp/tests/TraceContext/W3CTestService/trace-context/test/test.py", line 819, in test_tracestate_value_illegal_characters
  traceparent, tracestate = self.make_single_request_and_get_tracecontext([
  File "/usr/src/myapp/tests/TraceContext/W3CTestService/trace-context/test/test.py", line 113, in make_single_request_and_get_tracecontext
  return (self.get_traceparent(headers), self.get_tracestate(headers))
  File "/usr/src/myapp/tests/TraceContext/W3CTestService/trace-context/test/test.py", line 101, in get_traceparent
  self.assertEqual(len(retval), 1, 'expect one traceparent header, got {} {!r}'.format('more' if retval else 'zero', retval))
  AssertionError: 0 != 1 : expect one traceparent header, got zero []
  ```
- Going by the diagnosis message and looking through the TestController, I believe we're not returning the response header as expected. However I have not been able to figure out exactly what it is we should be returning in the
  headers/body. Please take a look: https://github.com/yuktea/opentelemetry-php/blob/dff26c7119e4285b0b2ff99c6267645b90cdb853/tests/TraceContext/W3CTestService/TestController.php#L44
- To reproduce on your end, please run `make w3c-test-service`. At this point, I have fixed the script and verified that the failure is reproducible.

/cc @bobstrecansky @tidal @brettmc

tests/TraceContext/W3CTestService/symfony-setup
```diff
@@ -8,25 +8,28 @@ curl -sS https://get.symfony.com/cli/installer | bash
 mv /root/.symfony/bin/symfony /usr/local/bin/symfony
 
 # Set up Symfony test application
-cd tests/TraceContext/W3CTestService
+cd tests/TraceContext/W3CTestService || exit
```
bobstrecansky, line 11: Not sure of the reasoning behind `|| exit` for these - I'd say either give a detailed error message or don't do this.

Nevay:
Regarding [TestController#L44](https://github.com/yuktea/opentelemetry-php/blob/dff26c7119e4285b0b2ff99c6267645b90cdb853/tests/TraceContext/W3CTestService/TestController.php#L44), this should be: 
```php
$traceCtxPropagator->inject($headers, null, $context);
```

Some causes for the remaining failures:
- the webserver concatenates multiple tracestate headers using ';' (see test_tracestate_empty_header)
- the testsuite expects that the entire tracestate is discarded on invalid member (currently invalid members are dropped) (not required by spec; "If the tracestate header cannot be parsed the vendor MAY discard the entire header. Invalid tracestate entries MAY also be discarded.")
- the tracestate propagator doesnt handle future versions (should allow >= 4 components instead of exactly 4 components if version is not 0x00)

yuktea:
> I'm able to reproduce your error, and I also get 40 errors. It appears you turned the debug log on, that's a good start. What inconsistencies have you found in the traceparent logic that would cause these errors? Remember, this pulls in the w3c test service here:
> 
> [https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/TraceContext/W3CTestService/symfony-setup#L47](https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/TraceContext/W3CTestService/symfony-setup?rgh-link-date=2022-05-31T12%3A23%3A39Z#L47)

I think the biggest consistency was discarding the traces the expected by the tests in our response headers to the test harness. Beyond this we still have some more failures to investigate that @Nevay's assessment seems to cover well.

As of [4cfc68c](https://github.com/open-telemetry/opentelemetry-php/pull/686/commits/4cfc68c1c2ea1eceaf579248aa0165808b6f0a33), we are passing 30/40 tests. I'm currently investigating the other failures with help from the comments above.

yuktea:
I think at this point the W3C test service is working fine. The 10 remaining failures we likely owe to not complying with the W3C trace context spec which we could work on outside of this PR.

@Nevay
> the webserver concatenates multiple tracestate headers using ';' (see test_tracestate_empty_header)
I tried getting around this by doing a string replacement `;` -> `,` but that breaks the same amount of tests it fixes. Apparently `;` is a valid character in tracestates :/. 
Is there any other workaround for this? I haven't been able to figure it out yet.

Nevay (commented):
> I tried getting around this by doing a string replacement ; -> , but that breaks the same amount of tests it fixes. Apparently ; is a valid character in tracestates :/.
Is there any other workaround for this? I haven't been able to figure it out yet.

I don't remember whether I solved this issue; nowadays i'm running the testsuite with symfony running in an `amphp/http-server` runtime, which doesn't have this issue.
Splitting the header at the last semicolon before `=` (something like `preg_replace('/;(?=[^,=]*=|\h*$)/', ',', $rawTracestate)` would resolve the failures but wouldn't handle edge cases like tracestate with trailing `;` vs. tracestate followed by empty tracestate. Switching to a server that doesn't concatenate the headers might be the easiest solution.

> Once we get this merged I'll open another PR to factor this to vanilla PHP for speed.

Either this, or use an instrumented HttpKernel + HttpClient with a controller that only processes the requests without interacting with OpenTelemetry directly to showcase an instrumented app.

tests/TraceContext/W3CTestService/TestController.php
```diff
@@ -15,39 +16,51 @@ class TestController
...
+            $url = $case['url'];
+            $arguments = $case['arguments'];
 
-                try {
-                    $context = TraceContextPropagator::extract($request->headers->all());
-                } catch (\InvalidArgumentException $th) {
-                    $context = SpanContext::generate();
-                }
+            $span = $tracer->spanBuilder($url)->setParent($context)->startSpan();
+            $spanScope = $span->activate();
```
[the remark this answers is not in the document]
yuktea: Would we still need the `->setParent()` call?

yuktea:
> > I tried getting around this by doing a string replacement ; -> , but that breaks the same amount of tests it fixes. Apparently ; is a valid character in tracestates :/.
> > Is there any other workaround for this? I haven't been able to figure it out yet.
> 
> I don't remember whether I solved this issue; nowadays i'm running the testsuite with symfony running in an `amphp/http-server` runtime, which doesn't have this issue. Splitting the header at the last semicolon before `=` (something like `preg_replace('/;(?=[^,=]*=|\h*$)/', ',', $rawTracestate)` would resolve the failures but wouldn't handle edge cases like tracestate with trailing `;` vs. tracestate followed by empty tracestate. Switching to a server that doesn't concatenate the headers might be the easiest solution.
> 
> After these changes the errors are down to:
> 
>     * 1x not supporting future versions, this has be resolved as it is required by spec
> 
>     * 4x not dropping the entire tracestate on invalid member
> 
> 
> > Once we get this merged I'll open another PR to factor this to vanilla PHP for speed.

I figure then we should switch to a non-concatenating server in #695 (only a draft because I'd already worked on it).

> 
> Either this, or use an instrumented HttpKernel + HttpClient with a controller that only processes the requests without interacting with OpenTelemetry directly to showcase an instrumented app.

If it would still be fast to set up and run `make w3c-test-service` with that, it sounds like a good idea to me! I lack context, but would this still be a Symfony app that requires the current time consuming setup?

tests/TraceContext/W3CTestService/TestController.php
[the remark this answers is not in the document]
Nevay: Yes, `::setParent()` is necessary unless you `::activate()` the extracted parent context before starting the child span.

Nevay:
> do we need the |\h*$ there or can do without?

The `|\h*$` interprets the trailing `;` case as tracestate + empty tracestate, which is used by some of the test-cases.

> would this still be a Symfony app that requires the current time consuming setup?

Yes, just with most logic extracted out of the controller into decorator/eventlistener. It might make sense to do this iff using an actual symfony example project that the documentation can refer to instead of creating the project on the fly.
(There are some general performance improvements that can be made like running the test-suite in a different container/python image and mounting composer cache, but the basic setup would be the same.)

yuktea:
> > do we need the |\h*$ there or can do without?
> 
> The `|\h*$` interprets the trailing `;` case as tracestate + empty tracestate, which is used by some of the test-cases.
> 
> > would this still be a Symfony app that requires the current time consuming setup?
> 
> Yes, just with most logic extracted out of the controller into decorator/eventlistener. It might make sense to do this iff using an actual symfony example project that the documentation can refer to instead of creating the project on the fly. (There are some general performance improvements that can be made like running the test-suite in a different container/python image and mounting composer cache, but the basic setup would be the same.)

I agree that would make much more sense as an external project we can refer to. For an internal implementation we can use in our CI, something lightweight and vanilla is more ideal. Maybe we could have both?

tests/TraceContext/W3CTestService/TestController.php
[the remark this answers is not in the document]
yuktea: Resolves 4x test cases in [3d59965](https://github.com/open-telemetry/opentelemetry-php/pull/686/commits/3d5996515747ab5518602b0c630b6748dc461c0a). Thank you!

yuktea:
> After these changes the errors are down to:
> 
>     * 1x not supporting future versions, this has be resolved as it is required by spec
> 
>     * 4x not dropping the entire tracestate on invalid member

Now we're down to these 5 failures in this PR. Should we cover them here?

/cc @bobstrecansky @tidal

bobstrecansky:
We should attempt to cover them in this PR if possible.

yuktea:
Should be good to go. Please check that the spec compliance changes are at the right place. Thanks

src/API/Trace/Propagation/TraceContextPropagator.php
```diff
@@ -97,15 +97,20 @@ private static function extractImpl($carrier, PropagationGetterInterface $getter
...
+        // If the header does not have at least 4 pieces, it is invalid -- restart the trace.
+        if (count($pieces) < 4) {
+            return SpanContext::getInvalid();
+        }
+
         [$version, $traceId, $spanId, $traceFlags] = $pieces;
 
         $versionIsFuture = hexdec($version) > hexdec(self::VERSION);
-        $versionIsInvalid = !preg_match(self::HEX_REGEX, $version)
+        $versionIsInvalid = !preg_match(self::VERSION_REGEX, $version)
```
Nevay, line 108:
```php
hexdec($version) === 0xff
```

yuktea: I believe the version regex `/^(?!ff)[\da-f]{2}$/` will take care of both `ff` and non-hex or >2 char hex sequences.

yuktea:
Updated unit tests.

src/API/Trace/Propagation/TraceContextPropagator.php
Nevay: Non-hex has be checked before calling `hexdec()` in the line above this, otherwise it might trigger a deprecation notice.

yuktea: Yes, absolutely! This was triggering an error on tests on I'm in the process of writing. Let me get this in order.

yuktea:
lets get this merged?

tests/Unit/API/Trace/TraceStateTest.php
```diff
@@ -87,32 +89,32 @@ public function test_max_tracestate_list_members(): void
         $validTracestate = new TraceState(implode(TraceState::LIST_MEMBERS_SEPARATOR, $rawTraceState));
         $this->assertSame(TraceState::MAX_TRACESTATE_LIST_MEMBERS, $validTracestate->getListMemberCount());
 
-        // Add a list-member to the tracestate that exceeds the max of 32. This will cause it to be truncated
+        // Add a list-member to the tracestate that exceeds the max of 32. This will the tracestate to be discarded.
```
brettmc, line 92:
```suggestion
        // Add a list-member to the tracestate that exceeds the max of 32. This will cause the tracestate to be discarded.
```

tests/TraceContext/W3CTestService/index.php
```diff
@@ -4,46 +4,41 @@
...
-use OpenTelemetry\SDK\Trace\SamplingResult;
 use OpenTelemetry\SDK\Trace\SpanProcessor\BatchSpanProcessor;
 use OpenTelemetry\SDK\Trace\TracerProvider;
 use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\ErrorHandler\Debug;
+use Symfony\Component\HttpClient\Psr18Client;
 use Symfony\Component\HttpFoundation\Request;
 
-(new Dotenv())->bootEnv(dirname(__DIR__) . '/.env');
+// @todo remove
```
brettmc, line 15: can this be removed now?

brettmc (approved):

tests/TraceContext/W3CTestService/index.php
```diff
@@ -4,46 +4,41 @@
...
+Debug::enable();
```
yuktea, line 16:
```suggestion
```

tests/TraceContext/W3CTestService/index.php
yuktea: done

tests/TraceContext/W3CTestService/index.php
```diff
@@ -4,46 +4,39 @@
...
-use OpenTelemetry\API\Trace as API;
-use OpenTelemetry\Context\Context;
 use OpenTelemetry\Contrib\Jaeger\Exporter as JaegerExporter;
-use OpenTelemetry\SDK\Common\Time\ClockFactory;
-use OpenTelemetry\SDK\Trace\Sampler\AlwaysOnSampler;
-use OpenTelemetry\SDK\Trace\SamplingResult;
 use OpenTelemetry\SDK\Trace\SpanProcessor\BatchSpanProcessor;
 use OpenTelemetry\SDK\Trace\TracerProvider;
 use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\ErrorHandler\Debug;
```
yuktea, line 11:
```suggestion
```

src/API/Trace/TraceState.php
```diff
@@ -134,16 +137,29 @@ private function parse(string $rawTracestate): array
...
-            $listMembers = explode(self::LIST_MEMBERS_SEPARATOR, $rawTracestate);
+        /**
+         * Some servers concatenate multiple headers with ';' -- we need to replace these with ','
+         * This is still a workaround and doesn't get around the problem fully, specifically it doesn't
+         * handle edge cases where the header has a trailing ';' or an empty trace state.
+         * We also need to trim trailing separators from the header, found when a header is empty.
+         * @var string $sanitizedTraceState
+         */
+        $sanitizedTraceState = (function () use ($rawTracestate) {
+            $fixedSeparators = preg_replace(self::SERVER_CONCAT_HEADERS_REGEX, self::LIST_MEMBERS_SEPARATOR, $rawTracestate);
```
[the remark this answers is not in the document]
yuktea: Sounds like a much more "correct" way of handling this. Thanks for the PR

Nevay (approved):

PR #694, opened 2022-06-01
#632 : Move stack trace formatting out of Span class

### LINKED ISSUE #632: Move stack trace formatting out of Span class
The method [formatStackTrace ](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Trace/Span.php#L175) should be moved from `Span` into a dedicated Util/Formatter class in the `SDK\Common` namespace since formatting stack traces is not the responsibility of a `Span`.

- Move logic in `formatStackTrace`  from `Span` into a dedicated Util/Formatter class
- Add Backward compatibility by calling the new method from `Span::formatStackTrace` and tigger deprecation notice with [Compatibility\Util::triggerMethodDeprecationNotice](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Common/Dev/Compatibility/Util.php#L23)

amber0612:
closes #632

### DIFF
--- a/src/SDK/Common/Util/TracingUtil.php
+++ b/src/SDK/Common/Util/TracingUtil.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Common\Util;
+
+use Throwable;
+
+class TracingUtil
+{
+    /**
+         * This function provides a more java-like stacktrace
+         * that supports exception chaining and provides exact
+         * lines of where exceptions are thrown
+         *
+         * Example:
+         * Exception: Thrown from grandparent
+         *  at grandparent_func(test.php:56)
+         *  at parent_func(test.php:51)
+         *  at child_func(test.php:44)
+         *  at (main)(test.php:62)
+         *
+         * Credit: https://www.php.net/manual/en/exception.gettraceasstring.php#114980
+         *
+         */
+    public static function formatStackTrace(Throwable $e, array &$seen = null): string
+    {
+        $starter = $seen ? 'Caused by: ' : '';
+        $result = [];
+        if (!$seen) {
+            $seen = [];
+        }
+        $trace  = $e->getTrace();
+        $prev   = $e->getPrevious();
+        $result[] = sprintf('%s%s: %s', $starter, get_class($e), $e->getMessage());
+        $file = $e->getFile();
+        $line = $e->getLine();
+        while (true) {
+            $current = "$file:$line";
+            if (in_array($current, $seen, true)) {
+                $result[] = sprintf(' ... %d more', count($trace)+1);
+
+                break;
+            }
+
+            // Lambda to format traces -- we want to format the trace with '.' separators
+            $slashToDot = function ($str) {
+                return str_replace('\\', '.', $str);
+            };
+
+            $traceHasKey = array_key_exists(0, $trace);
+            $traceKeyHasClass = $traceHasKey && array_key_exists('class', $trace[0]);
+            $traceKeyHasFunction = $traceKeyHasClass && array_key_exists('function', $trace[0]);
+
+            $result[] = sprintf(
+                ' at %s%s%s(%s%s%s)',
+                $traceKeyHasClass ? $slashToDot($trace[0]['class']) : '',
+                $traceKeyHasFunction ? '.' : '',
+                $traceKeyHasFunction ? $slashToDot($trace[0]['function']) : 'main',
+                $line === null ? $file : basename($file),
+                $line === null ? '' : ':',
+                $line ?? ''
+            );
+            $seen[] = "$file:$line";
+            if (!count($trace)) {
+                break;
+            }
+            $file = array_key_exists('file', $trace[0]) ? $trace[0]['file'] : 'Unknown Source';
+            $line = array_key_exists('file', $trace[0]) && array_key_exists('line', $trace[0]) && $trace[0]['line'] ? $trace[0]['line'] : null;
+            array_shift($trace);
+        }
+        $result = implode("\n", $result);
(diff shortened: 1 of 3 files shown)


brettmc (approved):

brettmc:
Can you sort out signing a CLA please? (We can't accept your contribution without this) And also try to write a unit test for the newly-created class, since code coverage has dropped with this refactor.

amber0612:
> Can you sort out signing a CLA please? (We can't accept your contribution without this) And also try to write a unit test for the newly-created class, since code coverage has dropped with this refactor.

Hi Bret, Unit test for formatStackTrace was not written previously. Can u help me with what am i supposed to assert equal for Unit test of the formatStackTrace function. As the function returns String which is stack Trace and string returned will be complex with Line numbers too. Do u have ideas on how to handle unit test for this...

Nevay:
I would write them as `.phpt` tests to have a clean stacktrace. There is an open issue with PHPUnit adding an additional frame; we can use [run-tests.php](https://github.com/php/php-src/blob/master/run-tests.php) instead (at least for now).

```php
--TEST--
Basic stacktrace format
--FILE--
<?php
use OpenTelemetry\SDK\Common\Util\TracingUtil;

require_once 'vendor/autoload.php';

echo TracingUtil::formatStackTrace(new Exception('message')), "\n";
?>
--EXPECT--
Exception: message
 at main(test_stacktrace_basic.php:6)
```

(I plan on creating a follow-up PR to resolve some issues with the current implementation and could take care of the tests if necessary.)

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+use function array_key_exists;
+use function array_shift;
+use function basename;
+use function in_array;
+use function sprintf;
+use function str_replace;
+
+use Throwable;
+
+class TracingUtil
```
tidal, line 16: Could you please rename the class to smth like `StackTraceFormatter` and the method to just `format`, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+         *  at parent_func(test.php:51)
+         *  at child_func(test.php:44)
+         *  at (main)(test.php:62)
+         *
+         * Credit: https://www.php.net/manual/en/exception.gettraceasstring.php#114980
+         *
+         */
+    public static function formatStackTrace(Throwable $e, array &$seen = null): string
+    {
+        $starter = $seen ? 'Caused by: ' : '';
```
tidal, line 35: Could you make `Caused by: ` a constant, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+        $result = [];
+        if (!$seen) {
```
tidal, line 37: Could you please check for a `null` value here, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            // Lambda to format traces -- we want to format the trace with '.' separators
+            $slashToDot = function ($str) {
+                return str_replace('\\', '.', $str);
+            };
+
+            $traceHasKey = array_key_exists(0, $trace);
+            $traceKeyHasClass = $traceHasKey && array_key_exists('class', $trace[0]);
+            $traceKeyHasFunction = $traceKeyHasClass && array_key_exists('function', $trace[0]);
```
tidal, line 60: Could you please male `class`and `function` class constants, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+            $result[] = sprintf(
+                ' at %s%s%s(%s%s%s)',
+                $traceKeyHasClass ? $slashToDot($trace[0]['class']) : '',
+                $traceKeyHasFunction ? '.' : '',
+                $traceKeyHasFunction ? $slashToDot($trace[0]['function']) : 'main',
+                $line === null ? $file : basename($file),
+                $line === null ? '' : ':',
+                $line ?? ''
+            );
+            $seen[] = "$file:$line";
```
tidal, line 71: Could you make this a `sprintf` call, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+            if (!count($trace)) {
```
tidal, line 72: Could you please check for a numeric value here, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+                break;
+            }
+            $file = array_key_exists('file', $trace[0]) ? $trace[0]['file'] : 'Unknown Source';
+            $line = array_key_exists('file', $trace[0]) && array_key_exists('line', $trace[0]) && $trace[0]['line'] ? $trace[0]['line'] : null;
```
tidal, line 76: Could you make `line` a class constant, please?

src/SDK/Common/Util/TracingUtil.php
```diff
@@ -0,0 +1,86 @@
...
+            array_shift($trace);
+        }
+        $result = implode("\n", $result);
+        if ($prev) {
```
tidal, line 80: Could you check for a type here, please?

tidal (changes requested):

src/SDK/Common/Util/TracingUtil.php
tidal: Yes

tidal (approved):

PR #701, opened 2022-06-04
Remove symfony quickstart and link to bundle instead

DrLuke:
The current symfony quickstart goes through the whole process of creating a symfony application while completely missing to use best practices when adding tracing. I believe it's unlikely that a complete symfony novice would be looking up a quickstart like this, making it quite unnecessary in my opinion.

At the same time, the [SDK bundle](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/Symfony/OtelSdkBundle) already has a readme explaining how to add tracing to your symfony application. Also using a bundle is the best practice and is what 99% of symfony users will want to do.

This is why I propose to ditch the quickstart and instead link to the SDK bundle in the readme, and put all future symfony-specific documentation there.


DrLuke:
Just for the record, your CLA signing process is extremely frustrating and I don't wish to submit any of my personal details like my signature or home address to you. Also it's missing a privacy statement.

bobstrecansky:
@DrLuke - I'm sorry about the CLA - that's the choice of the CNCF, not our choice.  Part of the requirements of working on a CNCF project.

bobstrecansky:
@kishannsangani included this in his PR; we can close this PR for now - this change will be added to our repo shortly.

PR #708, opened 2022-06-13
adding event dispatcher

### LINKED ISSUE #299: Implement Error Handling: Error Handlers must be configurable

brettmc:
To implement the otel spec requirement that ` SDK implementations MUST allow end users to change the library's default error handling behavior for relevant errors`, implement psr-14 compliant event system. It wraps our existing logging mechanism, so by default we still log errors etc as before, but it is now possible to provide any psr-14 implementation and handle events in some other way.
In passing, fix up some `covers` annotations in SpanTest which exposed some uncovered code in `SpanLimits`

Fixes #299
Replaces #603


examples/EventsExample.php
```diff
@@ -0,0 +1,68 @@
...
+ * the library. OpenTelemetry will always attempt to log and continue on error, and may provide no-op implementations
+ * of classes if it was unable to create them as requested.
+ * The default event handlers will use a PSR-3 logger to log events - verbosity and output format can be controlled
+ * through the usual PSR-3 log levels and formatters (if available).
+ * If you wish to do something different with events, you may provide an alternative PSR-14 implementation, but be
+ * aware that you will need to register handlers for all events that you are interested in.
+ */
+
+//logger used by default event handlers
+LoggerHolder::set(
```
brettmc, line 31: @tidal - this might be an opportunity to move `LoggerHolder` somewhere under `SDK\Event` ? A way to get rid of the static singleton could be to create something like an `AbstractLoggableEvent` (or trait) which accepts a `LoggerInterface` and is extended by our events as required?

tests/Unit/SDK/Common/Event/Event/WarningEventTest.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class WarningEventTest extends TestCase
+{
+    public function test_warning_event_with_throwable(): void
+    {
+        $message = 'foo';
+        $exception = new \Exception();
+        $event = new WarningEvent($message, $exception);
+        $this->assertSame($message, $event->getMessage());
+        $this->assertSame($exception, $event->getError());
```
kishannsangani, line 21: Trivial comment, but can we swap lines 21 and 22?

tests/Unit/SDK/Common/Event/Event/WarningEventTest.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->assertTrue($event->hasError());
```
kishannsangani, line 22: Trivial comment, but can we swap lines 21 and 22?

kishannsangani (approved):

kishannsangani (approved):

Nevay (commented):
What is the advantage over the current approach? Users can already configure the behavior by setting the global logger (Java seems to do the [same](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#java)).

brettmc:
> What is the advantage over the current approach? Users can already configure the behavior by setting the global logger (Java seems to do the [same](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#java)).
I agree that just for logging, a PSR-3 implementation such as monolog covers the requirement pretty well (particularly when you can add your own processors and formatters to the logger).

The idea of events came up in a SIG meeting a while back. Aside from being a way to log, the idea was that having events internally would be useful when it came to things like intra-signal messaging (eg, fire a metrics event on batch export with queue size, #records exporter). I believe it can also help with #298.

That said, since PSR-14 doesn't provide a mechanism to register event listeners, it concerns me a little that if we did have our own internal events (eg for metrics), then that functionality would be turned off by default if a user decides to use another PSR-14 package (and the onus is on them to register some/all events to re-enable that functionality). Would it then be safer to keep logging directly to a PSR-3 logger, and not expose our internal eventing?

src/SDK/Common/Event/Handler/ErrorEventHandler.php
```diff
@@ -0,0 +1,18 @@
...
+use OpenTelemetry\SDK\Behavior\LogsMessagesTrait;
+use OpenTelemetry\SDK\Common\Event\Event\ErrorEvent;
+
+class ErrorEventHandler
+{
+    use LogsMessagesTrait;
+
+    public function __invoke(ErrorEvent $event): void
+    {
+        self::logError($event->getError()->getMessage(), ['error' => $event->getError()]);
```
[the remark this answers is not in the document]
brettmc: fixed

brettmc:
Discussed in SIG to shrink this PR back to just providing an event dispatcher. Error logging to be via psr-3 as we currently do.

brettmc:
Further notes for the event system: ensure it stays public, and look at using [CloudEvents](https://cloudevents.io/) as the basis for the events themselves.

bobstrecansky:
Just a quick conflict to resolve with the `composer.json` ^

PR #711, opened 2022-06-15
Cleanup/fix context package

Nevay:
Removed:
`Context::withValue()`: Semi-broken due to creation of hidden scope
`Context::detach()`: Same functionality as `ScopeInterface::detach()`
`Context::attach()`: Same functionality as `Context::activate()`

`Context::getValue()`: Can be removed? Users will use class-specific `::fromContext()` / `::getCurrrent()` methods instead of this method anyways?

Changes:
`Context` and `ContextKey`: Changed to `final` to reduce scope / allow addition of new methods in the future without breaking backwards compatibility.
`ArrayAccessGetterSetter`: Resolves issues with case-insensitive keys and preserves casing as recommended by [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/94c9c75c4f82fbda08bddff02ce9a0c582fdd55c/specification/context/api-propagators.md#set).


src/Context/Propagation/TextMapPropagator.php
```diff
@@ -14,7 +15,9 @@ public static function composite(TextMapPropagatorInterface ...$propagators): Te
             case 0:
                 return NoopTextMapPropagator::getInstance();
             case 1:
-                return $propagators[0];
+                assert($propagators !== []);
```
brettmc, line 18: is this possible, given the `count()` above?

tests/Unit/Context/ContextTest.php
```diff
@@ -146,19 +125,19 @@ public function test_reusing_key_overwrites_value(): void
 
     public function test_ctx_value_not_found_throws(): void
```
brettmc, line 126: I don't think this test's name is accurate anymore. `s/throws/returns_null/` ?

src/Context/Propagation/TextMapPropagator.php
Nevay: No, added to make [psalm happy](https://github.com/open-telemetry/opentelemetry-php/runs/6907417904?check_suite_focus=true#step:11:18).

brettmc: ok. you could also `@psalm-suppress` which might be clearer for future devs...?

Nevay: Seems like I accidentially used `/*` instead of `/**` when I tried that yesterday - changed.

brettmc (approved):

bobstrecansky (approved):

PR #717, opened 2022-06-20
Upgrade deptrac

tidal:
- Upgrades deptrac to version ^0.22.1
- Change to new deptrac config format
- Adds missing dependency checks
- Add deptrac baseline (for false positive)


tidal:
Since he package validation commands in [devtools](https://github.com/opentelemetry-php/dev-tools) rely on the deptrac file for resolving the packages, there needs to be a change in these commands as well to not break CI.

brettmc (approved):

PR #720, opened 2022-06-20
Try making new class for getting/setting current span

### LINKED ISSUE #703: Investigate Trace / Context Interaction
We'd like to investigate [Trace/Context interaction](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction)

We'd like to ensure we have good methods for getting and setting the active span.

Grunet:
May eventually resolves #703 or might just be for organizing discussion on this and will be closed in favor of a fresh PR with the consensus instead.

Some thoughts/questions that came up as I tried this:

- ~~Is a trait the best option here? (adding static methods to existing API classes seemed odd no matter how I thought of it, which is what led me to think of this)~~ Feeling at SIG seemed to be that a class would be preferable. UPDATE: Moved to AbstractSpan for now
- If so, what should the trait and its methods be named? (I leaned generic on the trait name to start with in case it makes sense for other similar methods to be included/composed into it later)
- Should the methods be static- or instance-level? ([the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction) explicitly talks about "static methods", but I wonder if that's a vestige of where they were first made maybe?)
 
Will plan to add in ~~doc headers~~/~~tests~~/examples (see comment below about these) once the shape/direction is agreed upon


Grunet:
So here was my rough thought process for the "SpanUtils" name

1) If people are looking to get the current active span, they're probably not going to type "Trace" in their IDE or search based off of that keyword in general. Instead "Span" seems like it'd be more likely to get searched for in various forms
    - So it has to be "Span<...blah...>" to be consistent with the other classes/interfaces
2) It's not really "managing" anything itself directly, so "Manager" in the name seems like it would be off. 

It's more sort of just proxying functionality around Context that already exists. But also adding "Proxy" in the name seemed awkward cause the thing it's proxying is Context, so the name would have to be something like "SpanContextProxy" which idk, just seems confusing to me

3) Given all that I figured just leaving it fairly generic at "Utils" would sort of capture the helper method nature of it without being too verbose

Grunet:
Not sure why codecov thinks these changes decreased code coverage through TracerProvider? It doesn't even show up if I follow the link into their UI

brettmc:
My $0.05 is that `Span::getCurrent()` is clear, and that's what java does: https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/Span.java#L35 (although they've split `Span` into an API class and an SDK... I think we could achieve this, `AbstractSpan` becomes `API\Span`?

Grunet:
> My $0.05 is that `Span::getCurrent()` is clear, and that's what java does: https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/Span.java#L35 (although they've split `Span` into an API class and an SDK... I think we could achieve this, `AbstractSpan` becomes `API\Span`?

@brettmc oh yeah I didn't realize our setup here matched the Java one so closely. That makes sense to me, I'll go ahead and move the new setter into AbstractSpan and then on a separate PR work out the renaming and possible dismantling of it outlined in #723

Grunet:
Not sure how to handle that new method on the interface in case people have been directly relying on it themselves.

It seems a little awkward as is, but I guess that's part of the point of the follow up to break up AbstractSpan

Grunet:
Oh the symfony and laravel quickstarts already touch on how to use the existing getter, so I guess just some more docs on the setter are needed then?

Grunet:
Use cases I'm noticing in JS and Python for the setter, mostly from their contrib repos

- Several occasions of using it to set a NonRecordingSpan into a new context and passing that context to other operations, mostly propagation related - here's one - https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/propagator/opentelemetry-propagator-ot-trace/src/opentelemetry/propagators/ot_trace/__init__.py
- Start multiple new spans but delay attaching context for any of them immediately - https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/examples/web/examples/document-load/index.js
- Large gap in code between starting a span and actually activating it as the current span - https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts

Maybe given my lack of understanding on how exactly these situations will translate to PHP (or even the situations themselves...), it'd be best to leave off an example for now? And then maybe when the first contrib/instrumentation library needs it and/or faces challenges we can revisit?

Grunet:
And FYI those were all of the changes/investigations I was planning to dig into for now on this. Just lmk what else I should look into/think about!

src/API/Trace/AbstractSpan.php
```diff
@@ -27,6 +27,14 @@ final public static function getCurrent(): SpanInterface
         return self::fromContext(Context::getCurrent());
     }
 
+    /** @inheritDoc */
+    final public static function setCurrent(SpanInterface $span): Context
```
Nevay, line 31: I think we should keep at most two of `::storeInContext()`, `::setCurrent()` and `::activate()`, they all do very similar things.
My preference would be to keep `::storeInContext()` as is, rename `::activate()` to `::makeCurrent()` (and `::getCurrent()` to `::current()`) for consistency with Java.

brettmc:
> Not sure how to handle that new method on the interface in case people have been directly relying on it themselves.
@Grunet you might be able to solve this through our BC layer, it can proxy methods/classes through to their new locations, and will emit deprecation notices: https://github.com/open-telemetry/opentelemetry-php/tree/main/src/SDK/Common/Dev/Compatibility/BC

Grunet:
Ok so I think given the discussion here it sounds like the current functionality of the APIs is probably good for now, but their actual presentation might need a bit of sprucing up.

I think that part can probably be tackled as part of #723, and this PR and the corresponding issue closed.

Thanks everyone for the discussion on this!

PR #721, opened 2022-06-21
OtlpGrpc exporter for metrics

### LINKED ISSUE #700: OTLP exporter for metrics
Currently there is only a [`PrometheusExporter`](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Prometheus/PrometheusExporter.php) available to export metrics.

otel-col can read prometheus metrics and convert them to OTLP, but it would be a lot nicer to avoid this unnecessary conversion step.

erdemkose:
Closes #700 

A new exporter for metrics using OTLP format. It is basically a copy of the existing `OpenTelemetry\Contrib\OtlpGrpc\Exporter` exporter for traces. OTLP conversion is borrowed from `OpenTelemetry\Contrib\Otlp\SpanConverter`.

I also tried to update the metric related classes to use `ResourceInfo` and `InstrumentationScope` as it was defined in the documentation. I am not 100% sure if I understood/followed them correctly.


erdemkose:
It looks like I need to solve API on SDK issue first.

erdemkose:
Closing this PR since it was already implemented in https://github.com/open-telemetry/opentelemetry-php/pull/750

PR #741, opened 2022-07-02
Cleaning up the examples a little

### LINKED ISSUE #426: Revamp examples

Grunet:
Resolves #426 

~~I still need to smoke test the examples (and probably set up a shortcut to do that more easily in the future, as much as is feasible), but~~ mostly looking for initial feedback on if the new folder structure seems improved to folks. 

Otherwise I mostly tried to incorporate the thoughts I gathered from other folks in the issue thread.


examples/traces/demo/src/index.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 declare(strict_types=1);
-require 'vendor/autoload.php';
+require '/../../../../vendor/autoload.php';
```
tidal, line 4: This doesn't work, since the main libraries autoloader is out of the scope of the docker-compose context, and it would not pull in the dependencies of the demo (slim, ect.)

tidal:
@grunet Thanks!!!

There is only one small issue, which needs to be fixed, otherwise I'd merge this PR as is, as it already improves the examples by a lot.
For any additional improvements I'd just create another ticket.

examples/traces/demo/src/index.php
Grunet: Yup good call, switched it back.

I also was curious and tried to get this demo working, but got stuck after updating the tags on the opentelemetry-php-base images ("latest" was pulling in an image from 7 months ago) and exposing the 4317 port on the collector. 

Have a hunch something's off with the exporting from PHP but couldn't figure out how to troubleshoot (since this example wants to do all the config via environment variables and the OTEL_LOG_LEVEL one isn't supported yet afaict)

Grunet:
I also added in a make target to just try and run through more of the examples to make sure nothing crashes (I also found a bug along the way! Another incorrect path issue). Figured it might be a helpful sanity check in the future w/o a super high maintenance burden.

examples/traces/demo/src/index.php
tidal: Yap, I tried the demo as well a few days ago, and it didn't really work. However I used it about 2 months ago for a Otel workshop at work, and it worked fine.
Maybe @brettmc can have a loo at it, once he is back.

tidal:
> Figured it might be a helpful sanity check in the future w/o a super high maintenance burden (if there's a clever way not to have to list individual example files out too that'd be cool to learn about too)
Yeah, I have also thought about that, since for beta we should make sure very example really works, otherwise it could be confusing or frustrating for users. 

If you would like to work on that, you could add a command to https://github.com/opentelemetry-php/dev-tools.
I'm happy to help you with everything there, if you need help. 

For finding the respective file, one could use [FilesystemIterator](https://www.php.net/manual/de/class.filesystemiterator.php) or Symfony's [Finder Component](https://symfony.com/doc/current/components/finder.html).

To run the tests in an isolated process one could use Symfony's [Process Component](https://symfony.com/doc/current/components/finder.html)

Let me know, if that's somthing you would be interested in.

Other than that, I'll merge this PR. :)

tidal (approved):

Grunet:
> If you would like to work on that, you could add a command to https://github.com/opentelemetry-php/dev-tools. I'm happy to help you with everything there, if you need help.
> 
> For finding the respective file, one could use [FilesystemIterator](https://www.php.net/manual/de/class.filesystemiterator.php) or Symfony's [Finder Component](https://symfony.com/doc/current/components/finder.html).
> 
> To run the tests in an isolated process one could use Symfony's [Process Component](https://symfony.com/doc/current/components/finder.html)
> 
> Let me know, if that's somthing you would be interested in.
> 
> Other than that, I'll merge this PR. :)

Oh yeah interesting I hadn't thought of actually using PHP to manage that! (I was just trying to think a little if there was some clear 1 liner makefile thing, or shell script thing, but gave up pretty quick)

I think it's probably ok for now imo (seems safe/quick to refactor if it gets unwieldy). 

I'll go ahead and take a look at what else I can maybe help with amongst the issues

examples/traces/demo/src/index.php
brettmc: I sure can take a look at it, right now.

brettmc: @Grunet @tidal I found a regression/bug. It's kinda hard to write a test for, since it involves a TracerProvider being destroyed as soon as it is created (I think introduced in the fix for #716 )
In short, this works:
```php
$tracerProvider = (new TracerProviderFactory('example'))->create();
$tracer = $tracerProvider->getTracer('io.opentelemetry.contrib.php');
```
and this doesn't work:
```php
$tracer = (new TracerProviderFactory('example'))->create()->getTracer('io.opentelemetry.contrib.php');
```

I'll take this up on slack...

Issue #732, opened 2022-06-27
Invalid @covers annotations in SpanTest create warning in PHPunit runs


tidal:
**Steps to reproduce**
vendor/bin/phpunit tests/Unit/SDK/Trace/SpanTest.php

**What is the expected behavior?**
No warning

**What is the actual behavior?**

warning:
`"@covers ..." is invalid

amber0612:
@tidal Please assign this to me

PR #752, opened 2022-07-06
Invalid @covers annotations in SpanTest create warning in PHPunit runs

amber0612:
#732


tests/Unit/SDK/Trace/SpanTest.php
```diff
@@ -247,7 +247,7 @@ public function test_end_twice(): void
 
     /**
      * @group trace-compliance
-     * @covers ::isRecording
```
brettmc, line 250: can be resolved by adding a `@coversDefaultClass` in the class-level docblock

bobstrecansky:
Reran these tests after the fix; looks better now.

bobstrecansky:
I think the only thing left on this PR is @brettmc 's comments.

brettmc (approved):

PR #759, opened 2022-07-11
Tracer references tests

brettmc:
a recent bug in `examples/traces/demo` highlighted that if there is no reference to a tracer provider, then the shared state will shutdown, even if that shared state is being used by active tracers. Adding a phpdoc comment explaining this behavior, and some tests to demonstrate it

### DIFF
--- a/examples/traces/demo/src/index.php
+++ b/examples/traces/demo/src/index.php
@@ -70,12 +70,12 @@ $app = Bridge::create($container);
 
 //middleware starts root span based on route pattern, sets status from http code
 $app->add(function (Request $request, RequestHandler $handler) use ($tracer) {
-    $carrier = TraceContextPropagator::getInstance()->extract($request->getHeaders());
+    $parent = TraceContextPropagator::getInstance()->extract($request->getHeaders());
     $routeContext = RouteContext::fromRequest($request);
     $route = $routeContext->getRoute();
     $root = $tracer->spanBuilder($route->getPattern())
         ->setStartTimestamp((int) ($request->getServerParams()['REQUEST_TIME_FLOAT'] * 1e9))
-        ->setParent($carrier)
+        ->setParent($parent)
         ->setSpanKind(SpanKind::KIND_SERVER)
         ->startSpan();
     $root->activate();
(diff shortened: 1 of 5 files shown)


brettmc:
This documents and adds tests for the current behavior, but I did find the current behavior surprising. I think it's a fairly common practise to avoid creating intermediate objects, but in this case it has a side-effect.
Another option might be that `TracerProvider::shutdown` is not responsible for destroying shared state, but instead drops its own reference to shared state - thus, if the shared state is in use by a tracer, it can continue to be used until all tracers are removed. That may have been the original behavior prior to #716
Nevay's initial thoughts were:

> we can postpone the shutdown until all tracers are out of scope by storing the tracer provider in every tracer, not sure what is better (I'm slightly in favor of keeping the current behavior and just documenting it; mainly because explicit shutdowns are no longer possible if the tracer provider goes out of scope)

The spec doesn't say that `TracerProvider::shutdown` should affect tracers, only that it should provide NoopTracers after being shut down...

Nevay:
> The spec doesn't say that TracerProvider::shutdown should affect tracers, only that it should provide NoopTracers after being shut down...

`Tracer`s are effectively shut down due to shutting down `SpanProcessor` & co.

[TracerProvider shutdown](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown)
> Shutdown MUST be implemented at least by invoking Shutdown within all internal processors.

[SpanProcessor shutdown](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown-1)
> Shutdown SHOULD be called only once for each SpanProcessor instance. After the call to Shutdown, subsequent calls to OnStart, OnEnd, or ForceFlush are not allowed. SDKs SHOULD ignore these calls gracefully, if possible.

---

> I did find the current behavior surprising. I think it's a fairly common practise to avoid creating intermediate objects, but in this case it has a side-effect.

It depends on whether we want users to rely on the implicit shutdown behavior and expect them to not call `::shutdown()` at all; or treat it only as a fallback to try to export traces if something unexpected happens (exception / process timeout) and expect users to keep a reference to the `TracerProvider` and call `::shutdown()` explicitely under normal circumstances.

> thus, if the shared state is in use by a tracer, it can continue to be used until all tracers are removed.

Moving shutdown handling to shared state might be the best option for now.

bobstrecansky (approved):

PR #760, opened 2022-07-11
Remove automatic `::shutdown()`, add `ShutdownHandler::register()`

Nevay:
See #759, removes automatic shutdown handling. Users should keep a reference to the `TracerProvider` and call `::shutdown()` explicitely. `ShutdownHandler::register([$tracerProvider, 'shutdown'])` can be used to protect against process timeout / exit, but also requires users to keep a reference to the `TracerProvider`.

The tracer provider should be in scope anyways to be globally accessible in the current execution context (somewhat depends on how instrumentation registration is solved?).


tests/Integration/SDK/TracerProviderTest.php
```diff
@@ -17,12 +18,12 @@ class TracerProviderTest extends TestCase
...
-    public function test_tracer_shuts_down_immediately_when_out_of_scope(): void
+    public function test_tracer_provider_does_not_shut_down_when_out_of_scope_with_tracer_in_scope(): void
     {
         $prophet = new Prophet();
         $spanProcessor = $prophet->prophesize(SpanProcessorInterface::class);
         // @phpstan-ignore-next-line
-        $spanProcessor->shutdown()->shouldBeCalledTimes(1);
+        $spanProcessor->shutdown()->shouldBeCalledTimes(0);
 
         /* Because no reference is kept to the TracerProvider, it will immediately __destruct and shutdown,
```
brettmc, line 28: this comment is no longer true

brettmc (approved):

Nevay:
> shutdown is called if no new traces can be created

This approach won't work for metrics; the metrics API allows registration of permanent callbacks. [Asynchronous Instrument API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-instrument-api):
> The API MUST support creation of asynchronous instruments by passing zero or more callback functions to be permanently registered to the newly created instrument.

I see three options:
1. shutdown tracer provider after no new traces can be created, shut down meter provider after last synchronous writer (and non-permanent asynchronous callback?) goes out of scope, which might be unexpected if the user registers only permanent callbacks
2. keep the current behavior and shut down when tracer provider / meter provider are destroyed
3. remove automatic shutdown behavior (at least on `__destruct()`, keep `register_shutdown_function()` handler?)

We should also consider that `::shutdown()` is a blocking operation, triggering this at an arbitrary point might lead to an unexpected delay for the application.

src/SDK/Common/Util/ShutdownHandler.php
```diff
@@ -0,0 +1,130 @@
...
+            $object = $target;
+            $cleanup = $weakMap[$object] ??= new self();
+        }
+        $cleanup->ids[] = $handlerId;
+    }
+
+    /**
+     * @see https://github.com/amphp/amp/blob/f682341c856b1f688026f787bef4f77eaa5c7970/src/functions.php#L140-L191
+     */
+    private static function weaken(Closure $closure, ?object &$target = null): Closure
```
Nevay, line 75: This is also very useful for the metrics API; we can add a `$weak` parameter to asynchronous instruments `::observe()` method that would allow users to provide normal callbacks with bound `$this` instead of forcing them to take care of weakening the `$this` reference themselves.

Nevay (commented):
> Does that resurrect any tracer provider leaks that you fixed up in the earlier PR?

No, `__destruct()` could simply call `::unregisterShutdownFunction()` instead of `::shutdown()`.

Converting this to draft for now; I've decided to create a POC that allows users to opt-in instead of enabling it by default.
```php
ShutdownHandler::register($tracerProvider->shutdown(...)); // Closure::fromCallable([$provider, 'shutdown']) if php < 8.1
```

brettmc (approved):

brettmc:
@Nevay let us know if it's ok to merge this now.

Nevay:
Yes, this can be merged.

PR #763, opened 2022-07-12
Adding event dispatcher

brettmc:
When discussing #708 in SIG last week, we decided that using it to perform logging was overkill (since psr-3 can provide enough control to meet the spec requirements).
This implements a simplified event dispatcher, based loosely on psr-14 and using `CloudEvents` to represent events (as suggested during SIG)

Replaces #708


bobstrecansky (approved):

composer.json
```diff
@@ -8,6 +8,7 @@
     "license": "Apache-2.0",
     "require": {
         "php": "^7.4 || ^8.0",
+        "cloudevents/sdk-php": "^v1.0.1",
```
tidal, line 11: This needs to be added to the API package's composer config as well.
(We should have a test which checks for this, for example running the unit tests of a package with its composer file.)

src/API/Behavior/EmitsEventsTrait.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Behavior;
```
tidal, line 5: I think this would be better placed in the `OpenTelemetry\API\Common\Event`namespace as well.
I know, I introduced these `Behavior` namespaces, but this was before we had the `Common` ones.

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Event;
+
+use CloudEvents\V1\CloudEventInterface;
+use OpenTelemetry\Context\Context;
+use OpenTelemetry\Context\ContextKey;
+
+class Dispatcher
```
tidal, line 11: This class needs an interface.
This is the API package after all and an interface would help to avoid a hard dependency on the implementation.

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    private static ?ContextKey $key = null;
+    /** @var array<string, array<int, array<callable>>> */
+    private array $listeners = [];
+
+    private function __construct()
```
tidal, line 17: Please, please do not make this a singleton. If we need a static reference to an instance, we already have it with the method below, and there are other ways to solve this in the future.
In fact I can think of a lot of places where an additional dispatcher would make sense.

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+    }
+
+    public static function getInstance(): self
+    {
+        self::$key ??= new ContextKey(self::class);
```
tidal, line 23: Could you make this a separate method (`getConstantKeyInstance`)?

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+        $dispatcher = Context::getCurrent()->get(self::$key);
+        if ($dispatcher === null) {
+            $dispatcher = new self();
+            Context::getCurrent()->with(self::$key, $dispatcher)->activate();
```
tidal, line 27: Could you make this a separate method (`createInstance`)?

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        return $dispatcher;
+    }
+
+    /**
+     * @internal
+     */
+    public function reset(): void
```
tidal, line 36: What is this method needed for?

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $this->listeners = [];
+    }
+
+    public function getListenersForEvent(CloudEventInterface $event): iterable
```
tidal, line 41: Exposing a `Generator` on a public method can lead to bugs, as they can be exhausted when used.

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+    public function listen(string $type, callable $listener, int $priority = 0): void
+    {
+        $this->listeners[$type][$priority][] = $listener;
+        ksort($this->listeners[$type]);
+    }
+
+    public function dispatch(CloudEventInterface $event): object
+    {
+        $listeners = $this->getListenersForEvent($event);
+        $this->dispatchEvent($listeners, $event);
```
tidal, line 59: Can you ake this one call?

```php
        $this->dispatchEvent(
            $this->getListenersForEvent($event), 
            $event
        );
```

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,70 @@
...
+
+        return $event;
```
tidal, line 61: Why does the method need to return its passed argument?

tidal:
> When discussing https://github.com/open-telemetry/opentelemetry-php/pull/708 in SIG last week, we decided that using it to perform logging was overkill (since psr-3 can provide enough control to meet the spec requirements).

I don't know what the spec requirements that should be in that case. However I see the spec as a minimum requirement which must be fulfilled to get to the fun parts. Using an event listener for logs would allow for more fine grained control, but this could still be implemented for (additional logs) in the future.

src/API/Common/Event/Dispatcher.php
brettmc: it doesn't, fixed.

src/API/Common/Event/Dispatcher.php
brettmc: done

src/API/Common/Event/Dispatcher.php
brettmc: I've made the method private (it was just public for ease of testing). Is it safe enough behind a private method, or should I remove the generator?

src/API/Common/Event/Dispatcher.php
brettmc: working around singleton - no longer required & removed

src/API/Common/Event/Dispatcher.php
brettmc: si. done.

src/API/Common/Event/Dispatcher.php
brettmc: done

src/API/Common/Event/Dispatcher.php
brettmc: done

src/API/Common/Event/Dispatcher.php
brettmc: done

src/API/Behavior/EmitsEventsTrait.php
brettmc: moved

composer.json
brettmc: done

brettmc:
> After reading through the CloudEvents spec: I think that using PSR-14 (as initially done in https://github.com/open-telemetry/opentelemetry-php/pull/708) makes more sense, it is widely used and would allow users to use a single event system in their application.

@Nevay I thought that initially, too. But without an interface in PSR-14 for registering events in a listener, it wasn't possible to have our code register its own events in a user-provided PSR-14 implementation. Effectively, you could bring your own psr-14, but you'd then be responsible for knowing about and hooking up of our internal events, including future events that we come up with. I'm happy to be corrected here, though - we're all for standards and choice

brettmc:
@tidal I think I've addressed all review comments now

tidal:
@brettmc 
Thanks!!!

src/API/Common/Event/Dispatcher.php
```diff
@@ -0,0 +1,63 @@
...
+
+    private static function getConstantKeyInstance(): ContextKey
+    {
+        return self::$key ??= new ContextKey(self::class);
+    }
+
+    private static function createInstance(): self
+    {
+        $dispatcher = new self();
+        Context::getCurrent()->with(self::getConstantKeyInstance(), $dispatcher)->activate();
```
Nevay, line 59: The "root" dispatcher should use a static instance and not the context; currently leaks the created scope.
Additionally leads to different behavior depending on when the dispatcher is retrieved for the first time as listeners will then be added to the same instance.

PR #765, opened 2022-07-14
return no-op span builder after shutdown

brettmc:
this completes a todo in code comments: when tracer shared state has shutdown, getting a span builder should return a no-op implementation

### DIFF
--- a/src/SDK/Trace/Tracer.php
+++ b/src/SDK/Trace/Tracer.php
@@ -6,6 +6,7 @@ namespace OpenTelemetry\SDK\Trace;
 
 use function ctype_space;
 use OpenTelemetry\API\Trace as API;
+use OpenTelemetry\Context\Context;
 use OpenTelemetry\SDK\Common\Instrumentation\InstrumentationScopeInterface;
 
 class Tracer implements API\TracerInterface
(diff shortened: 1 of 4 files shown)


brettmc:
NB that I needed to adjust a bunch of tests, as tracer shared state was immediately shutting down due to their not maintaining a reference to the tracer provider. Should be fixed by #760

src/SDK/Trace/Tracer.php
```diff
@@ -34,7 +35,7 @@ public function spanBuilder(string $spanName): API\SpanBuilderInterface
         }
 
         if ($this->tracerSharedState->hasShutdown()) {
-            // TODO: Return a noop tracer
+            return new API\NoopSpanBuilder(Context::storage());
```
tidal, line 38: Can you add this to the class' unit tests?

brettmc: done

tidal (approved):

tidal:
There is a huge drop in code-coverage, which is unrelated. However [SpanTest](https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/Unit/SDK/Trace/SpanTest.php) has only a @coversDefaultClass annotation, which is never used, so all those tests are completely ignored by code-coverage.

Issue #508, opened 2021-12-09
Create unified HTTP Client (Wrapper)


tidal:
In order to reduce dependencies in the HTTP exporters as unified HTTP Client (Wrapper) should be created.
The HTTP Client (Wrapper) should be configurable by users, if needed.
The HTTP Client (Wrapper) should be usable by other components (for exmple detectors).
The HTTP Client (Wrapper) must be able to work with PSR compatible implementations.

tidal:
@stale
ickwarnda

brettmc:
I believe we can close this, as `PsrTransport` achieves this?

PR #767, opened 2022-07-16
Add DsnParser + Otlp/HttpEndpointResolver and remove dependency on nyholm/dsn

tidal:
This PR removes the dependency on the nyholm/dsn package by introducing dedicated DsnParser + Otlp/HttpEndpointResolver components. 
The DSN we use in the Trace/ExporterFactory (and probably other factories in the future) uses a `protocol=type+scheme` pattern to identify different exporters. This is not supported by nyholm/dsn, so there had to be custom logic in the first place.
The other place where the package was used was resolving the OTLP HTTP endpoints according to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp). This logic has been used into a dedicated class, which can be used for exporters of other signals in the future.

- Adds Dsn, Dsn/Parser and Dsn/Factory.
- Adds Otlp/HttpEndpointResolver
- Removes dependency on nyhol/dsn

(Note: There were some outdated assumptions in the OTLPHttpExporterTest which dated back to before the introduction of signal specific Env Vars. These have just been removed for now. The OTLExporter tests need some rework anyway, which will be provided in a future PR.)

related:
- https://github.com/open-telemetry/opentelemetry-php/issues/508
- https://github.com/open-telemetry/opentelemetry-php/issues/451


tidal:
Phan error should be resolved, when https://github.com/open-telemetry/opentelemetry-php/pull/768 is merged.

Nevay:
[enqueue/dsn](https://github.com/php-enqueue/dsn) supports scheme extensions / their use case is very similar to ours. I think we should use [ConnectionFactoryFactory](https://github.com/php-enqueue/enqueue/blob/master/ConnectionFactoryFactory.php) as orientation for our implementation.

src/SDK/Common/Otlp/HttpEndpointResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+namespace OpenTelemetry\SDK\Common\Otlp;
+
+use OpenTelemetry\API\Common\Signal\Signals;
+use Psr\Http\Message\UriInterface;
+
+interface HttpEndpointResolverInterface
+{
+    public const TRACE_DEFAULT_PATH = 'v1/traces';
+    public const METRICS_DEFAULT_PATH = 'v1/traces';
```
brettmc, line 13: `v1/metrics` and `v1/logs` ?

src/SDK/Common/Otlp/HttpEndpointResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+    public const LOGS_DEFAULT_PATH = 'v1/traces';
+    public const DEFAULT_PATHS = [
+        SIGNALS::TRACE => self::TRACE_DEFAULT_PATH,
```
brettmc, line 16:
```suggestion
        Signals::TRACE => self::TRACE_DEFAULT_PATH,
```

src/SDK/Common/Otlp/HttpEndpointResolverInterface.php
tidal: Ohh, good catch!!!

src/SDK/Common/Otlp/HttpEndpointResolverInterface.php
tidal: Fixed and added a test for it.

src/SDK/Common/Dsn/Factory.php
```diff
@@ -0,0 +1,70 @@
...
+
+    public static function create(array $defaults = []): self
+    {
+        return new self($defaults);
+    }
+
+    /**
+     * @phan-suppress PhanUndeclaredClassAttribute
+     */
+    #[\ReturnTypeWillChange]
```
brettmc, line 33: will it change? I usually see this used to quieten 8.1 deprecations warnings, but this looks like it's returning the type is should

brettmc (approved):

brettmc:
looks good to me, pending Nevay's comment about enqueue/dsn ?

src/SDK/Common/Dsn/Factory.php
tidal: Yes, the interface defines a return type of `DsnInterface`

tidal:
>[enqueue/dsn](https://github.com/php-enqueue/dsn) supports scheme extensions / their use case is very similar to ours. I think we should use [ConnectionFactoryFactory](https://github.com/php-enqueue/enqueue/blob/master/ConnectionFactoryFactory.php) as orientation for our implementation.

I guess you mean the ExporterFactory. I can look into that.

PR #780, opened 2022-07-19
store root dispatcher in a static var

### LINKED ISSUE #778: `Event\Dispatcher::getInstance()` leaks scope
**Steps to reproduce**
```php
$key = Context::createKey('-');
$scope = Context::getCurrent()->with($key, 1)->activate();
Dispatcher::getInstance();
$result = $scope->detach();

assert($result === 0); // currently MISMATCH | 1
assert(Context::getCurrent()->get($key) === null); // currently 1
```

`Event\Dispatcher::getInstance()` does not create a scope.

`Event\Dispatcher::getInstance()` creates a scope that is never detached.

**Additional context**
See also https://github.com/open-telemetry/opentelemetry-php/pull/763#discussion_r923037358.

brettmc:
This seems to be the safest way to store the main dispatcher. Using context to store the main event dispatcher causes a scope to be created: activating the newly-created context which stores dispatcher causes a new scope, and that can break things.
I also looked at storing dispatcher in the root context, but that can't work because contexts are immutable.

Closes #778

### DIFF
--- a/src/API/Common/Event/Dispatcher.php
+++ b/src/API/Common/Event/Dispatcher.php
@@ -5,20 +5,16 @@ declare(strict_types=1);
 namespace OpenTelemetry\API\Common\Event;
 
 use CloudEvents\V1\CloudEventInterface;
-use OpenTelemetry\Context\Context;
-use OpenTelemetry\Context\ContextKey;
 
 class Dispatcher implements DispatcherInterface
 {
-    private static ?ContextKey $key = null;
+    private static ?self $root = null;
     /** @var array<string, array<int, array<callable>>> */
     private array $listeners = [];
 
-    public static function getInstance(): self
+    public static function getRoot(): self
     {
-        $key = self::getConstantKeyInstance();
-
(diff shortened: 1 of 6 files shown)


src/API/Common/Event/Dispatcher.php
```diff
@@ -5,20 +5,16 @@
...
 
 class Dispatcher implements DispatcherInterface
 {
-    private static ?ContextKey $key = null;
+    private static ?self $root = null;
     /** @var array<string, array<int, array<callable>>> */
     private array $listeners = [];
 
-    public static function getInstance(): self
+    public static function getRoot(): self
```
brettmc, line 15: I chose `root` because it sounds less singleton-y, since you can in fact have as many dispatchers as you wish. This is just the default one.

tidal:
Oki.
I will take care of providing a static referenceto the dispatcher  with the API/SDK anyway (soone done) , because that's needed  for the instrumentations. However it doesn't hurt if it has its ownone.

tidal (approved):

PR #784, opened 2022-07-25
Use of "iterable" Type May Need Re-Evaluation Across the Repo 

amber0612:


tests/Unit/SDK/Common/Util/FunctionsTest.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $this->assertTrue(isEmpty([]));
+    }
+
+    public function test_for_exhausted_generator(): void
+    {
+        $generator = gen();
+        foreach ($generator as $value) {
+        }
+        $value=isEmpty($generator) ? true : false;
```
Nevay, line 30: Obsolete ternary operator

Nevay (commented):
Failing test:
```php
assert(isEmpty((fn() => yield from [])()));
```

tests/Unit/SDK/Common/Util/FunctionsTest.php
amber0612: Resolved

Nevay:
Another failing test:
```php
$generator = (fn() => yield from [1, 2])();
$generator->next();
assert(!isEmpty($generator));
```

amber0612:
$generator = (fn() => yield from [1, 2])();
$generator->next();
assert(!isEmpty($generator));

This is a non empty generator case and hence isEmpty() will return false. Hence the test case is passing

Nevay:
> isEmpty() will return false

Currently it returns true (`foreach` tries to rewind the generator and triggers an exception).

---

What is the expected result for an exhausted rewindable iterator?
```php
$iterator = new ArrayIterator([1]);
$iterator->next();
$empty = isEmpty($iterator); // ?
```

Should `isEmpty()` be allowed to modify the given iterable (somewhat obsolete if we mark the function as internal)?
```php
$iterator = new ArrayIterator([1, 2]);
$iterator->next();
isEmpty($iterator);
$value = $iterator->current(); // ?
```

Side note: We could simply remove the [special handling of an empty batch](https://github.com/open-telemetry/opentelemetry-php/blob/bf912ed7bd58a7a9bfb034bcf2907b2789bdb4e2/src/SDK/Trace/Behavior/SpanExporterTrait.php#L44-L46) and the `isEmpty()` function, our span processors call `SpanExporter::export()` with a non-empty batch / exporters can skip empty exports after conversion.

amber0612:
> 
Makes sense. The iterable will be modified.

Nevay (approved):

brettmc (approved):

PR #788, opened 2022-08-02
Prevent concurrent `::export()` calls in span processor

Nevay:
> [Export()](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch) will never be called concurrently for the same exporter instance.

This PR calls `::export()` in a synchronous loop to prevent concurrent calls to `::export()`.
Note that this might block the `Span::end()` call that triggered the export longer. The blocking autoflush behavior of `BatchSpanProcessor::onEnd()` can be disabled, spans will then only be flushed on explicit `::forceFlush()` / `::shutdown()` calls.
> [OnEnd()](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onendspan) MUST be called synchronously within the [Span.End() API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#end), therefore it should not block or throw an exception.

Disallowing concurrent export calls means that we have to reintroduce support for concurrent exports by changing the return type of `::export()` to a promise/future-like value - I plan on working on that after this PR is merged.

Alternative to #787, see also https://github.com/open-telemetry/opentelemetry-php/pull/787#discussion_r935643354.

### DIFF
--- a/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -8,6 +8,7 @@ use function assert;
 use function count;
 use InvalidArgumentException;
 use OpenTelemetry\Context\Context;
+use OpenTelemetry\SDK\Behavior\LogsMessagesTrait;
 use OpenTelemetry\SDK\Common\Future\CancellationInterface;
 use OpenTelemetry\SDK\Common\Time\ClockInterface;
 use OpenTelemetry\SDK\Trace\ReadableSpanInterface;
@@ -15,15 +16,13 @@ use OpenTelemetry\SDK\Trace\ReadWriteSpanInterface;
 use OpenTelemetry\SDK\Trace\SpanDataInterface;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
 use OpenTelemetry\SDK\Trace\SpanProcessorInterface;
-use Psr\Log\LoggerAwareInterface;
-use Psr\Log\LoggerAwareTrait;
 use SplQueue;
 use function sprintf;
 use Throwable;
 
-class BatchSpanProcessor implements SpanProcessorInterface, LoggerAwareInterface
+class BatchSpanProcessor implements SpanProcessorInterface
 {
-    use LoggerAwareTrait;
+    use LogsMessagesTrait;
 
     public const DEFAULT_SCHEDULE_DELAY = 5000;
     public const DEFAULT_EXPORT_TIMEOUT = 30000;
@@ -166,9 +165,7 @@ class BatchSpanProcessor implements SpanProcessorInterface, LoggerAwareInterface
 
                             continue;
                         }
-                        if ($this->logger !== null) {
-                            $this->logger->error(sprintf('Unhandled %s error', $flushMethod), ['exception' => $e]);
-                        }
+                        self::logError(sprintf('Unhandled %s error', $flushMethod), ['exception' => $e]);
                     }
                 }
 
@@ -185,9 +182,7 @@ class BatchSpanProcessor implements SpanProcessorInterface, LoggerAwareInterface
                 try {
                     $this->exporter->export($this->queue->dequeue())->await();
                 } catch (Throwable $e) {
-                    if ($this->logger !== null) {
-                        $this->logger->error('Unhandled export error', ['exception' => $e]);
-                    }
+                    self::logError('Unhandled export error', ['exception' => $e]);
                 } finally {
                     $this->queueSize -= $batchSize;
                 }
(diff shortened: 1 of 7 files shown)


src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -4,142 +4,183 @@
...
-        int $scheduledDelayMillis = null,
-        int $exporterTimeoutMillis = null,
-        int $maxExportBatchSize = null
+        SpanExporterInterface $exporter,
+        ClockInterface $clock,
+        int $maxQueueSize = self::DEFAULT_MAX_QUEUE_SIZE,
+        int $scheduledDelayMillis = self::DEFAULT_SCHEDULE_DELAY,
+        int $exportTimeoutMillis = self::DEFAULT_EXPORT_TIMEOUT,
+        int $maxExportBatchSize = self::DEFAULT_MAX_EXPORT_BATCH_SIZE,
+        bool $autoFlush = true
```
Nevay, line 56: Should be changed to default `false` before beta to be consistent with async implementations that will only export once the event loop gains control. Using default `true` for now to keep the scope of the PR small / to not break examples etc.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -4,142 +4,183 @@
...
         }
     }
 
-    /** @inheritDoc */
     public function forceFlush(?CancellationInterface $cancellation = null): bool
     {
-        if (!$this->running || $this->exporter === null) {
-            return true;
+        if ($this->closed) {
+            return $this->flush->isEmpty();
```
Nevay, line 117: Not too sure about this, should either return `false` or call `::flush()` to allow recovering from a shutdown that was terminated by an exception.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -4,142 +4,183 @@
...
         }
-
-        if (!$this->running) {
+        if (!$span->getContext()->isSampled()) {
             return;
         }
 
-        if ($span->getContext()->isSampled() && !$this->queueReachedLimit()) {
-            $this->queue[] = $span->toSpanData();
+        if (count($this->queue) === $this->maxQueueSize && count($this->batch) === $this->maxBatchSize) {
```
brettmc, line 99: should this be `>=` ?

Nevay: This defines the upper limit of the queue and cannot be exceeded.

Side note: Currently it allows `maxQueueSize` + up to `maxExportBatchSize` (size of in-flight batch) not exported spans, will change it to consider in-flight batches for consistency. We need this change anyways as soon as we add concurrent exports in a follow up PR.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
brettmc: it makes sense that it would try to call `flush` again, the method is _force_ flush

Nevay: After rewriting the async implementation to match this PR: I think that preventing termination of the worker by swallowing/logging `::export()` exceptions (and `::forceFlush()` / `::shutdown()` exceptions if we cannot rethrow them into the original caller) makes the most sense. This would make recovery post-shutdown obsolete.

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -4,142 +4,218 @@
...
+use Psr\Log\LoggerAwareTrait;
+use SplQueue;
+use function sprintf;
+use Throwable;
 
-class BatchSpanProcessor implements SpanProcessorInterface
+class BatchSpanProcessor implements SpanProcessorInterface, LoggerAwareInterface
 {
-    use EnvironmentVariablesTrait;
+    use LoggerAwareTrait;
```
brettmc, line 26: we have `LogsMessagesTrait` as a standard mechanism for logging

src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -4,142 +4,218 @@
...
+        }
+
+        $this->exporter = $exporter;
+        $this->clock = $clock;
+        $this->maxQueueSize = $maxQueueSize;
+        $this->scheduledDelayNanos = $scheduledDelayMillis * 1_000_000;
+        $this->maxExportBatchSize = $maxExportBatchSize;
+        $this->autoFlush = $autoFlush;
+
+        $this->queue = new SplQueue();
```
bobstrecansky, line 85: TIL about https://www.php.net/manual/en/class.splqueue.php

brettmc (approved):

PR #790, opened 2022-08-05
Add support for concurrent exports

Nevay:
Related to #788. 

> The return of [Export()](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch) is implementation specific. In what is idiomatic for the language the Exporter must send an ExportResult to the Processor. ExportResult has values of either Success or Failure

### DIFF
--- a/src/SDK/Common/Future/CompletedFuture.php
+++ b/src/SDK/Common/Future/CompletedFuture.php
@@ -21,7 +21,7 @@ final class CompletedFuture implements FutureInterface
         $this->value = $value;
     }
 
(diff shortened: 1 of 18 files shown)


brettmc:
Looks good to me. Futures look quite similar to Promises, I was going to ask why not use promises (esp given recent cancellation work), but I think I talked myself out of it.
I'll wait a day or two to see if anybody else wants to also review.

brettmc (approved):

bobstrecansky (approved):

PR #798, opened 2022-08-11
use a stable released version of psalm

brettmc:
master breaks our builds occasionally, so switch to a tagged release. Use v4.26.0 which is the latest stable.


brettmc:
discussion with other maintainers... since we already run psalm in our `php` action, and the serif uploading is only marginally useful (we have `make all` now, so devs can and should run their tests locally first), I will just remove the psalm security analysis step.

PR #804, opened 2022-08-23
gitsplit contrib components

brettmc:
Split components from `contrib`, so they can be published to packagist independently.


src/Contrib/Context/Swoole/composer.json
```diff
@@ -0,0 +1,23 @@
...
+  "authors": [
+    {
+      "name": "opentelemetry-php contributors",
+      "homepage": "https://github.com/open-telemetry/opentelemetry-php/graphs/contributors"
+    }
+  ],
+  "require": {
+    "php": "^7.4 || ^8.0",
+    "open-telemetry/context": "self.version",
+    "ext-openswoole": "^4"
```
brettmc, line 16: @Nevay - this was an educated guess as to the swoole requirements

.gitsplit.yml
```diff
@@ -20,6 +20,22 @@ splits:
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/sdk.git"
   - prefix: "src/Contrib"
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/sdk-contrib.git"
+  - prefix: "src/Contrib/Jaeger"
```
kishannsangani, line 23: With the recent PR merged, there is an addition for `src/Extension` in .gitsplit.yml.

src/Contrib/Context/Swoole/composer.json
Nevay: I don't think we should add this constraint, the code should be compatible with `ext-swoole` and `ext-openswoole`.

brettmc: removed, in favour of mentioning both swoole variants in description and tags.

src/Contrib/Otlp/composer.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "open-telemetry/exporter-otlp-common",
+  "description": "OTLP common code for OpenTelemetry exporters.",
```
brettmc, line 3: I've just realised that this is also where otel metrics exporter lives. This will need some thought and probably moving things around.

src/Contrib/Zipkin/composer.json
```diff
@@ -0,0 +1,26 @@
...
+    "php": "^7.4 || ^8.0",
+    "open-telemetry/api": "self.version",
+    "open-telemetry/sdk": "self.version",
+    "php-http/async-client-implementation": "^1.0",
+    "php-http/discovery": "^1.14",
+    "psr/http-factory-implementation": "^1.0"
+  },
+  "autoload": {
+    "psr-4": {
+      "OpenTelemetry\\Contrib\\Newrelic\\": "."
```
kishannsangani, line 23: The path should be "OpenTelemetry\\Contrib\\Zipkin\\".

kishannsangani (changes requested):

kishannsangani (approved):

src/Contrib/Otlp/composer.json
brettmc: on reflection, I think the metrics stuff does belong here and it's not a problem.

bobstrecansky (approved):

PR #820, opened 2022-09-09
 Add interfaces for `Context` and `ContextKey`

### LINKED ISSUE #795: Add interfaces for `Context` and `ContextKey`
Simplifies mocking and allows using `enum`s as `ContextKey`.

Nevay:
Resolves #795.


src/API/Trace/SpanBuilderInterface.php
```diff
@@ -15,14 +15,14 @@
...
-     * Sets the parent {@see Context} to use.
+     * Sets the parent {@see ContextInterface} to use.
      *
      * If no {@see SpanInterface} is available in the provided context, the resulting span will become a root span,
      * as if {@see SpanBuilderInterface::setNoParent} was called.
      *
      * Defaults to {@see Context::getCurrent} when {@see SpanBuilderInterface::startSpan} was called if not explicitly set.
      */
-    public function setParent(Context $parentContext): SpanBuilderInterface;
+    public function setParent(ContextInterface $parentContext): SpanBuilderInterface;
```
Nevay, line 25:
```suggestion
    public function setParent(ContextInterface|false|null $parentContext): SpanBuilderInterface;
```
Should this be changed (same typehint is already used in metrics API/SDK for context parameters) and `::setNoParent()` be removed?

src/Context/ImplicitContextKeyedInterface.php
```diff
@@ -26,8 +26,8 @@ interface ImplicitContextKeyedInterface
     public function activate(): ScopeInterface;
 
     /**
-     * Returns a new {@see Context} created by setting `$this` into the provided [@see Context}.
-     * It is suggested to use {@see Context::withContextValue()} instead of using this method directly.
+     * Returns a new {@see ContextInterface} created by setting `$this` into the provided [@see ContextInterface}.
+     * It is suggested to use {@see ContextInterface::withContextValue()} instead of using this method directly.
```
Nevay, line 30: What is the reason for this suggestion? Using this method feels more fluent.
```php
$spanBuilder
    ->setParent($parent)
    ->startSpan()
    ->storeInContext($parent)
    ->activate()
```

src/Context/Propagation/NullPropagator.php
```diff
@@ -5,6 +5,7 @@
 namespace OpenTelemetry\Context\Propagation;
 
 use OpenTelemetry\Context\Context;
+use OpenTelemetry\Context\ContextInterface;
 
 class NullPropagator implements TextMapPropagatorInterface
```
Nevay, line 10: Should be removed, duplicates `NoopTextMapPropagator`.

src/API/Trace/SpanBuilderInterface.php
brettmc: Yes that makes sense. Are false and null equivalent, here?

src/Context/Propagation/NullPropagator.php
brettmc: agree

src/Context/ImplicitContextKeyedInterface.php
brettmc: It's before my time, I think it's ok to remove the comment.

src/SDK/Trace/SpanBuilder.php
```diff
@@ -51,17 +53,9 @@ public function __construct(
     }
 
     /** @inheritDoc */
-    public function setParent(Context $parentContext): API\SpanBuilderInterface
+    public function setParent($context): API\SpanBuilderInterface
```
brettmc, line 56: I think we should document here (or in the API?) what ContextInterface/null/false mean

Nevay: `SpanBuilderInterface::setParent()`:
```
 * @param ContextInterface|false|null $context the parent context, null to use the
 *        current context, false to set no parent
```

brettmc (approved):
approved, some conflicts to be resolved before we can merge

PR #821, opened 2022-09-10
Use array lookup instead of recursion in `Context`

Nevay:
Prevents leaking memory if `Context::with()` is called iteratively; following example throws an out of memory error with the recursive implementation:
```php
$contextKey = Context::createKey('-');
$context = Context::getCurrent();
for ($i = 0;; $i++) {
    $context = $context->with($contextKey, $i);
}
```


src/Context/Context.php
```diff
@@ -4,111 +4,35 @@
...
-    /**
-     * Makes `$this` the currently active {@see Context}.
-     *
-     * @todo: Implement this on the API side
-     */
-    public function activate(): ScopeInterface
+    private function __construct()
     {
-        return self::storage()->attach($this);
+        self::$spanContextKey = SpanContextKey::instance();
```
Nevay, line 30: Should this use `class_exists()` or should the context key be moved to `open-telemetry/context`?

brettmc: IMO, move to open-telemetry/context if it's not too invasive.

src/Context/Context.php
```diff
@@ -19,15 +16,15 @@ final class Context
 
     // Optimization for spans to avoid copying the context array.
     private static ContextKey $spanContextKey;
-    private ?SpanInterface $span = null;
+    private ?object $span = null;
```
brettmc, line 19: `$span` is probably no longer the right name for this variable, it looks like it can be anything?

Nevay: Still the correct name, this property holds the value that is assigned to the span context key.

brettmc (approved):

bobstrecansky (approved):

PR #824, opened 2022-09-13
Rename `AbstractSpan` to `Span` for consistent access to static methods

Nevay:
Unifies access to static interface methods. Other API interfaces with static methods have a class without the `Interface` suffix that provides an implementation, e.g.:
`Baggage::getCurrent(): BaggageInterface`
`SpanContext::getInvalid(): SpanContextInterface`

I considered creating a `Span` class that does not implement `SpanInterface` but disliked the idea of splitting the PHPDoc documentation in different classes.


brettmc (changes requested):
LGTM. Can you please add a BC compatibility entry for this one, as it's likely to break a lot of user code.

brettmc (approved):

PR #828, opened 2022-09-18
Add self-diagnostic metrics to `BatchSpanProcessor`

Nevay:


src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
```diff
@@ -56,7 +67,8 @@ public function __construct(
         int $scheduledDelayMillis = self::DEFAULT_SCHEDULE_DELAY,
         int $exportTimeoutMillis = self::DEFAULT_EXPORT_TIMEOUT,
         int $maxExportBatchSize = self::DEFAULT_MAX_EXPORT_BATCH_SIZE,
-        bool $autoFlush = true
+        bool $autoFlush = true,
+        ?MeterProviderInterface $meterProvider = null
```
brettmc, line 71: I think this should be a `MeterInterface`

Nevay: I think `MeterProviderInterface` is the correct type for instrumentation - it's not the responsibility of the user to define the name / version / schema url for our internal instrumentation. ([Java uses MeterProvider too](https://github.com/open-telemetry/opentelemetry-java/blob/6209109e0adc0997d5f1bdfaabd09bc5f3711a75/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessorBuilder.java#L129), couldn't find other languages that use metrics for self-diagnostics.)

brettmc (approved):

bobstrecansky:
@Nevay - this looks ready, just has a merge conflict that needs to be remediated.

PR #831, opened 2022-09-24
Transport - move `contentType` from `::send()` argument to transport property

Nevay:
Move content type from `::send()` argument to property.

Also adds OTLP exporters that support protobuf and json format, based on the `::contentType()` of the provided transport.


src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -20,8 +20,16 @@
 
 final class GrpcTransportFactory implements TransportFactoryInterface
 {
+    /**
+     * @psalm-param "application/x-protobuf" $contentType
```
Nevay, line 24: Didn't find a solution that works with template type on `TransportFactoryInterface`.

src/Contrib/OtlpHttp/Exporter.php
```diff
@@ -8,40 +8,27 @@
...
 use OpenTelemetry\SDK\Common\Export\TransportInterface;
-use OpenTelemetry\SDK\Common\Future\CancellationInterface;
-use OpenTelemetry\SDK\Common\Future\FutureInterface;
 use OpenTelemetry\SDK\Common\Otlp\HttpEndpointResolver;
 use OpenTelemetry\SDK\Trace\SpanExporterInterface;
 use function sprintf;
-use Throwable;
 
-class Exporter implements SpanExporterInterface
+class Exporter
```
Nevay, line 21: `OtlpHttp` and `OtlpGrpc` should be removed, kept for BC for now.

Makefile
```diff
@@ -15,7 +15,7 @@ update: ## Update dependencies
 	$(DC_RUN_PHP) env XDEBUG_MODE=off composer update
 test: test-unit test-integration ## Run unit and integration tests
 test-unit: ## Run unit tests
-	$(DC_RUN_PHP) env XDEBUG_MODE=coverage vendor/bin/phpunit --testsuite unit --colors=always --coverage-text --testdox --coverage-clover coverage.clover --coverage-html=tests/coverage/html --log-junit=junit.xml
+	$(DC_RUN_PHP) env XDEBUG_MODE=coverage vendor/bin/phpunit --testsuite unit --colors=always --testdox --coverage-clover coverage.clover --coverage-html=tests/coverage/html --log-junit=junit.xml
```
Nevay, line 18: The `coverage-text` output clutters the CLI / hides the test results; `coverage-html` seems more useful overall.

src/SDK/Trace/SpanExporterInterface.php
```diff
@@ -12,16 +12,14 @@
  */
 interface SpanExporterInterface
 {
-    public static function fromConnectionString(string $endpointUrl, string $name, string $args);
```
Nevay, line 15: Should use a `SpanExporterFactoryInterface` instead to allow injecting dependencies.

src/Contrib/OtlpHttp/Exporter.php
brettmc: how about a `@deprecated` class annotation (or, even raising a deprecation notice in the constructor) to warn users?

brettmc (approved):

Makefile
bobstrecansky: I think we originally used `coverage-text` because `coverage-html` didn't give us everything we wanted at the time.  I'm sure it's better now.

src/SDK/Trace/SpanExporterInterface.php
bobstrecansky: Seems like a reasonable thing to do.

src/Contrib/OtlpHttp/Exporter.php
Nevay: We don't provide an alternative yet - will be deprecated/replaced in #838.

PR #838, opened 2022-10-07
refactor otlp exporter

brettmc:
Remove OtlpHttp and OtlpGrpc exporters, and replace with Otlp\SpanExporter and Otlp\MetricExporter using a transport to manage grpc and http/protobuf complexity.
Allow setting signal and protocol on transport factories (particularly for grpc, which uses signal to look up grpc method)
Adding BC for OtlpHttp and OtlpGrpc, which will still fail but with a more helpful message.
Fix failing examples.
Send user-agent header with otlp exports.
Allow otlp http/json


src/Contrib/OtlpHttp/OtlpHttpTransportFactory.php
```diff
@@ -0,0 +1,89 @@
...
+        array $headers = [],
+        $compression = null,
+        float $timeout = 10.,
+        int $retryDelay = 100,
+        int $maxRetries = 3,
+        ?string $cacert = null,
+        ?string $cert = null,
+        ?string $key = null
+    ): PsrTransport {
+        $endpoint ??= $this->hasEnvironmentVariable(Env::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
```
Nevay, line 45: Can we move this into `ExporterFactory::fromEnvironment()` instead?
Should be combined with the `grpc` implementation, the only differences are the default endpoint and the transport factory.

src/SDK/Trace/ExporterFactory.php
```diff
@@ -133,12 +148,42 @@ private static function getOptionFromDsn(DsnInterface $dsn, string $parameter):
...
+
+    /**
+     * @phan-suppress PhanParamTooFew
+     * @psalm-suppress TooFewArguments
+     */
+    private static function buildTransport(string $protocol): TransportInterface
+    {
+        $factoryClass = self::KNOWN_TRANSPORT_FACTORIES[self::normalizeProtocol($protocol)];
+        self::validateClass($factoryClass);
+        $factory = new $factoryClass();
```
Nevay, line 183: Users should be able to provide custom transport factory implementations.

src/SDK/Common/Environment/EnvironmentVariablesTrait.php
```diff
@@ -17,6 +17,11 @@ public function getIntFromEnvironment(string $key, int $default): int
         return Accessor::getInt($key, (string) $default);
     }
 
+    public function getFloatFromEnvironment(string $key, float $default): float
+    {
+        return Accessor::getInt($key, (string) $default);
```
Nevay, line 22:
```suggestion
        return Accessor::getFloat($key, (string) $default);
```

src/SDK/Common/Export/TransportFactoryInterface.php
```diff
@@ -29,4 +29,8 @@ public function create(
         ?string $cert = null,
         ?string $key = null
     ): TransportInterface;
+
+    public function withSignal(string $signal): self;
```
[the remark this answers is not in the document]
brettmc: I agree, this was to get around SDK->Contrib dependency violation. I can use `method_exists` to achieve the same thing.

src/SDK/Trace/ExporterFactory.php
brettmc: I think we would need to invent a new env var for that, something like `OTEL_PHP_TRANSPORT_FACTORY_CLASS`...

src/Contrib/OtlpHttp/OtlpHttpTransportFactory.php
brettmc: Yes, or I might create an otlp-specific exporter factory (which I think can also help to remove some of our sneaky deptrac violations)

src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -39,24 +50,44 @@ public function create(
         ?string $cert = null,
         ?string $key = null
     ): TransportInterface {
+        $endpoint ??= $this->hasEnvironmentVariable(Env::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT) ?
```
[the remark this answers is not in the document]
brettmc: moved into Otlp\SpanExporterFactory

src/Contrib/OtlpHttp/OtlpHttpTransportFactory.php
brettmc: moved to Otlp\SpanExporterFactory

src/SDK/Trace/ExporterFactory.php
brettmc: I'm thinking of a follow-up PR which allows registering exporter and transport factories, and using composer autoload->files to register our initial set.

src/Contrib/OtlpHttp/OtlpHttpTransportFactory.php
```diff
@@ -0,0 +1,45 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\OtlpHttp;
+
+use OpenTelemetry\Contrib\Otlp\OtlpTransportFactoryInterface;
+use OpenTelemetry\SDK\Common\Environment\EnvironmentVariablesTrait;
+use OpenTelemetry\SDK\Common\Export\Http\PsrTransport;
+use OpenTelemetry\SDK\Common\Export\Http\PsrTransportFactory;
+
+class OtlpHttpTransportFactory implements OtlpTransportFactoryInterface
```
Nevay, line 12: Should be moved to `OpenTelemetry\Contrib\Otlp`, no need for an additional package.

src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -75,11 +82,25 @@ public function create(
         return new GrpcTransport(
             $grpcEndpoint,
             $opts,
-            $method,
+            GrpcTransport::method($this->signal),
             $headers,
         );
     }
 
+    public function withSignal(string $signal): self
```
Nevay, line 90: Should be moved to `OpenTelemetry\Contrib\Otlp`.

src/Contrib/Otlp/OtlpTransportFactoryInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Contrib\Otlp;
+
+use OpenTelemetry\SDK\Common\Export\TransportFactoryInterface;
+
+interface OtlpTransportFactoryInterface extends TransportFactoryInterface
```
Nevay, line 9: IMO this should be internal to `SpanExporterFactory` / I don't think we need this interface. We could resolve the grpc endpoint similar to how we resolve the http endpoint.

src/Contrib/OtlpHttp/OtlpHttpTransportFactory.php
brettmc: done, which mostly removes the need for `OtlpHttp` (except for http discovery + virtual packages; I've just moved them into suggestions for the time being)

src/Contrib/Otlp/OtlpTransportFactoryInterface.php
brettmc: removed the interface and moved the logic to SpanExporterFactory

src/Contrib/Grpc/GrpcTransportFactory.php
brettmc: removed altogether as part of removing OtlpTransportFactoryInterface

src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -80,6 +92,15 @@ public function create(
         );
     }
 
+    public static function method(string $signal): string
```
Nevay, line 95: Should be moved to `Otlp`.

src/Contrib/Grpc/composer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "open-telemetry/exporter-otlp-grpc",
+  "name": "open-telemetry/exporter-grpc",
```
Nevay, line 2:
```suggestion
  "name": "open-telemetry/transport-grpc",
```

Should remove `ext-protobuf`, `open-telemetry/exporter-otlp-common` and `open-telemetry/gen-otlp-protobuf` dependency.

src/Contrib/Grpc/GrpcTransportFactory.php
brettmc: k, I'll also move the user-agent header over to Otlp, since that's only a requirement for otlp exporting

bobstrecansky (approved):

erdemkose:
Great improvement. Thanks for the effort. I want to try it on a small project to see it in a real-world environment. Do you plan to create a release?

brettmc:
Hi @erdemkose good idea, we'll get one out soon.

PR #839, opened 2022-10-09
Grpc metadata value must be an array

leo108:
It will throw an `InvalidArgumentException` with the message `Bad metadata value given` if the value in metadata is not an array.

The doc:
https://grpc.io/docs/what-is-grpc/core-concepts/#metadata:~:text=and%20the%20values%20are%20typically%20strings

The grpc source code:
https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/src/php/ext/grpc/call.c#L114
https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/src/php/ext/grpc/call.c#L340-L342


brettmc (approved):

brettmc:
Thanks for the fix, @leo108 .
I'm in the process of refactoring the otlp exporters, and I'll make sure this fix makes it across.

PR #844, opened 2022-10-12
resolve deptrac baseline

### LINKED ISSUE #718: Resolve Deptrac baseline
The [deptrac](https://github.com/qossmic/deptrac) dependency check reports a (supposedly) dependency violation, which has been added in the [deptrac baseline file](https://github.com/open-telemetry/opentelemetry-php/blob/main/deptrac.baseline.yaml).

This violation should be resolved (either be fixing the cause or confirming the violation error to be a bug on deptrac's side)

brettmc:
Nevay found that an @internal annotation in NonRecordingSpan was the cause of the violation, so remove it.

Closes #718

### DIFF
--- a/deptrac.baseline.yaml
+++ b/deptrac.baseline.yaml
@@ -1,4 +1,2 @@
 deptrac:
   skip_violations:
-    OpenTelemetry\SDK\Trace\Span:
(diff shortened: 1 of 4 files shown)


brettmc:
@bobstrecansky we can ignore those failures, they're fixed in other PRs

PR #845, opened 2022-10-17
Sdk builder

### LINKED ISSUE #301: Configuration Interface and default configuration

brettmc:
Fluent SDK builder mechanism to create and configure a working SDK.

Closes #301


src/SDK/SdkBuilder.php
```diff
@@ -0,0 +1,48 @@
...
+        return $this;
+    }
+
+    public function setPropagator(TextMapPropagatorInterface $propagator): self
+    {
+        $this->propagators[] = $propagator;
+
+        return $this;
+    }
+
```
brettmc, line 39: java's implementation also has a `buildAndRegisterGlobal()` method. A similar method here could forward on to `OpenTelemetry\API\Common\Instrumentation\Configurator` ?

src/SDK/Sdk.php
```diff
@@ -0,0 +1,56 @@
...
+        TextMapPropagatorInterface $propagator
+    ) {
+        $this->tracerProvider = $tracerProvider;
+        $this->meterProvider = $meterProvider;
+        $this->propagator = $propagator;
+    }
+
+    public static function isDisabled(): bool
+    {
+        return Accessor::getBool(Variables::OTEL_SDK_DISABLED);
```
tcarrio, line 34: I was initially going to suggest using the functions from `EnvironmentVariablesTrait` but those are actually instance methods not static.

I wonder if that's something that should be refactored, as there is no instance access being performed by that trait.

src/SDK/Metrics/MeterProvider.php
```diff
@@ -117,4 +117,9 @@ public function forceFlush(): bool
 
         return $success;
     }
+
+    public function builder(): MeterProviderBuilder
```
tcarrio, line 121: Should this be a static method? That's how it's accessed in the `sdk_builder.php` example.

See https://github.com/open-telemetry/opentelemetry-php/pull/845/files#diff-6abbe36ded0213fbcc94f09c6e4c6ab26180913a32a73db97533baeed94c95a1R39

tcarrio (commented):
Mostly a vested citizen in the CNCF channel excited for the OTel PHP SDK. A couple things I noticed here that I thought I would comment on :slightly_smiling_face:

src/SDK/Metrics/MeterProvider.php
brettmc: yes it should be static, thanks for picking that up :)

src/SDK/Sdk.php
brettmc: Yeah, that's why I didn't use `EnvironmentVariablesTrait` directly. I've had the same thought as you before, the methods in that trait could be static. Are you interested in working on a PR for it?

tcarrio: Sure thing! I could definitely refactor that out 👍

tcarrio: @brettmc I opened a PR to refactor this at #850.

src/SDK/Trace/TracerProviderBuilder.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function build(): TracerProviderInterface
+    {
+        $tracerProvider = new TracerProvider(
+            $this->spanProcessors,
+            $this->sampler ?? new ParentBased(new AlwaysOnSampler()),
+            $this->resource ?? ResourceInfoFactory::defaultResource(),
+        );
+        if ($this->autoShutdown) {
+            ShutdownHandler::register(fn (?CancellationInterface $cancellation = null): bool => $tracerProvider->shutdown($cancellation));
```
Nevay, line 61: This won't work, `ShutdownHandler::register()` does not keep a strong reference to `$this`.
```suggestion
            ShutdownHandler::register([$tracerProvider, 'shutdown']);
```

Also should probably be in `SdkBuilder` instead.

src/SDK/Trace/TracerProvider.php
```diff
@@ -66,7 +66,7 @@ public function getTracer(
         ?string $schemaUrl = null,
         iterable $attributes = []
     ): API\TracerInterface {
-        if ($this->tracerSharedState->hasShutdown() || SDK::isDisabled()) {
+        if ($this->tracerSharedState->hasShutdown() || Sdk::isDisabled()) {
```
Nevay, line 69: I think `Sdk::isDisabled()` should only affect `TracerProvider`s that are created from environment variables.

src/SDK/Trace/SpanProcessor/BatchSpanProcessorBuilder.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function setMeterProvider(MeterProviderInterface $meterProvider): self
+    {
+        $this->meterProvider = $meterProvider;
+
+        return $this;
+    }
+
+    public function setClock(ClockInterface $clock): self
```
Nevay, line 30: I don't think we should make the clock configurable in the builder as this is specific to the current implementation. If somebody needs this flexibility, they can use the constructor directly.

src/SDK/Trace/TracerProviderBuilder.php
```diff
@@ -0,0 +1,66 @@
...
+    private bool $autoShutdown = false;
+
+    public function addSpanProcessor(SpanProcessorInterface $spanProcessor): self
+    {
+        $this->spanProcessors[] = $spanProcessor;
+
+        return $this;
+    }
+
+    public function setResource(ResourceInfo $resource): self
```
Nevay, line 29: IMO this belongs in the `SdkBuilder`.

src/SDK/Trace/TracerProvider.php
brettmc: Fair enough. I've left the one in MeterProvider for now, but noted that it also should be removed when we can create meter providers from env vars.

src/SDK/Trace/TracerProviderBuilder.php
brettmc: This one, I'm not sure about. The SdkBuilder currently expects to be given a tracer & meter provider (with resources already provided, since it's a constructor arg).
The only way I can think of this working is if the SdkBuilder actually builds the tracer/meter providers (which is not what java does, and that was my inspiration: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdkBuilder.java )

bobstrecansky: Java is a good reference lang to be following for otel semantics.

bobstrecansky (approved):

PR #848, opened 2022-10-21
removing event dispatcher

brettmc:
discussed in SIG to remove this unused code. It was going to be infrastructure for emitting metrics from other parts of the code, but that's been done in a better way now.


Nevay (approved):
Has to remove `EventContextIntegrationTest` / #780 too.

bobstrecansky (approved):

PR #849, opened 2022-10-24
Add support for global instrumentation initializer

Nevay:


brettmc:
This looks good to me, do you want to mark it ready and we can try to integrate it with auto-instrumentation?

brettmc (approved):

PR #850, opened 2022-10-25
refactor: static access for EnvironmentVariablesTrait methods

tcarrio:
The `EnvironmentVariablesTrait` provided instance accessor methods with only static calls. This refactors those methods to be static, removing several cases of anonymous class instantiation to provide its functionality.

It's still possible to access static methods through `$this->` so this is also non-breaking to consumers.

### DIFF
--- a/src/Contrib/Otlp/ExporterTrait.php
+++ b/src/Contrib/Otlp/ExporterTrait.php
@@ -5,12 +5,10 @@ declare(strict_types=1);
 namespace OpenTelemetry\Contrib\Otlp;
 
 use OpenTelemetry\SDK\Behavior\LogsMessagesTrait;
-use OpenTelemetry\SDK\Common\Environment\EnvironmentVariablesTrait;
 use OpenTelemetry\SDK\Trace\Behavior\UsesSpanConverterTrait;
 
 trait ExporterTrait
 {
-    use EnvironmentVariablesTrait;
     use LogsMessagesTrait;
     use UsesSpanConverterTrait;
 }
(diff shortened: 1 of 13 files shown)


Nevay:
IMO we should drop the trait and use `Accessor` directly.

tcarrio:
I agree, generally the trait doesn't add anything other than passing arguments on to the `Accessor`, besides a couple cases where it's coercing types to be what's "expected". `Accessor` appears to want strings always.

> Example:
> `Accessor::getFloat` expects a string default value
> `EnvironmentVariablesTrait::getFloatFromEnvironment` expects a float for the default value

When it comes to entirely static traits though, the benefits of this type of mixin behavior are somewhat lost. A single `EnvironmentVariables` class with static methods would instead provide all of these things- unless it's consumed as part of the public API directly such that `Sdk` should have public methods for retrieving environment variables.

So options:

1. Make the trait's methods static
2. Remove `EnvironmentVariablesTrait` entirely
3. Provide a more friendly-typed `EnvironmentVariables` class with static methods

The largest impact in my opinion on this is whether anything using the trait does so not for internal private access, but **public APIs**. Namely, `Sdk` uses this trait, and it may be expected that those methods exist, which provide more friendly types (the `EnvironmentVariablesTrait` methods) and environment variable lookup behavior (the underlying `Accessor`).

brettmc:
I'm in favour of option 3 - it doesn't need to be a trait, and I appreciate the typing and self-documenting that a class such as `EnvironmentVariables` would give.
I'm not aware of any code outside of our own which might be accessing env vars, but I could imagine 3rd party `-contrib` code could do in future.

brettmc (approved):

PR #859, opened 2022-11-06
refactor environment configuration

brettmc:
Refactoring Environment access to allow configuration from multiple sources. Initially, the sources are env and php.ini, but this should allow us to add more sources in future without changing the API.


src/SDK/Common/Configuration/Configuration.php
```diff
@@ -0,0 +1,100 @@
...
+        foreach (self::resolvers() as $resolver) {
+            if ($resolver->hasVariable($name)) {
+                return $resolver;
+            }
+        }
+
+        return DefaultResolver::instance();
+    }
+
+    public static function getInt(string $key, int $default): int
```
Nevay, line 52: Do we need the `$default` parameters or could we return `null` instead?

src/SDK/Common/Configuration/Configuration.php
```diff
@@ -0,0 +1,100 @@
...
+        foreach (self::resolvers() as $resolver) {
+            if ($resolver->hasVariable($name)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public static function getResolver(string $name): Resolver
```
Nevay, line 41: Encapsulate this in a `CompositeResolver` and provide the composite resolver to `Accessor::getXxx()` to get rid of the `self::getResolver()` calls?

src/SDK/Common/Configuration/EnvironmentResolver.php
```diff
@@ -0,0 +1,30 @@
...
+namespace OpenTelemetry\SDK\Common\Configuration;
+
+/**
+ * @internal
+ */
+class EnvironmentResolver extends Resolver
+{
+    public function hasVariable(string $variableName): bool
+    {
+        return getenv($variableName) !== false || isset($_ENV[$variableName]);
```
Nevay, line 14: `$_SERVER`

src/SDK/Common/Configuration/EnvironmentResolver.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * @psalm-suppress InvalidReturnStatement
+     * @psalm-suppress InvalidReturnType
+     */
+    public function retrieveValue(string $variableName, ?string $default = ''): ?string
+    {
+        $value = getenv($variableName);
+        if ($value === false) {
+            $value = $_ENV[$variableName] ?? $default;
```
Nevay, line 25: `$value` retreived from `$_ENV` can be any type. I think we should allow resolvers to return `mixed`.

brettmc: That's true, but there is a bunch of logic in `Accessor` to parse and validate string values, and I think it makes sense to leave it there.

src/SDK/Common/Configuration/Configuration.php
brettmc: I did something similar, but I thought it looked neater if `Accessor` just referenced the `CompositeResolver` instance (which is closer to the original code)... wdyt?

src/SDK/Common/Configuration/Configuration.php
brettmc: We could return `null`, it would mean changing most of our factories to match. Providing a default has the benefit that the default is validated against known types and possible values, so seems safer to me.

src/SDK/Common/Configuration/EnvironmentResolver.php
Nevay: I think we should skip parsing if `$value` is not a string and just validate/coerce the type.
Limiting the type to `string` will become an actual issue when configuration files with support for more complex types are added to the spec.

src/SDK/Common/Configuration/Resolver.php
```diff
@@ -19,22 +23,17 @@ public static function resolveValue(string $variableName, $default = null): stri
...
-        return self::hasVariable($variableName) ? self::retrieveValue($variableName) : null;
-    }
-
-    public static function hasVariable(string $variableName): bool
-    {
-        return getenv($variableName) !== false || isset($_ENV[$variableName]);
+        return $this->hasVariable($variableName) ? $this->retrieveValue($variableName) : null;
     }
 
     public static function getDefault(string $variableName)
```
Nevay, line 39: Should be a separate resolver.

src/Contrib/Otlp/SpanExporterFactory.php
```diff
@@ -37,28 +37,28 @@ public function fromEnvironment(): SpanExporterInterface
...
-        $endpoint = EnvironmentVariables::has(Env::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
-            ? EnvironmentVariables::getString(Env::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
-            : EnvironmentVariables::getString(Env::OTEL_EXPORTER_OTLP_ENDPOINT);
+        $endpoint = Configuration::has(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
+            ? Configuration::getString(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
+            : Configuration::getString(Variables::OTEL_EXPORTER_OTLP_ENDPOINT);
         if ($protocol === Protocols::GRPC) {
             $endpoint .= OtlpUtil::method(Signals::TRACE);
         } else {
             $endpoint = HttpEndpointResolver::create()->resolveToString($endpoint, Signals::TRACE);
```
Nevay, line 51: This is broken, we must use `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` as-is, [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp).

src/SDK/Common/Configuration/Resolver/EnvironmentResolver.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @psalm-suppress InvalidReturnStatement
+     * @psalm-suppress InvalidReturnType
+     */
+    public function retrieveValue(string $variableName, ?string $default = ''): ?string
+    {
+        $value = getenv($variableName);
+        if ($value === false) {
+            $value = $_SERVER[$variableName] ?? $default;
```
Nevay, line 27: I think we should check `$_SERVER` first. For example vlucas/phpdotenv populates only `$_SERVER` and `$_ENV` by default, "[Using getenv() and putenv() is strongly discouraged due to the fact that these functions are not thread safe](https://github.com/vlucas/phpdotenv#putenv-and-getenv)".

Do we have to use [`local_only: true`](https://www.php.net/manual/en/function.getenv.php)?
> Warning If PHP is running in a SAPI such as Fast CGI, this function will always return the value of an environment variable set by the SAPI, even if [putenv()](https://www.php.net/manual/en/function.putenv.php) has been used to set a local environment variable of the same name. Use the local_only parameter to return the value of locally-set environment variables.

src/SDK/Common/Configuration/Configuration.php
```diff
@@ -46,9 +53,4 @@ public static function getRatio(string $key, float $default = null): float
     {
         return Accessor::getRatio($key, $default ? (string) $default : null);
```
Nevay, line 54: `Configuration::getRatio('not-set', 0)` should not throw.

src/SDK/Common/Configuration/Configuration.php
Nevay: I missed that we throw if the key is not set and no default is given - changing the default parameter to nullable / default null to match the other methods is sufficient then.

src/SDK/Common/Configuration/Configuration.php
Nevay: I'm fine with either solution; we should also move the `Accessor` logic into `Configuration` as they do pretty much the same now. (This is the change that I had first in mind; I opted for the `CompositeResolver` argument because it improves testability a bit.)

src/SDK/Common/Configuration/Configuration.php
morrisonlevi: I'm not sure about OTel specifically, but in the past it's been important on occasion to be able to distinguish between "not set" and "default", when there are multiple things that control something. Usually this comes up because of legacy env vars. For instance, and this is totally made up for example: `OTEL_HOST=collector.local` but then later we realize that means it can't support `https` as is, so there's a new one `OTEL_COLLECTOR_URI=https://collector.local`. For the sake of falling back to an older config, you generally want to be able to distinguish between default and unset.

brettmc: @morrisonlevi we currently manage that scenario with `::has`, although it might start to get gnarly with more variables:
```php
$endpoint = Configuration::has(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
    ? Configuration::getString(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)
    : Configuration::getString(Variables::OTEL_EXPORTER_OTLP_ENDPOINT);
```

src/SDK/Common/Configuration/Resolver/EnvironmentResolver.php
brettmc: I think if we want to remove getenv/putenv support it can be in a separate PR. I was worried about what that would mean for all the tests which rely on putenv, but I think that phpunit's `@backupGlobals` feature should do (and not having to support both $_ENV and getenv should simplify tests a little)

src/SDK/Common/Configuration/Configuration.php
brettmc: I might have fixed this in a recent commit, it's behaving as it should now and returning `0.0`; I've added a test for it.

src/SDK/Common/Configuration/EnvironmentResolver.php
brettmc: resolvers can return `mixed` now. Rather than skip parsing, I updated the parsers to allow their own type (eg bool|string)...

bobstrecansky (approved):

PR #862, opened 2022-11-16
adding exporter + transport registry

brettmc:
To remove the hidden dependency between SDK and contrib, move all of the exporter knowledge out of exporter factory, and create a registry to hold those values.
The registry can be added to by contrib modules as part of composer autoloading. This also allows users to override things like transports (eg, provide their own grpc transport).
Remove exporter's fromConnectionString as it's not in spec, and didn't fit in with the registry idea.


src/SDK/Metrics/MetricExporterFactoryInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Metrics;
+
+interface MetricExporterFactoryInterface
+{
+    public function fromEnvironment(): MetricExporterInterface;
```
brettmc, line 9: `fromEnvironment` might not be the best name, given recent Environment->Configuration changes. Maybe just `::create()` ?

brettmc:
I haven't yet tested for race conditions (eg between autloading and registering), file loading order. Also need to ensure sdk/contrib registered factories do not clobber user-defined ones.

src/SDK/Trace/SpanExporter/ConsoleSpanExporterFactory.php
```diff
@@ -0,0 +1,18 @@
...
+namespace OpenTelemetry\SDK\Trace\SpanExporter;
+
+use OpenTelemetry\SDK\Common\Export\Stream\StreamTransport;
+use OpenTelemetry\SDK\Trace\SpanExporterInterface;
+
+class ConsoleSpanExporterFactory implements SpanExporterFactoryInterface
+{
+    public function fromEnvironment(): SpanExporterInterface
+    {
+        $transport = new StreamTransport(fopen('php://stdout', 'w'), 'application/json');
```
Nevay, line 14:
```suggestion
        $transport = FactoryRegistry::transportFactory('stream')->create('php://stdout', 'application/json');
```

src/SDK/FactoryRegistry.php
```diff
@@ -0,0 +1,145 @@
...
+use RuntimeException;
+
+class FactoryRegistry
+{
+    private static array $spanExporterFactories = [];
+    private static array $transportFactories = [];
+    private static array $metricExporterFactories = [];
+
+    /**
+     * @param callable-object|class-string $factory Factory implementing TransportFactoryInterface
```
Nevay, line 19:
```suggestion
     * @param callable-object|class-string<TransportFactoryInterface> $factory Factory implementing TransportFactoryInterface
```

`callable-object` doesn't seem to match current implementation.

src/Contrib/Otlp/_register.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+\OpenTelemetry\SDK\FactoryRegistry::registerSpanExporterFactory('otlp', \OpenTelemetry\Contrib\Otlp\SpanExporterFactory::class);
+\OpenTelemetry\SDK\FactoryRegistry::registerMetricExporterFactory('otlp', \OpenTelemetry\Contrib\Otlp\MetricExporterFactory::class);
+
+\OpenTelemetry\SDK\FactoryRegistry::registerTransportFactory('http/protobuf', \OpenTelemetry\Contrib\Otlp\OtlpHttpTransportFactory::class);
+\OpenTelemetry\SDK\FactoryRegistry::registerTransportFactory('http/json', \OpenTelemetry\Contrib\Otlp\OtlpHttpTransportFactory::class);
+\OpenTelemetry\SDK\FactoryRegistry::registerTransportFactory('http/ndjson', \OpenTelemetry\Contrib\Otlp\OtlpHttpTransportFactory::class);
```
Nevay, line 9: Transport should be just `http`.

brettmc: done, I've updated `TransportFactory` to deal with this, rather than burden the various factories with stripping the content type off the protocol...

src/SDK/FactoryRegistry.php
brettmc: thanks, phan was unhappy with `callable` but I think what I really wanted was just `TransportFactoryInterface`...

bobstrecansky (approved):

PR #864, opened 2022-11-21
adding http server span context key

brettmc:
adds a context key to the API to refer to http server spans, and an example of how it can be used


pdelewski:
@brettmc It seems that it does not make sense to continue work on that?

brettmc:
closing, no longer required

PR #867, opened 2022-11-23
Remove internal tag from Context::setStorage

### LINKED ISSUE #866: Remove internal flag from Context::setStorage
The only way I found to set a context storage today is through the [`Context::setStorage`](https://github.com/open-telemetry/opentelemetry-php/blob/deeb07a890297d6975837a357d81d868418307b9/src/Context/Context.php#L41) (e.g., integrating with the Swoole storage).

But given it's internal, how are we supposed to configure it? Clarifying it on the docs or adding an example would help a lot.

marcospassos:
Fixes #866

### DIFF
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -34,8 +34,6 @@ final class Context implements ContextInterface
     }
 
     /**
-     * @internal
-     *
      * @param ContextStorageInterface&ExecutionContextAwareInterface $storage
      */
     public static function setStorage(ContextStorageInterface $storage): void


brettmc:
Hi @marcospassos I haven't worked on context in very much depth, can you elaborate on your use-case for needing custom context storage?

marcospassos:
Sure! We've been using the [Swoole context](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Context/Swoole/SwooleContextStorage.php) for some time now, and it works perfectly. But the only way to use it is by setting it through  [`Context::setStorage`](https://github.com/open-telemetry/opentelemetry-php/blob/deeb07a890297d6975837a357d81d868418307b9/src/Context/Context.php#L41) (e.g., integrating with the Swoole storage). However, this method is flagged as internal, but it doesn't make much sense since there is no other way of specifying custom contexts.

brettmc:
That makes sense, @marcospassos - I don't see any other way to do it. Can you please sign the CLA so that we can accept your work?
And, since you're using the library, we'd love your feedback on whether there's an easier way for others to use this module. For example, we could hook into composer's autoloader and have it set the default context storage automatically (maybe enabled by a config/env var) - I don't know whether that's viable with swoole or not, but we're trying to reduce/eliminate setup code for users.

marcospassos:
> Can you please sign the CLA so that we can accept your work?

Done!

brettmc (approved):

omigafu:
> Sure! We've been using the [Swoole context](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Contrib/Context/Swoole/SwooleContextStorage.php) for some time now, and it works perfectly. But the only way to use it is by setting it through [`Context::setStorage`](https://github.com/open-telemetry/opentelemetry-php/blob/deeb07a890297d6975837a357d81d868418307b9/src/Context/Context.php#L41) (e.g., integrating with the Swoole storage). However, this method is flagged as internal, but it doesn't make much sense since there is no other way of specifying custom contexts.

@marcospassos Can you provide a more detailed example?  I need a concrete demo because I don't have a deep research on this, thanks!

PR #869, opened 2022-11-25
allow 16 char b3 trace id

### LINKED ISSUE #863: Support for 16 character trace IDs

We are using Traefik in front of a PHP application and we want to connect the traces of both. Currently they are not connected, because Traefik sends a 16 char trace ID and opentelemetry-php [expects 32 chars](https://github.com/open-telemetry/opentelemetry-php/blob/0.0.16/src/API/Trace/SpanContextValidator.php#L17).

According to the B3 specs and to the Open Telemetry specs the trace ID is either 16 or 32 chars long. Therefore opentracing-php should allow both tracing lengths.

> The first position of the b3 value is the 32 or 16 lower-hex character TraceId. Unless propagating only the Sampling State, the TraceId is required. — https://github.com/openzipkin/b3-propagation#traceid-2

> TraceId is the identifier for a trace. It is worldwide unique with practically sufficient probability by being made as 16 randomly generated bytes. — https://opentelemetry.io/docs/reference/specification/overview/#spancontext

~We tried making Traefik send 32 char trace IDs, but did not find any config option for this.~

Edit: Traefik offers a [gen128bit](https://doc.traefik.io/traefik/observability/tracing/jaeger/#gen128bit) option that somehow results in 32 chars. While this works for us, requiring 32 chars is still not what the specs say.

**Additional context**

None

brettmc:
b3 spec allows trace id to be 16 or 32 char. other SIGs left-pad with zero to resolve this.

Closes #863


tests/Unit/Extension/Propagator/B3/B3PropagatorTest.php
```diff
@@ -20,11 +20,10 @@
  */
 class B3PropagatorTest extends TestCase
 {
-    private const B3_SINGLE_TRACE_ID_BASE16 = 'ff0000000000051791e0000000000041';
-    private const B3_SINGLE_SPAN_ID_BASE16 = 'ff00051791e00041';
-    private const B3_SINGLE_HEADER_SAMPLED = self::B3_SINGLE_TRACE_ID_BASE16 . '-' . self::B3_SINGLE_SPAN_ID_BASE16 . '-1';
-    private const B3_MULTI_TRACE_ID_BASE16 = 'ff000000000000301710000000000041';
-    private const B3_MULTI_SPAN_ID_BASE16 = 'ff00003017100041';
+    private const B3_TRACE_ID_BASE8 = 'ff00051791e00041';
```
Nevay, line 23: Confusing/incorrect name, `BASE16` was likely referring to hex encoding (otherwise `B3_SPAN_ID_BASE16` should be renamed).

Nevay (approved):

bobstrecansky (approved):

PR #872, opened 2022-11-27
decode OTEL_RESOURCE_ATTRIBUTES values

### LINKED ISSUE #786: Decode values from `OTEL_RESOURCE_ATTRIBUTES`
While using the Go SDK I noticed an inconsistency in how `OTEL_RESOURCE_ATTRIBUTES` baggage values are encoded and decoded. This was due to a difference in the version of the W3C Baggage spec used for [baggage propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#propagation) and the [`OTEL_RESOURCE_ATTRIBUTES` decoding](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable).

https://github.com/open-telemetry/opentelemetry-specification/pull/2670 updates the W3C Baggage spec version used by the Resource spec to require percent-decoding values from  `OTEL_RESOURCE_ATTRIBUTES`.

brettmc:
per spec, resource attributes coming from environment should be decoded in the same way that they would be if coming from a header

Closes #786

### DIFF
--- a/tests/Unit/SDK/Resource/Detectors/EnvironmentTest.php
+++ b/tests/Unit/SDK/Resource/Detectors/EnvironmentTest.php
@@ -16,6 +16,13 @@ class EnvironmentTest extends TestCase
 {
     use EnvironmentVariables;
 
+    private Detectors\Environment $detector;
+
+    public function setUp(): void
+    {
+        $this->detector = new Detectors\Environment();
+    }
+
     public function tearDown(): void
     {
         $this->restoreEnvironmentVariables();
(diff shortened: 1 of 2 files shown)


tests/Unit/SDK/Resource/Detectors/EnvironmentTest.php
```diff
@@ -35,19 +41,38 @@ public function test_environment_get_resource_with_resource_attributes(): void
...
         $this->assertSame('value_foo', $resource->getAttributes()->get('key_foo'));
         $this->assertSame('value_bar', $resource->getAttributes()->get('key_bar'));
     }
 
+    /**
+     * @dataProvider encodedResourceValueProvider
+     */
+    public function test_environment_get_resource_with_encoded_value(string $value, string $expected): void
+    {
+        $key = '_key';
```
pdelewski, line 55: just curious, any reason of using underscore prefix?

pdelewski (approved):

tests/Unit/SDK/Resource/Detectors/EnvironmentTest.php
brettmc: no good reason. I've removed it to avoid any confusion.

bobstrecansky (approved):

PR #875, opened 2022-11-29
update contrib autoloading and gitsplit

brettmc:
* match grpc's gitsplit destination repo to match package name: 'transport-grpc'
* set autoload->files for contrib modules which support factory registry


brettmc:
note to self: add transport-grpc to packagist after merge

bobstrecansky (approved):

Issue #878, opened 2022-12-01
contrib psr18 auto-instrumentation also instruments exporter


brettmc:
Whilst doing some initial testing of the new psr-18 auto-instrumentation, with PHP 8.1 + otel dev-main, using any exporter which uses `PsrTransport` (everything except grpc), I noticed that it also creates spans/traces for the exporting call.

If using batch processor, that's one extra trace exported (because it happens as part of shutdown, when there is no active context).
If using simple processor, it's pathological: each span exported creates a new http request which is also exported and traced. I didn't let it continue (one trace contained 27k spans, so it would probably only stop on resource exhaustion).

First trace is with simple span exporter (request aborted after some time), second two are with batch span exporter:
![psr18-simple-vs-batch](https://user-images.githubusercontent.com/4978962/204961623-9a99368d-b8f5-4801-93cf-dad3e98cb6ea.png)

Detail of simple span exporter's trace:
![psr18-simple-detailed](https://user-images.githubusercontent.com/4978962/204962125-923d565b-f84a-4f57-9dd9-e97f551367a1.png)

Proposed solution:
Create a generic way to be able to identify psr-7 requests which are exporting. My best idea so far is to add a known header such as `X-Is-Otel-Exporter` from `PsrTransport`. HTTP auto-instrumentations such as psr-18 can check for this header, and ignore it (and drop the header before it is sent), essentially a pseudo-attribute for psr-7 requests.

pdelewski:
I don't see better solution for now. Is dropping this header necessary?

brettmc:
> I don't see better solution for now. Is dropping this header necessary or it's just optimization?

It's probably not necessary to drop it, I might ask on the maintainers list whether this has come up in other SIGs. Dropping does seem unnecessary, since if there wasn't an auto-instrumentation to check/drop, it would be sent to the the collector/other endpoint.

Nevay:
IMO we should run span processor flush and metric reader collect with the `Context` that was active during creation ([ref](https://github.com/open-telemetry/opentelemetry-php/pull/822#pullrequestreview-1111531210)).
The `Globals` initializer already activates noop implementations before invoking the initializers to support this / to disable auto-instrumentation of SDK requests etc. by default.

PR #879, opened 2022-12-01
adding an identification header to export requests

brettmc:
allow other parts of our code (eg auto-instrumentation) to identify whether an http request
is self-generated. use-case: do not auto-instrument our own code.

### DIFF
--- a/src/API/Common/Export/Headers.php
+++ b/src/API/Common/Export/Headers.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Export;
+
+class Headers
+{
+    const EXPORTER_HEADER = 'X-Is-Otel-Exporter';
+}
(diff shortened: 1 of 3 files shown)


src/API/Common/Export/Headers.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Export;
+
+class Headers
+{
+    const EXPORTER_HEADER = 'X-Is-Otel-Exporter';
```
brettmc, line 9: I chose a new class in `API` rather than adding this to `PsrTransport` so that auto-instrumentations did not need to depend on `SDK`...

pdelewski:
@brettmc As I commented on https://github.com/open-telemetry/opentelemetry-php/issues/878, I don't see better solution for now

src/API/Common/Export/Headers.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Export;
+
+class Headers
```
pdelewski, line 7: It would be good to add description to this class, at least its purpose to avoid confusions

brettmc:
closing, replaced by #880

PR #880, opened 2022-12-01
Run exports with the `Context` the `SpanProcessor` was created in

### LINKED ISSUE #878: contrib psr18 auto-instrumentation also instruments exporter

Nevay:
Resolves #878.

(Matches the behavior of async implementations that create the export fiber in the constructor, [example](https://github.com/Nevay/opentelemetry-php-async/blob/4e55a7d835af42636330689b4e90e53e55a94346/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php#L129)).

### DIFF
--- a/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
+++ b/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php
@@ -9,6 +9,7 @@ use function count;
 use InvalidArgumentException;
 use OpenTelemetry\API\Metrics\MeterProviderInterface;
 use OpenTelemetry\API\Metrics\ObserverInterface;
+use OpenTelemetry\Context\Context;
 use OpenTelemetry\Context\ContextInterface;
 use OpenTelemetry\SDK\Behavior\LogsMessagesTrait;
 use OpenTelemetry\SDK\Common\Future\CancellationInterface;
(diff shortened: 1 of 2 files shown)


brettmc:
I tested this out, it works for me - exporter requests are no longer instrumented. I can see that the instrumentation handler `pre` function now gets a no-op tracer when handling exporting requests.

bobstrecansky (approved):

PR #881, opened 2022-12-03
allow setting service name from auto instrumentation context

pdelewski:
Currently there is no way to set service name from auto instrumentation context for zipkin and newrelic exporters.
We should read that from `OTEL_SERVICE_NAME` or any other configuration settings IMO.

### DIFF
--- a/src/Contrib/Newrelic/Exporter.php
+++ b/src/Contrib/Newrelic/Exporter.php
@@ -29,19 +29,16 @@ class Exporter implements SpanExporterInterface
     public const DATA_FORMAT_VERSION_DEFAULT = '1';
 
     private TransportInterface $transport;
-    private string $name;
     private string $endpointUrl;
 
     public function __construct(
-        $name,
         TransportInterface $transport,
         string $endpointUrl,
         SpanConverter $spanConverter = null
     ) {
-        $this->name = $name;
         $this->endpointUrl = $endpointUrl;
         $this->transport = $transport;
-        $this->setSpanConverter($spanConverter ?? new SpanConverter($name));
+        $this->setSpanConverter($spanConverter ?? new SpanConverter());
     }
 
     /**
@@ -54,7 +51,7 @@ class Exporter implements SpanExporterInterface
 
     private function convert(iterable $spans): array
     {
-        $commonAttributes = ['attributes' => [ 'service.name' => $this->name,
+        $commonAttributes = ['attributes' => [
             'host' => $this->endpointUrl, ]];
 
         return [[ 'common' => $commonAttributes,
--- a/src/Contrib/Newrelic/SpanConverter.php
+++ b/src/Contrib/Newrelic/SpanConverter.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace OpenTelemetry\Contrib\Newrelic;
 
 use OpenTelemetry\SDK\Common\Time\Util as TimeUtil;
+use OpenTelemetry\SDK\Resource\ResourceInfoFactory;
 use OpenTelemetry\SDK\Trace\SpanConverterInterface;
 use OpenTelemetry\SDK\Trace\SpanDataInterface;
+use OpenTelemetry\SemConv\ResourceAttributes;
 
 /**
  * @see https://docs.newrelic.com/docs/distributed-tracing/trace-api/report-new-relic-format-traces-trace-api/#new-relic-guidelines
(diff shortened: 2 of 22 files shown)


Nevay:
IMO the service name should not be a constructor argument and should be read in `::export()` from `SpanDataInterface::getResource()` instead.

pdelewski:
> IMO the service name should not be a constructor argument and should be read in `::export()` from `SpanDataInterface::getResource()` instead.

Thx. That's the discussion I wanted to start opening this PR

brettmc:
I think the spec agrees with Nevay here; the objective of providing the name in the exporter is eventually to pass it through as a service name, but that should indeed come from a resource attribute: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service
Since not all exporters accept a name, it looks like we should remove the constructor argument totally (assuming we can instead get the service name out of resource attributes)

pdelewski:
> IMO the service name should not be a constructor argument and should be read in `::export()` from `SpanDataInterface::getResource()` instead.

I'm currently looking into this code and have some concerns. `export` takes whole batch

Nevay:
Yes, https://github.com/open-telemetry/opentelemetry-php/blob/587bc5aaf0b1c341391ac5843fa68e926852fe7a/src/Contrib/Zipkin/SpanConverter.php#L85 should be changed to something like
```php
'serviceName' => $span->getResource()->getAttributes()->get(ResourceAttributes::SERVICE_NAME),
```

pdelewski:
> Yes,
> 
> https://github.com/open-telemetry/opentelemetry-php/blob/587bc5aaf0b1c341391ac5843fa68e926852fe7a/src/Contrib/Zipkin/SpanConverter.php#L85
> 
> should be changed to something like
> ```
> 'serviceName' => $span->getResource()->getAttributes()->get(ResourceAttributes::SERVICE_NAME),
> ```

Thx. That's exactly what I was thinking about

src/Contrib/Zipkin/SpanConverter.php
```diff
@@ -78,11 +72,12 @@ private function convertSpan(SpanDataInterface $span): array
         $startTimestamp = TimeUtil::nanosToMicros($span->getStartEpochNanos());
         $endTimestamp = TimeUtil::nanosToMicros($span->getEndEpochNanos());
 
+        $serviceName =  $span->getResource()->getAttributes()->get(ResourceAttributes::SERVICE_NAME) ?? 'zipkin';
```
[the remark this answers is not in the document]
pdelewski: Yes, that's right, it's still work in progress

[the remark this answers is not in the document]
pdelewski: @Nevay fixed fallback

src/Contrib/Newrelic/Exporter.php
```diff
@@ -23,25 +23,23 @@
  */
 class Exporter implements SpanExporterInterface
 {
+    public $name;
```
Nevay, line 26: Should be removed.

src/Contrib/Newrelic/SpanConverter.php
```diff
@@ -40,12 +35,16 @@ private function convertSpan(SpanDataInterface $span): array
         $startTimestamp = TimeUtil::nanosToMillis($span->getStartEpochNanos());
         $endTimestamp = TimeUtil::nanosToMillis($span->getEndEpochNanos());
 
+        $serviceName =  $span->getResource()->getAttributes()->get(ResourceAttributes::SERVICE_NAME)
+                        ??
+                        ResourceInfoFactory::defaultResource()->getAttributes()->get(ResourceAttributes::SERVICE_NAME);
```
brettmc, line 40: depending on how many detectors are enabled (by default, all of them are), this is a bit of overhead we could avoid? Perhaps setting a default/fallback service name in the constructor?

pdelewski: Good point

src/Contrib/Newrelic/SpanConverter.php
```diff
@@ -18,9 +20,9 @@ class SpanConverter implements SpanConverterInterface
 
     private string $serviceName;
```
brettmc, line 21: a better name might be `fallbackServiceName` or `defaultServiceName` ?

tests/Unit/Contrib/Zipkin/ZipkinSpanConverterTest.php
```diff
@@ -61,14 +61,14 @@ public function test_should_convert_a_span_to_a_payload_for_zipkin(): void
...
-        $converter = new SpanConverter('test.name');
+        $converter = new SpanConverter();
         $row = $converter->convert([$span])[0];
 
         $this->assertSame($span->getContext()->getSpanId(), $row['id']);
         $this->assertSame($span->getContext()->getTraceId(), $row['traceId']);
         $this->assertSame('1000000000000000', $row['parentId']);
 
-        $this->assertSame('test.name', $row['localEndpoint']['serviceName']);
+        $this->assertSame('unknown_service', $row['localEndpoint']['serviceName']);
```
brettmc, line 71: you might need to merge in main branch, this is something I changed in a commit in the last day or so... `unknown_service:php`

pdelewski: yes, that's right. Will do that when this PR will be approved

brettmc (approved):

PR #884, opened 2022-12-06
span status compliance

brettmc:
implement spec-compliance for span status:
- Ok is final
- Unset should be ignored
- only Error may have a desription
- Error can be updated to Ok

ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status

### DIFF
--- a/src/SDK/Trace/Span.php
+++ b/src/SDK/Trace/Span.php
@@ -256,7 +256,18 @@ final class Span extends API\Span implements ReadWriteSpanInterface
             return $this;
         }
 
-        $this->status = StatusData::create($code, $description);
+        // An attempt to set value Unset SHOULD be ignored.
+        if ($code === API\StatusCode::STATUS_UNSET) {
+            return $this;
+        }
+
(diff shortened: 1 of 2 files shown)


src/SDK/Trace/Span.php
```diff
@@ -256,7 +256,18 @@ public function setStatus(string $code, string $description = null): self
...
+            return $this;
+        }
+
+        // When span status is set to Ok it SHOULD be considered final and any further attempts to change it SHOULD be ignored.
+        if ($this->status->getCode() === API\StatusCode::STATUS_OK) {
+            return $this;
+        }
+
+        // Description MUST be IGNORED for StatusCode Ok & Unset values.
+        $this->status = StatusData::create($code, ($code === API\StatusCode::STATUS_ERROR) ? $description : null);
```
Nevay, line 270: Already handled in `StatusData::create()`.

Nevay (approved):

bobstrecansky (approved):

PR #885, opened 2022-12-06
create text map propagators from registry

brettmc:
removing undocumented dependency between sdk and extension by adding text map propagators to a registry as part of composer autoloading, similar to how exporters, transports etc are treated.


src/Extension/Propagator/B3/_register.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+\OpenTelemetry\SDK\FactoryRegistry::registerTextMapPropagatorFactory(
+    \OpenTelemetry\SDK\Common\Configuration\KnownValues::VALUE_B3,
+    \OpenTelemetry\Extension\Propagator\B3\B3SinglePropagatorFactory::class
```
Nevay, line 7: I would prefer if we add a `ClosureTextMapPropagatorFactory` (or allow registering closure factories directly) to avoid adding a factory class for every implementation.

brettmc: I've just removed the factories altogether, and registered the propagator instances instead. It's simpler, but also adding non-factories to a `FactoryRegistry` which is a little shady.

pdelewski: Agree, that it's simpler, but as you written it breaks rule of storing only factories there. I don't have strong opinions for or against. Maybe, we should consider some renaming as `FactoryRegistry` will no longer store factories only.

brettmc: I've renamed it to `Registry`, which isn't very imaginative but seems valid.

pdelewski (approved):

PR #887, opened 2022-12-06
fixing default service name

brettmc:
default service name should contain runtime name if available: java just appends ":java" so I did the same.
ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service

### DIFF
--- a/src/SDK/Resource/Detectors/SdkProvided.php
+++ b/src/SDK/Resource/Detectors/SdkProvided.php
@@ -10,7 +10,7 @@ use OpenTelemetry\SDK\Resource\ResourceInfo;
 use OpenTelemetry\SemConv\ResourceAttributes;
 
 /**
- * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service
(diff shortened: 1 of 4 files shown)


src/SDK/Resource/Detectors/SdkProvided.php
```diff
@@ -10,14 +10,14 @@
 use OpenTelemetry\SemConv\ResourceAttributes;
 
 /**
- * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service
```
Nevay, line 13:
```suggestion
 * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value
```

I think we should continue linking to the resource attributes that are provided by this detector.

pdelewski:
> default service name should contain runtime name if available: java just appends ":java" so I did the same.

Interesting, I was thinking that this requires something more sophisticated :)

pdelewski (approved):

PR #888, opened 2022-12-06
adding baggage setValue and removeValue, plus example usage

brettmc:
setValue and removeValue were added to spec after our initial baggage implementation, so implement the missing features.
ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#set-value
ref: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#remove-value


Nevay:
I don't think we need this - functionality is already available via `::toBuilder()`.

bobstrecansky (approved):

PR #890, opened 2022-12-07
removing jaeger exporters

brettmc:
Jaeger has dropped support for the jaeger exporter in favour of opentelemetry exporter. The upstream repos are now archived. Support has been dropped from the opentelemetry collector, and other SIGs should follow soon.
ref: https://github.com/open-telemetry/opentelemetry-specification/pull/2858 and https://github.com/open-telemetry/opentelemetry-specification/issues/2859


pdelewski:
> Jaeger has dropped support for the jaeger exporter in favour of opentelemetry exporter. The upstream repos are now archived. Support has been dropped from the opentelemetry collector, and other SIGs should follow soon. ref: [open-telemetry/opentelemetry-specification#2858](https://github.com/open-telemetry/opentelemetry-specification/pull/2858) and [open-telemetry/opentelemetry-specification#2859](https://github.com/open-telemetry/opentelemetry-specification/issues/2859)

@brettmc Is not that too early? It seems that there will be deprecation period regarding removing it from spec

pdelewski:
> > Is not that too early? It seems that there will be deprecation period regarding removing it from spec
> 
> I missed SIG yesterday, but Bob noted that he thought better to remove it before beta. It's definitely going away (upstream are not even accepting security fixes anymore), so removing it earlier means less for us to maintain. A deprecation makes more sense for GA libraries where people may be using it in production and need some time to migrate, but I think doesn't apply to us?

Bob mentioned that yesterday, but after reading attached tickets, I was not sure we should do that now, however I agree that it should not apply to us

Nevay (approved):

bobstrecansky (approved):

PR #891, opened 2022-12-08
store logger in context, move LogsMessagesTrait into API

### LINKED ISSUE #889: make logging available from API
There are a few outstanding todo's in the API layer to log warnings.

Currently, we cannot log errors/warnings from API as the logger lives in the SDK.

Options:

- move `LoggerHolder` and `LogsMessagesTrait` from SDK to API as-is
- make logger available from `Globals`, stored in context and marked `@internal` (may be initialised too late to be useful for startup messages?)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#configuring-error-handlers makes is sound like configuring error logging should be an SDK thing, but https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#self-diagnostics seems clear that all layers should be able to log errors.

brettmc:
remove global/static LoggerHolder, and store logger in Context via Globals. LoggerHolder is deprecated, but a temporary BC hack should avoid immediate breakage.
Move LogsMessagesTrait into API, since there is a requirement (and some outstanding todo's) in the API layer to emit warnings in certain situations.

Closes #889


brettmc:
> I don't think we should make a PSR logger accessible via Globals, considering that the spec contains an [experimental logging backend](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md) (also the reason why I dropped Globals::logger() from https://github.com/open-telemetry/opentelemetry-php/pull/822). IMO we should instead switch to using an otel Logger for our self-diagnostics once implemented.

That makes sense, I'll close this one and do a simple migration from SDK to API.

PR #902, opened 2022-12-18
allow OTEL_SDK_DISABLED to disable sdk autoloading

brettmc:


src/SDK/SdkAutoloader.php
```diff
@@ -24,7 +24,7 @@ class SdkAutoloader
     public static function autoload(): bool
     {
         try {
-            self::$enabled ??= Configuration::getBoolean(Variables::OTEL_PHP_AUTOLOAD_ENABLED);
+            self::$enabled ??= Configuration::getBoolean(Variables::OTEL_PHP_AUTOLOAD_ENABLED) && !Sdk::isDisabled();
```
Nevay, line 27: This should be moved into the initializer callback - currently it's checked during autoloader initialization which makes usage with `.env` files that are loaded afterwards impossible.

Nevay (changes requested):
Must not disable propagators:
> This setting has no effect on propagators configured through the OTEL_PROPAGATORS variable.

bobstrecansky (approved):

PR #903, opened 2022-12-20
Incorrect metrics endpoint path

jlabedo:
The path is correctly appended in SpanExporterFactory, but not in MetricExporterFactory

### DIFF
--- a/src/Contrib/Otlp/MetricExporterFactory.php
+++ b/src/Contrib/Otlp/MetricExporterFactory.php
@@ -55,7 +55,7 @@ class MetricExporterFactory implements MetricExporterFactoryInterface
             case KnownValues::VALUE_HTTP_PROTOBUF:
             case KnownValues::VALUE_HTTP_JSON:
                 return $factory->create(
-                    $endpoint,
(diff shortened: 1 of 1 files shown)


brettmc (changes requested):
`OtlpUtil::method()` is only for grpc, but you're correct that a value should be appended to the provided endpoint, sometimes.

See what you can make of [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) but I think what's clear is that `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` should be used as-is for http protocols, but `OTEL_EXPORTER_OTLP_ENDPOINT` should have a path appended.
For gRPC, we append a path for either variable.
Would appending `/v1/metrics` to line 40 be enough?

bobstrecansky:
@jlabedo - are you still planning on continuing with this work?

bobstrecansky:
@jlabedo - we are happy to pick up this bug squash if you don't have bandwidth to complete this PR.

brettmc:
Closing, resolved in another PR

PR #909, opened 2023-01-11
adding a configuration for disabling auto-instrumentations

brettmc:
addressing review feedback from pdelewski, provide a mechanism that can be used to disable auto-instrumentation packages without actually uninstalling them.
A new variable, OTEL_PHP_DISABLED_INSTRUMENTATIONS, has been added which accepts a list of instrumentation names to be disabled. Each instrumentation can call Instrumentation::isDisabled to determine if they should bail without registering hooks.


src/SDK/Common/Instrumentation.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Common;
+
+use OpenTelemetry\API\Behavior\LogsMessagesTrait;
+use OpenTelemetry\SDK\Common\Configuration\Configuration;
+use OpenTelemetry\SDK\Common\Configuration\Variables;
+
+class Instrumentation
```
pdelewski, line 11: I'm thinking whether we need new class for it or if name of it is correct. It of course depends what is planned here in the future, but maybe for now we could just have a function that would be part of configuration, something like
`isInstrumentationDisabled` ?

brettmc: I'll move it into `Sdk` class, where a similar method already exists.

pdelewski (approved):

PR #918, opened 2023-01-31
relax composer version requirements

### LINKED ISSUE #917: self.version in composer.json incompatible with subtree splits
**Describe your environment**
Since we decided to version all of our packages independently, it's quite likely (and desired) that we have different packages at different tags.
However, many of our packages are incompatible with each other due to composer requirements using `self.version` which were originally set up when we deployed the whole repository via one release on the monorepo.

**Steps to reproduce**
Require open-telemetry/sdk 1.0.0beta2 and open-telemetry/exporter-otlp 1.0.0beta1 specifically.

Packages installed via composer

```
$ composer update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires open-telemetry/exporter-otlp 1.0.0beta1 -> satisfiable by open-telemetry/exporter-otlp[1.0.0beta1].
    - open-telemetry/exporter-otlp 1.0.0beta1 requires open-telemetry/sdk 1.0.0beta1 -> found open-telemetry/sdk[1.0.0beta1] but it conflicts with your root composer.json require (1.0.0beta2).
```

brettmc:
using self.version does not work when we have different components being released individually. ^1.0 seems like a sensible initial setting now that we have reached beta.
Fixes #917


composer.json
```diff
@@ -34,12 +34,12 @@
...
-        "open-telemetry/gen-otlp-protobuf": "self.version",
-        "open-telemetry/sdk": "self.version",
-        "open-telemetry/sdk-contrib": "self.version",
-        "open-telemetry/sem-conv": "self.version"
+        "open-telemetry/api": "^1.0",
+        "open-telemetry/context": "^1.0",
+        "open-telemetry/gen-otlp-protobuf": "^1.0",
+        "open-telemetry/sdk": "^1.0",
+        "open-telemetry/sdk-contrib": "^1.0",
+        "open-telemetry/sem-conv": "^1.0"
```
Nevay, line 42: Should probably be exact versions, we don't want `open-telemetry/opentelemetry` to replace future versions of the packages.

brettmc: Yes, I think you're right. We need to work out what `open-telemetry/opentelemetry` is. Since we no longer tag&split the monorepo to a bunch of packages at the same version, I suspect the meta-package only currently works with `dev-main`.
Setting up conflicts in composer between `open-telemetry/opentelemetry` and the other core packages might help this.

brettmc: I set these back to `self.version`. Also, apparently `replace` creates an implicit `conflicts` on the same name, so is not necessary.

PR #922, opened 2023-02-13
fixing phpunit 9.6 deprecations

### LINKED ISSUE #921: phpunit 9.6/10 deprecations
PHPUnit v9.6 introduces a bunch of [deprecations](https://github.com/sebastianbergmann/phpunit/blob/main/ChangeLog-9.6.md#960---2023-02-03) for things which will break in v10.

brettmc:
phpunit v9.6 introduces a bunch of deprecations for things that will break in v10. This fixes the tests that are currently complaining.
Since PHPUnit v10 requires php 8.1+, it will be some time before we can upgrade

Closes #921

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -376,7 +376,6 @@ return [
         'vendor/php-http',
         'vendor/phan/phan/src/Phan',
         'vendor/phpunit/phpunit/src',
-        'vendor/promphp/prometheus_client_php/src',
         'vendor/google/protobuf/src',
     ],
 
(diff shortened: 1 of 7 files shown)


brettmc:
It's worth noting that the "fix" I've applied here w.r.t `trigger_error` should fail in phpunit v10: in PHPUnit 10, `trigger_error` is [not converted to an exception](https://phpunit.de/announcements/phpunit-10.html) and so we can't make assertions on that behaviour.
However, in v9 they _are_ still converted to exceptions, but the behaviour is deprecated. So, the only workable solution I can see for now is to use `PHPUnit\Framework\Exception` which is marked as `internal` but the only class that doesn't trigger deprecation notices.

bobstrecansky (approved):

pdelewski (approved):

PR #923, opened 2023-02-13
remove older BC implementations

brettmc:
These BC features have been in place for multiple releases, it's time for them to retire.


bobstrecansky (approved):
🪓

PR #933, opened 2023-03-02
adding container detector

### LINKED ISSUE #931: Resource Detection for container properties
Detecting the id of the container that holds the service instrumented with otel allows end-users to correlate container issues with service issues.

A resource detector that can be added to the SDK and then does the reading of container.id for me automatically.

Injecting the container id via an environment variable is always possible.

**Additional context**

Similar implementations exist for Go, JS, Java, .NET and the logic can be reused for a PHP implementation as well:

* https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/detectors/node/opentelemetry-resource-detector-container/src/detectors/ContainerDetector.ts
* https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Extensions.Docker/Resources/DockerResourceDetector.cs
* https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/container.go
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/ContainerResourceProvider.java

brettmc:
Detect container id from cgroup or mountinfo, based on how otel-js does it.

Closes #931

### DIFF
--- a/src/SDK/Common/Configuration/KnownValues.php
+++ b/src/SDK/Common/Configuration/KnownValues.php
@@ -165,6 +165,7 @@ interface KnownValues
     public const VALUE_DETECTORS_PROCESS_RUNTIME = 'process_runtime';
     public const VALUE_DETECTORS_SDK = 'sdk';
     public const VALUE_DETECTORS_SDK_PROVIDED = 'sdk_provided';
+    public const VALUE_DETECTORS_CONTAINER = 'container';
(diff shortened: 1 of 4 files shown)


pdelewski:
Looks good to me.

pdelewski (approved):

PR #934, opened 2023-03-08
adding logs signal

brettmc:
This is an initial implementation of the Logs signal. I think it works and is spec-compliant. I've included examples of it working, and an example monolog integration.
Still outstanding:
- [x] write lots more tests
- [x] documentation
- [x] fix up annotated `todo`s
- [x] improve LogRecord/LogRecordData
- [x] create psr3->otel severity mapping class


src/API/Logs/LoggerInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Logs;
+
+interface LoggerInterface
```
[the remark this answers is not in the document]
brettmc: Hello, welcome, and thanks for taking the time to review!
This interface relates of this part of the otel spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logger
I don't _think_ that there's anything in the spec that disallows your suggestion, and it could make it more useful... Let's see if anybody else has thoughts on this. (I'll be away for the next couple of days, so might not respond for a bit...)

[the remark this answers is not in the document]
bobstrecansky: I think being compliant with PSR-3 would help a lot with adoption, I'm 👍🏼 on it.

examples/logs/features/monolog.php
```diff
@@ -0,0 +1,80 @@
...
+                return 19;
+            case LogLevel::EMERGENCY:
+                return 21;
+            default:
+                return 0;
+        }
+    }
+};
+
+$monolog = new Logger('otel-php', [$otelHandler, $streamHandler]);
```
bobstrecansky, line 76: **nitpick (non-blocking): ** should we use a different name here like `otel-php-monolog` to be a little more explicit?

src/API/Common/Instrumentation/Configurator.php
```diff
@@ -93,6 +100,13 @@ public function withPropagator(?TextMapPropagatorInterface $propagator): Configu
         $self = clone $this;
         $self->propagator = $propagator;
 
+        return $self;
+    }
+    public function withLoggerProvider(?LoggerProviderInterface $loggerProvider): Configurator
+    {
+        $self = clone $this;
```
bobstrecansky, line 107: **note: ** TIL about `clone` in PHP.  That's rad.

src/API/Logs/LogRecord.php
```diff
@@ -0,0 +1,94 @@
...
+        $this->attributes[$name] = $value;
+
+        return $this;
+    }
+
+    public function toLogRecordData(): LogRecordData
+    {
+        $spanContext = Span::fromContext($this->context ?? Context::getCurrent())->getContext();
+
+        //@todo improve how this is stored
```
bobstrecansky, line 77: **question:** how would you like to improve how this is stored?

src/API/Logs/LogRecordBuilder.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Logs;
+
+class LogRecordBuilder
+{
+    /**
+     * @todo is this needed, or just use LogRecord directly?
```
bobstrecansky, line 10: **comment: ** might not be.  Guess we can leave it for now until we decide we don't need it.

src/SDK/Logs/LoggerProviderInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Logs;
+
+use OpenTelemetry\API\Logs as API;
+
+interface LoggerProviderInterface extends API\LoggerProviderInterface
+{
```
bobstrecansky, line 10: **question:** Are there other functions you'd like to add besides `shutdown()` and `forceFlush()` here?

bobstrecansky (commented):
Added some initial comments.  Great start!

files/collector/otel-collector-config.yml
```diff
@@ -9,7 +9,7 @@ exporters:
   zipkin:
     endpoint: "http://zipkin:9411/api/v2/spans"
   logging:
-    loglevel: debug
+    verbosity: detailed
```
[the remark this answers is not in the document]
brettmc: that's deprecated from at least v0.72 of the collector:
```
warn	loggingexporter@v0.72.0/factory.go:109	'loglevel' option is deprecated in favor of 'verbosity'. Set 'verbosity' to equivalent value to preserve behavior.	{"kind": "exporter", "data_type": "logs", "name": "logging", "loglevel": "debug", "equivalent verbosity level": "Detailed"}
```
Although the message above hints that using "debug" should work, it in fact doesn't, so I used `detailed` which I pulled from here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md

src/API/Logs/LogRecord.php
brettmc: Just storing the return value in an array seems a little loose and php4-ish, so I guess `LogRecordData` is a typed class, and we have a method to create one from a `LogRecord`...

src/SDK/Logs/LoggerProviderInterface.php
brettmc: no, I think those are the only interface methods that the SDK adds... note that this interface extends the API interface, which is also how we do it for tracing.

src/API/Logs/LoggerInterface.php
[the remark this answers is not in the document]
brettmc: I'm happy with that, it's a very good suggestion. I will add psr-3 soon.

src/SDK/Logs/Processor/BatchLogsProcessor.php
```diff
@@ -0,0 +1,273 @@
...
+        if ($this->running) {
+            return false;
+        }
+
+        $success = true;
+        $exception = null;
+        $this->running = true;
+
+        try {
+            for (;;) {
```
lahaxearnaud, line 205: nitpick (non-blocking): Is there any other way to flush better than this infinite loop ?

Did you do that in case of messages comes during the flush ?

examples/logs/features/monolog.php
brettmc: updated

src/SDK/Logs/Processor/BatchLogsProcessor.php
brettmc: I believe that's the reason - this batch exporter is mostly copied from the traces batch exporter, which I don't completely understand...

src/API/Logs/LoggerInterface.php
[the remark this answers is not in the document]
brettmc: added psr-3 compliance

src/SDK/Logs/ReadableLogRecord.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace OpenTelemetry\SDK\Logs;
+
+use OpenTelemetry\API\Logs\LogRecordData;
+use OpenTelemetry\Context\ContextInterface;
+use OpenTelemetry\SDK\Common\Attribute\AttributesInterface;
+use OpenTelemetry\SDK\Common\Instrumentation\InstrumentationScopeInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+
+class ReadableLogRecord
```
pdelewski, line 13: I'm wondering, from engineering point of view, why this class is not derived from `Api::LogRecord`? We could prohibit write operations by throwing exceptions and we could get rid of `LogRecordData` type which is not mentioned in the spec and seems artificial. I see that different implementation do that differently and java follows the same pattern.

README.md
```diff
@@ -81,7 +81,7 @@ Additional packages, demos and tools are hosted or distributed in the [OpenTelem
 |---------|--------|---------|
 | Traces  | Beta   | N/A     |
 | Metrics | Beta   | N/A     |
-| Logs    | N/A    | N/A     |
+| Logs    | Alpha  | N/A     |
```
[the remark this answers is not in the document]
brettmc: I'm thinking we need a bit of a docs overhaul before RC, to change it from alpha/developer-focussed to end-user focussed? And link to the official opentelemetry.io docs more, too.

src/SDK/Logs/ReadableLogRecord.php
brettmc: Agreed, I've reworked this part a bit, and removed `LogRecordDate` - I think it's closer to spec now.

README.md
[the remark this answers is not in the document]
brettmc: added some documentation

src/SDK/Logs/LogRecordExporterInterface.php
```diff
@@ -0,0 +1,18 @@
...
+use OpenTelemetry\SDK\Common\Future\CancellationInterface;
+use OpenTelemetry\SDK\Common\Future\FutureInterface;
+
+interface LogRecordExporterInterface
+{
+    /**
+     * @param iterable<ReadableLogRecord> $batch
+     */
+    public function export(iterable $batch, ?CancellationInterface $cancellation = null): FutureInterface;
+    public function forceFlush(?CancellationInterface $cancellation = null): bool;
```
pdelewski, line 16: I see that these two functions `forceFlush` and `shutdown` are declared in at least few interfaces. `LogRecordProcessorInterface` contains them as well. They are also in other places inside SDK. Maybe we could extract common interface?

pdelewski:
@brettmc Overall, LGTM so far.

src/SDK/Logs/LogRecordExporterInterface.php
brettmc: That's true, I can find 10 interfaces in the SDK that implement both `forceFlush` and `shutdown`, in two different flavours (some have cancellation and some don't).
I guess we could extract those methods into two more general interfaces and have all those other interfaces extend one, but struggling to think of a good name for them.

src/SDK/Logs/ReadableLogRecord.php
```diff
@@ -0,0 +1,86 @@
...
+        return $this->body;
+    }
+
+    public function getAttributes(): AttributesInterface
+    {
+        $factory = $this->loggerSharedState->getLogRecordLimits()->getAttributeFactory();
+        $builder = $factory->builder($this->attributes);
+
+        return $builder->build();
+    }
```
Nevay, line 85: We should convert the attributes when constructing the `LogRecord`, otherwise attribute limits won't affect memory usage of records that are queued in a batch processor.

src/SDK/Logs/Processor/BatchLogsProcessor.php
```diff
@@ -0,0 +1,275 @@
...
+            return;
+        }
+
+        if ($this->queueSize === $this->maxQueueSize) {
+            $this->dropped++;
+
+            return;
+        }
+
+        $record->setContext($context);
```
Nevay, line 157: IMO this should be done by the `Logger`, not within the processors.

src/API/Logs/LoggerInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Logs;
+
+use Psr\Log\LoggerInterface as Psr3LoggerInterface;
+
+interface LoggerInterface extends Psr3LoggerInterface
```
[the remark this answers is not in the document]
brettmc: agreed and removed, I missed that part of the spec.

src/SDK/Logs/Processor/BatchLogsProcessor.php
brettmc: This is the way that I interpret the spec says to do it: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#onemit - the Logger passes a context on to the processor.

src/SDK/Logs/ReadableLogRecord.php
brettmc: I don't think we can do it during LogRecord constructor; LogRecord is an [API method](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logrecord) but attribute limits are an [SDK feature](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#logrecord-limits)

However, I can drop/convert attributes when creating a `ReadableLogRecord` from a `LogRecord` which I think achieves the same thing.

src/Contrib/Otlp/LogsConverter.php
```diff
@@ -0,0 +1,108 @@
...
+        $scopeLogs = new ScopeLogs();
+        $resourceLogs = new ResourceLogs();
+
+        foreach ($logs as $log) {
+            if ($resourceLogs->getResource() === null) {
+                $resourceLogs->setResource($this->convertResource($log->getResource()));
+            }
+            if ($scopeLogs->getScope() === null) {
+                $scopeLogs->setScope($this->convertInstrumentationScope($log->getInstrumentationScope()));
+            }
```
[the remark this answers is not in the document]
brettmc: done.

src/SDK/Logs/ReadableLogRecord.php
Nevay: We could change `Logger::logRecord(LogRecord)` to `Logger::logRecordBuilder(): LogRecordBuilder` to apply attribute limits on `::setAttribute()`. (I'm not sure how useful that would be considering that the logging implementation will most likely keep references to the original values until the log record is emitted.)

([Java uses Logger::logRecordBuilder()](https://github.com/open-telemetry/opentelemetry-java/blob/ff79d892a379f4b26df22447863616f083dcfcf1/api/logs/src/main/java/io/opentelemetry/api/logs/Logger.java#L29))

pdelewski: @Nevay I don't see advantages (at least as I'm writing this) introducing `LogRecordBuilder`. Seems like implementation detail

src/SDK/Logs/Processor/BatchLogsProcessor.php
brettmc: I see what you mean. Java accepts but ignores `context`, having already added `spanContext` to the record. I guess that's still spec-compliant. Will make that change.

src/API/Logs/LoggerInterface.php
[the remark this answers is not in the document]
brettmc: update: on review, the otel spec does explicitly disallow this: "The API is not intended to be called by application developers directly." so I've reverted psr-3.

bobstrecansky (approved):
Woohoo!  Great work!

PR #952, opened 2023-03-31
Resource detector

piotrantosik:
- add composer to defaults
- update readme: missing info about container & composer

### DIFF
--- a/tests/Integration/SDK/ResourceInfoFactoryTest.php
+++ b/tests/Integration/SDK/ResourceInfoFactoryTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace OpenTelemetry\Tests\Integration\SDK\Resource;
 
 use AssertWell\PHPUnitGlobalState\EnvironmentVariables;
+use Composer\InstalledVersions;
 use OpenTelemetry\SDK\Resource\ResourceInfoFactory;
 use OpenTelemetry\SemConv\ResourceAttributes;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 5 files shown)


tests/Integration/SDK/ResourceInfoFactoryTest.php
```diff
@@ -43,10 +44,11 @@ public function test_all_default_resources(): void
         $this->assertNotNull($resource->getAttributes()->get(ResourceAttributes::TELEMETRY_SDK_LANGUAGE));
         $this->assertNotNull($resource->getAttributes()->get(ResourceAttributes::TELEMETRY_SDK_VERSION));
         $this->assertNotNull($resource->getAttributes()->get(ResourceAttributes::SERVICE_NAME));
+        $this->assertNotNull($resource->getAttributes()->get(ResourceAttributes::SERVICE_VERSION));
 
         $this->assertEquals('opentelemetry', $resource->getAttributes()->get(ResourceAttributes::TELEMETRY_SDK_NAME));
         $this->assertEquals('php', $resource->getAttributes()->get(ResourceAttributes::TELEMETRY_SDK_LANGUAGE));
-        $this->assertEquals('unknown_service:php', $resource->getAttributes()->get(ResourceAttributes::SERVICE_NAME));
+        $this->assertEquals('open-telemetry/opentelemetry', $resource->getAttributes()->get(ResourceAttributes::SERVICE_NAME));
```
brettmc, line 51: this doesn't look right, according to spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service (my understanding is if the service name was not explicitly set to something, then `unknown_service` must be used)

piotrantosik: partly right - but if we add the composer resource to the default it will pick that name from the main composer.json file - if there is no composer.json file it will return `unknown_service:php` as in sdk default
I don't know if we can test this any other way.
maybe that was the reason why composer resource has not been added to default before?

brettmc: @piotrantosik you might be right...if I was part of that decision I've long ago forgotten it. If composer detector is installed and a higher-precedence method is used to set service name (`OTEL_SERVICE_NAME` or I guess anything user-provided), which is used?

brettmc: ok, that seems reasonable.

PR #959, opened 2023-04-05
documenting logs bridge api, adding memory exporter

brettmc:
Documenting what a logs bridge API is.
Adding an InMemory exporter for logging, to aide testing.
Adding builders for LoggerProvider.


brettmc:
NB that this is pending some clarification on the maintainers channel, but I think this is the right way to do the bridge API.

tests/Unit/SDK/SdkTest.php
```diff
@@ -76,7 +77,8 @@ public function test_getters(): void
         $propagator = $this->createMock(TextMapPropagatorInterface::class);
         $meterProvider = $this->createMock(MeterProviderInterface::class);
         $tracerProvider = $this->createMock(TracerProviderInterface::class);
-        $sdk = new Sdk($tracerProvider, $meterProvider, $propagator);
+        $loggerProvider = $this->createMock(LoggerProviderInterface::class);
+        $sdk = new Sdk($tracerProvider, $meterProvider, $loggerProvider, $propagator);
         $this->assertSame($propagator, $sdk->getPropagator());
         $this->assertSame($meterProvider, $sdk->getMeterProvider());
         $this->assertSame($tracerProvider, $sdk->getTracerProvider());
```
bobstrecansky, line 84: **question**: Do we need to have a:
`$this->assertSame($tracerProvider, $sdk->getLoggerProvider());`

bobstrecansky (approved):
On test question, but not a blocker.  LGTM.

src/SDK/Logs/LoggerProviderBuilder.php
```diff
@@ -0,0 +1,38 @@
...
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+
+class LoggerProviderBuilder
+{
+    private ?LogRecordProcessorInterface $processor = null;
+    private ?ResourceInfo $resource = null;
+
+    public function addLogRecordProcessor(LogRecordProcessorInterface $processor): self
+    {
+        $this->processor = $processor;
```
Nevay, line 19:
```suggestion
        $this->processors[] = $processor;
```

src/SDK/Logs/LoggerProviderBuilder.php
```diff
@@ -0,0 +1,38 @@
...
+
+        return $this;
+    }
+    public function addResource(ResourceInfo $resource): self
```
Nevay, line 23:
```suggestion
    public function setResource(ResourceInfo $resource): self
```

src/API/Logs/Bridge.php
```diff
@@ -0,0 +1,35 @@
...
+use OpenTelemetry\API\Common\Instrumentation\Globals;
+
+/**
+ * OpenTelemetry logs bridge API. To be used when implementing a handler/appender for existing logging libraries.
+ * This API is not meant to be used by application developers, but rather by library developers for the purpose
+ * of integrating existing logging libraries with OpenTelemetry.
+ *
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md
+ */
+class Bridge
```
Nevay, line 16: What is the purpose of this class / why should `new Bridge('scope', $loggerProvider)` be used instead of `$loggerProvider->getLogger('scope')`?
> The Logs Bridge API consist of these main classes:
    [LoggerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#loggerprovider) is the entry point of the API. It provides access to Loggers.
    [Logger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logger) is the class responsible for emitting logs as [LogRecords](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logrecord).

> The log appender implementation will typically acquire a [Logger](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logger) from the global [LoggerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#loggerprovider) at startup time, then call [Emit LogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#emit-logrecord) for LogRecords received from the application.

src/SDK/Logs/LoggerProviderBuilder.php
brettmc: I think you're right that we need to support multiple processors (the spec is clearer for logs, whereas in traces having more than one processor is optional). I'll change these methods to `set` for now, but create an issue to support multiple.

src/API/Logs/Bridge.php
brettmc: It's a good question - I am struggling to work out what the bridge API "is". Is it a class (like I've implemented here), or an interface for log appenders to implement, or is it even looser than that: some stated criteria that a log appender should conform to?
The second quote there hints that it's the third option; a log appender satisfies bring a bridge API if it does the things that the spec says. I don't think we can enforce that via interfaces, perhaps as an abstract class?

brettmc: I've updated the PR to remove Bridge, and added a small README describing bridge/log-appender expectations.

PR #965, opened 2023-04-12
update semantic conventions to v1.19

brettmc:
taking a lead from the go SIG, who seem to be using semconv the most, I've split
semantic conventions up into smaller files representing top-level semconvs.
I can't find any usage of the various AttributeValues being used, and other SIGs
no longer generate them, so removing them
[BREAKING] some attributes have moved, notably EXCEPTION_ESCAPED from TraceAttributes to EventAttributes, which requires a corresponding fix in some contrib packages.


brettmc:
asked a question on spec channel about how to organize semconv - it looks like other SIGs are starting to create "resource" + "everything else" which is different to what we're doing.

src/API/Trace/functions.php
```diff
@@ -35,7 +35,7 @@ function trace(SpanInterface $span, Closure $closure, iterable $args = [])
         return $c(...$a, ...($a = []));
     } catch (Throwable $e) {
         $s->setStatus(StatusCode::STATUS_ERROR, $e->getMessage());
-        $s->recordException($e, [TraceAttributes::EXCEPTION_ESCAPED => true]);
+        $s->recordException($e, [EventAttributes::EXCEPTION_ESCAPED => true]);
```
Nevay, line 38:
```suggestion
        $s->recordException($e, ['exception.escaped' => true]);
```
Can we drop the `open-telemetry/sem-conv` from the API-package to avoid breaking the API on sem-conv updates?

brettmc: That's a good question. If we did drop it, what is the point of having the sem-conv package? I've noticed that other SIGs version their semantic convention package the same as the spec version (eg 1.19.0 in this case), and presumably then pin APIs etc to that version. I have a feeling that might be very painful if packages are dependant on different versions, though.

The other thing some SIGs are doing is moving all attributes into two classes: `ResourceAttributes`, and `SemanticAttributes`, the latter being "everything except resources". That should at least solve the problem of things moving about as the spec evolves.

Nevay: IMO contrib/instrumentation packages are the main consumers of the sem-conv package.
(Java added removed attributes to the template to avoid a BC-break https://github.com/open-telemetry/opentelemetry-java/commit/188210bdba24d9ba45d8fcf9edd68deacc98afd4, I like this approach because it makes pinning to a specific version in instrumentation packages obsolete when specifying the schema url as literal instead of relying on `TraceAttributes::SCHEMA_URL`.)

Regarding the API dependency: even if we resolve the BC-break I don't see a reason to keep the sem-conv dependency in the API package, we only use it for a small helper function. Iff `::recordException()` / `exception.escaped` are changed in v1 in a breaking way, we should deprecate the `trace()` function. (The SDK uses literals for the other 'exception.' attributes in `::recordException()` too / any breaking change would very likely also affect the SDK.)

brettmc: Thanks for pointing that out - adding removed items to the template and marking them as `@deprecated` is a good idea. I've loosely copied java's deprecation pattern, and decided to remove the new semver classes I've added, since java hasn't done anything yet and I'd be more comfortable if they go first and we copy.

cedricziel (approved):

bobstrecansky (approved):

kayw-geek:
Hi @brettmc, I noticed have many const be deprecated, like the `TraceAttributes::HTTP_HOST`, What should I use to replace these deprecated constants?

brettmc:
The ones that we deprecate (from https://github.com/open-telemetry/opentelemetry-php/blob/main/script/semantic-conventions/templates/trace_deprecations.php.partial) are attributes that were removed from upstream, and we add them back with the deprecation notice to try to avoid breaking changes for users. Because they were totally removed from semconv (rather than being deprecated there), we don't have further info about which attributes to use instead.

In the case of `http.host`, it was removed from https://github.com/open-telemetry/opentelemetry-specification/issues/2114 and from scanning the comments I think `net.host.name`. I'd suggest referring to https://opentelemetry.io/docs/specs/semconv/http/http-spans/ and finding the attributes that best suit the values you want to use.

kayw-geek:
@brettmc  I got it, Thank you for your detailed reply.

PR #971, opened 2023-04-14
fixing otlp json trace/span ids

### LINKED ISSUE #969: RuntimeException(code: 400)
Following example https://github.com/open-telemetry/opentelemetry-php/blob/main/examples/traces/exporters/otlp_http_json.php
fails with following issue 
```
[2023-04-12T12:18:06.647406+00:00] otlp-example.ERROR: Export failure {"exception":"[object] (RuntimeException(code: 400): Bad Request at /Users/pdelewski/Projects/agent-apm-php/vendor/open-telemetry/sdk/Common/Export/Http/PsrTransport.php:106
```
It seems to be related to incorrect format of `trace_id` and `span_id`

brettmc:
The encoding applied by protobuf, which works for protobuf-encoded output, isn't what http/json requires. Rather than build a json serializer, this change modifies the almost-correct JSON output generated by protobuf, replacing the base64+binary-encoded trace ids and span ids with their original values.

Fixes #969

### DIFF
--- a/examples/logs/exporters/otlp_http_json_with_trace.php
+++ b/examples/logs/exporters/otlp_http_json_with_trace.php
@@ -9,21 +9,40 @@ use OpenTelemetry\API\Logs\EventLogger;
 use OpenTelemetry\API\Logs\LogRecord;
 use OpenTelemetry\Contrib\Otlp\LogsExporter;
 use OpenTelemetry\Contrib\Otlp\OtlpHttpTransportFactory;
+use OpenTelemetry\Contrib\Otlp\SpanExporter;
 use Opentelemetry\Proto\Logs\V1\SeverityNumber;
 use OpenTelemetry\SDK\Common\Instrumentation\InstrumentationScopeFactory;
 use OpenTelemetry\SDK\Logs\LoggerProvider;
 use OpenTelemetry\SDK\Logs\LogRecordLimitsBuilder;
 use OpenTelemetry\SDK\Logs\Processor\SimpleLogsProcessor;
+use OpenTelemetry\SDK\Trace\SpanProcessor\SimpleSpanProcessor;
+use OpenTelemetry\SDK\Trace\TracerProvider;
 use Psr\Log\LogLevel;
 
 require __DIR__ . '/../../../vendor/autoload.php';
+
+/**
+ * This example demonstrates sending logs and traces via OTLP http/json. The logs should contain the trace id
+ * and span id of the active span.
+ */
+
+echo 'Starting OTLP+json with trace example' . PHP_EOL;
 
 LoggerHolder::set(
     new Logger('otel-php', [new StreamHandler(STDOUT, LogLevel::DEBUG)])
 );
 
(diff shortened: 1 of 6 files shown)


src/Contrib/Otlp/ProtobufSerializer.php
```diff
@@ -50,10 +50,9 @@ public function serialize(Message $message): string
             case self::PROTOBUF:
                 return $message->serializeToString();
             case self::JSON:
-                //@todo https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#json-protobuf-encoding
-                return $message->serializeToJsonString();
+                return self::fixJsonOutput($message->serializeToJsonString());
```
pdelewski, line 53: That's 1 option I mentioned in #969. Do you think that it's good enough or at some point we would like to rewrite whole serializer?

examples/logs/exporters/otlp_http_json_with_trace.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+echo 'Starting OTLP+json with trace example' . PHP_EOL;
+
+LoggerHolder::set(
+    new Logger('otel-php', [new StreamHandler(STDOUT, LogLevel::DEBUG)])
+);
+
+$transport = (new OtlpHttpTransportFactory())->create('http://collector:4318/v1/logs', 'application/json');
+$exporter = new LogsExporter($transport);
```
pdelewski, line 35: I would suggest to change name to `logsExporter` or move this into initialization of `SimpleLogsProcessor` to be consistent with `traceExporter`

src/Contrib/Otlp/ProtobufSerializer.php
brettmc: I think we probably will need to write a custom serializer at some point, given the other documented differences between json and other protobuf. I might go ask around in the collector SIG for advice here.
In the meantime, I think I'll refactor this PR to move most of this logic into a `JsonSerializer` class.

Nevay:
Does using `base64_decode($span->getContext()->getTraceId())` instead of `$span->getContext()->getTraceIdBinary()` etc. work? If yes, we could add a flag to `SpanConverter::convert()` that switches between the two modes.

brettmc:
> Does using base64_decode($span->getContext()->getTraceId()) instead of $span->getContext()->getTraceIdBinary() etc. work? If yes, we could add a flag to SpanConverter::convert() that switches between the two modes.

It doesn't work in my testing: because trace_id etc are defined as `bytes`, protobuf always [applies base64 encoding](https://github.com/protocolbuffers/protobuf/blob/main/php/src/Google/Protobuf/Internal/GPBJsonWire.php#L222) which the collector chokes on.

I suppose that if we were to propose some JSON-related options for protobuf, we could invent a `encodeBytes` option to control that capability, too. It looks like there is some precedent in the Ruby protobuf implementation for switching encoding bytes on/off...

pdelewski (approved):

brettmc:
Closing in favour of #973

PR #986, opened 2023-04-28
resource info spec compliance, instrumentation attribute

brettmc:
* When merging resources, later (updating) keys should override existing keys, per spec. Update to work this way, and adjust some tests accordingly.
* Reorganise detectors so that environment runs last (per spec, OTEL_SERVICE_NAME should be the highest priority way of setting service name)
* Adding the experimental telemetry.auto.version attribute, if the auto-instrumentation extension is installed.


src/SDK/Resource/ResourceInfoFactory.php
```diff
@@ -13,17 +13,17 @@
...
-     * @return ResourceInfo
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/resource/sdk.md#merge
      */
     public static function merge(ResourceInfo ...$resources): ResourceInfo
     {
         $attributes = [];
 
         foreach ($resources as $resource) {
-            $attributes += $resource->getAttributes()->toArray();
+            $attributes = array_merge($attributes, $resource->getAttributes()->toArray());
```
Nevay, line 26: Can't use `array_merge`, the attributes array can contain numeric keys.

Nevay (commented):
IMO we should just specify that the resources are merged with descending priority, the spec doesn't dictate the argument order.

src/SDK/Resource/ResourceInfoFactory.php
brettmc: > the spec doesn't dictate the argument order

I think it does..."old" and "updating" resource params are listed in that order:

> Required parameters:
> * the old resource
> * the updating resource whose attributes take precedence

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/resource/sdk.md#merge

And in the case that more than two Resources are provided, I think it then follows that later ones take precedence.

> attributes array can contain numeric keys

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/common/README.md#attribute says `The attribute key MUST be a non-null and non-empty string` - perhaps we need to cast numeric keys to strings, which should resolve not being able to use `array_merge`... It's certainly not covered by any tests, and probably should be. What do you think?

brettmc: @Nevay you were right about `array_merge` (I didn't realise PHP forces numeric-looking keys to numbers internally), so I've instead applied array_reverse to deal with precedence. I added some tests around numeric keys + merging.

tests/Unit/SDK/Common/Attribute/AttributesTest.php
```diff
@@ -41,6 +41,16 @@ public function test_numeric_attribute_name(): void
         }
     }
 
+    public function test_integer_attribute_key_is_cast_to_string(): void
```
Nevay, line 44: Test above (`test_numeric_attribute_name`) can be removed, duplicates this test.

Nevay (approved):

src/SDK/Resource/ResourceInfoFactory.php
Nevay: Does the spec specify that the parameter order must/should be followed? I always assumed that this is entirely up to the implementation.
(fwiw I'm using `Resource::merge(...$this->resources, _default: Resource::getDefault())` in my implementation which can't be written as concisely with ascending priority.)

brettmc: Not exactly, but it does have this `MUST`:

> `If a key exists on both the old and updating resource, the value of the updating resource MUST be picked (even if the updated value is empty)`

To me, that's saying that if you start with a resource, then merge a second resource, then the second resource's keys take precedence and should overwrite the first.

Looking at other language implementations, `merge` is usually a method on a resource eg `$resourceOne->merge($resourceTwo)` which I think is clearer as to which is the old and the updating resource, and the spec was probably written with that in mind.

brettmc:
Sorry, I broke my forked repo and started over, and this got closed when I deleted my fork. Recreated in #989

PR #989, opened 2023-05-03
resourceinfo spec compliance, instrumentation attribute

brettmc:
When merging resources, later (updating) keys should override existing keys, per spec. Update to work this way, and adjust some tests accordingly. Reorganise detectors so that environment runs last (per spec, OTEL_SERVICE_NAME should be the highest priority way of setting service name) Adding the experimental telemetry.auto.version attribute, if the auto-instrumentation extension is installed.

Replaces #986


Nevay:
> Looking at other language implementations, merge is usually a method on a resource eg $resourceOne->merge($resourceTwo) which I think is clearer as to which is the old and the updating resource, and the spec was probably written with that in mind.

I agree, changing it to an instance method seems preferably; it also breaks currently used static calls instead of silently changing the behavior.

bobstrecansky (approved):

brettmc:
> I agree, changing it to an instance method seems preferably; it also breaks currently used static calls instead of silently changing the behavior.

This is a good idea. I've reverted changes to the static method and deprecated it, and then implemented an instance method `ResourceInfo::merge()`

Nevay (approved):

PR #990, opened 2023-05-03
temporality preference from env

brettmc:
Implement OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE, which controls the temporality of meters when creating from environment vars. Add OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION to variables, although we do not yet have anything that uses it.

### DIFF
--- a/src/Contrib/Otlp/MetricExporterFactory.php
+++ b/src/Contrib/Otlp/MetricExporterFactory.php
@@ -11,7 +11,6 @@ use OpenTelemetry\SDK\Common\Configuration\Variables;
 use OpenTelemetry\SDK\Common\Export\TransportFactoryInterface;
 use OpenTelemetry\SDK\Common\Export\TransportInterface;
 use OpenTelemetry\SDK\Common\Otlp\HttpEndpointResolver;
-use OpenTelemetry\SDK\Metrics\Data\Temporality;
 use OpenTelemetry\SDK\Metrics\MetricExporterFactoryInterface;
 use OpenTelemetry\SDK\Metrics\MetricExporterInterface;
 use OpenTelemetry\SDK\Registry;
@@ -69,14 +68,16 @@ class MetricExporterFactory implements MetricExporterFactoryInterface
         );
     }
 
-    private function getTemporality(): string
+    private function getTemporality(): ?string
     {
         $value = Configuration::getEnum(Variables::OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE);
         switch (strtolower($value)) {
-            case strtolower(Temporality::CUMULATIVE):
-                return Temporality::CUMULATIVE;
-            case strtolower(Temporality::DELTA):
-                return Temporality::DELTA;
+            case 'cumulative':
+                return 'cumulative';
+            case 'delta':
+                return 'delta';
+            case 'lowmemory':
+                return null;
             default:
                 throw new \UnexpectedValueException('Unknown temporality: ' . $value);
         }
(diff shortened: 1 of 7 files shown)


achetronic (approved):

src/SDK/Common/Configuration/KnownValues.php
```diff
@@ -71,6 +77,17 @@ interface KnownValues
         self::VALUE_HTTP_JSON,
     ];
 
+    public const VALUES_TEMPORALITY_PREFERENCE = [
+        self::VALUE_TEMPORALITY_CUMULATIVE,
+        self::VALUE_TEMPORALITY_DELTA,
+        //self::VALUE_TEMPORALITY_LOW_MEMORY, //not implemented
```
Nevay, line 83: `LowMemory` is supported by using `null` as exporter temporality.

src/Contrib/Otlp/MetricExporterFactory.php
```diff
@@ -67,6 +69,19 @@ private function buildTransport(string $protocol): TransportInterface
         );
     }
 
+    private function getTemporality(): string
+    {
+        $value = Configuration::getEnum(Variables::OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE);
+        switch (strtolower($value)) {
+            case strtolower(Temporality::CUMULATIVE):
```
Nevay, line 76: Should use literals, all usages of `Temporality` should treat these constants as [`string|Temporality`](https://github.com/open-telemetry/opentelemetry-php/blob/1976f53f92fabdcd7f53f83fccacfb95fde007de/src/SDK/Metrics/Data/Temporality.php#L10) to allow converting the class to an enum once support for 8.0 is dropped.

bobstrecansky (approved):

src/Contrib/Otlp/MetricExporterFactory.php
```diff
@@ -67,6 +68,21 @@ private function buildTransport(string $protocol): TransportInterface
         );
     }
 
+    private function getTemporality(): ?string
+    {
+        $value = Configuration::getEnum(Variables::OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE);
+        switch (strtolower($value)) {
+            case 'cumulative':
```
pdelewski, line 75: I think we could merge these two cases 
```
case 'cumulative':
case 'delta':
  return $value;
```

src/Contrib/Otlp/MetricExporterFactory.php
```diff
@@ -67,6 +68,21 @@ private function buildTransport(string $protocol): TransportInterface
...
+                return 'cumulative';
+            case 'delta':
+                return 'delta';
+            case 'lowmemory':
+                return null;
+            default:
+                throw new \UnexpectedValueException('Unknown temporality: ' . $value);
+        }
+    }
```
Nevay, line 84:
```suggestion
    private function getTemporality()
    {
        $value = Configuration::getEnum(Variables::OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE);
        switch (strtolower($value)) {
            case 'cumulative':
                return Temporality::CUMULATIVE;
            case 'delta':
                return Temporality::DELTA;
            case 'lowmemory':
                return null;
            default:
                throw new \UnexpectedValueException('Unknown temporality: ' . $value);
        }
    }
```

Nevay: The correct typehint would be `string|Temporality|null`, union types are >= 8.0.

PR #999, opened 2023-05-09
adding resource detectors to registry

brettmc:
To enable registering non-SDK resource detectors within the SDK, extend the SDK Registry to also handle resource detectors.
Detectors will be looked up by name and applied as part of 'all', or as part of a list of detectors, or if a list of detectors is provided which contains a non-SDK detector name.


src/SDK/Resource/ResourceInfoFactory.php
```diff
@@ -38,17 +43,21 @@ public static function defaultResource(): ResourceInfo
...
+            $all = (new Detectors\Composite([
                 new Detectors\Host(),
                 new Detectors\OperatingSystem(),
                 new Detectors\Process(),
                 new Detectors\ProcessRuntime(),
-                new Detectors\Container(),
                 new Detectors\Sdk(),
                 new Detectors\SdkProvided(),
                 new Detectors\Composer(),
                 new Detectors\Environment(),
```
Nevay, line 54:
```suggestion
                ...Registry::resourceDetectors(),
                new Detectors\Environment(),
```
Should merge env resource after merging `Registry::resourceDetectors()`.

PR #1005, opened 2023-05-16
adding console metric exporter

brettmc:
allow the existing metric exporter to be created from env vars, by creating a factory and registering it against 'console'. Now, `OTEL_METRICS_EXPORTER=console` will create a metrics provider with console output. Note that it lives in contrib/otlp, since it uses the existing otlp converters.

### DIFF
--- a/examples/metrics/exporters/otlp_grpc.php
+++ b/examples/metrics/exporters/otlp_grpc.php
@@ -20,7 +20,6 @@ $reader = new ExportingReader(
     new MetricExporter(
         (new GrpcTransportFactory())->create('http://collector:4317' . OtlpUtil::method(Signals::METRICS))
     ),
-    $clock
 );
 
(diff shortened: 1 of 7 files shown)


examples/metrics/exporters/otlp_grpc.php
```diff
@@ -20,7 +20,6 @@
     new MetricExporter(
         (new GrpcTransportFactory())->create('http://collector:4317' . OtlpUtil::method(Signals::METRICS))
     ),
-    $clock
```
pdelewski, line 23: Seems that our examples are often out of date. Would be good to think about better consistency.

pdelewski (approved):

examples/metrics/exporters/otlp_grpc.php
brettmc: I agree. Maybe we can get our static code analysis tools to also run against the examples?

PR #1036, opened 2023-06-16
adding support info to sdk composer

brettmc:
adding some links to try to make it easier for users to find the right support/repositories etc.
just doing one composer file to see if it works, and will follow-up with more in a later PR

### DIFF
--- a/src/SDK/composer.json
+++ b/src/SDK/composer.json
@@ -7,7 +7,7 @@
         "issues": "https://github.com/open-telemetry/opentelemetry-php/issues",
         "source": "https://github.com/open-telemetry/opentelemetry-php",
         "docs": "https://opentelemetry.io/docs/php",
-        "forum": "https://app.slack.com/client/T08PSQ7BQ/C01NFPCV44V"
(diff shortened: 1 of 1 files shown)


src/SDK/composer.json
```diff
@@ -3,6 +3,12 @@
     "description": "SDK for OpenTelemetry PHP.",
     "keywords": ["opentelemetry", "otel", "metrics", "tracing", "logging", "apm", "sdk"],
     "type": "library",
+    "support": {
+        "issues": "https://github.com/open-telemetry/opentelemetry-php/issues",
+        "source": "https://github.com/open-telemetry/opentelemetry-php",
+        "docs": "https://opentelemetry.io/docs/php",
+        "forum": "https://app.slack.com/client/T08PSQ7BQ/C01NFPCV44V"
```
brettmc, line 10: "forum" or "chat" ? https://getcomposer.org/doc/04-schema.md#support

bobstrecansky: I feel like chat would be the more appropriate thing here, there really isn't an OT "forum" per se.

PR #1037, opened 2023-06-16
deprecate sdk/common/dsn

brettmc:
I can't find any usage of these classes anywhere in main or contrib repo. rather than delete, lets deprecate and remove closer to release candidate


bobstrecansky (approved):
I'm pretty sure these are vestigial too.

PR #1044, opened 2023-06-20
fixing backwards-compatibility

brettmc:
- sdk BC was not being loaded at all in the sdk package (only in the monorepo), so add it to autoload->files
- fixing api/common BC layer (removing .php and fixing one of the FQCNs)


brettmc:
this should fix most of the build failures over in -contrib... they're using dev-main (because of a branch alias), and not finding some of the recently-moved files.

pdelewski (approved):

PR #1055, opened 2023-06-26
adding sdk console metrics exporter

brettmc:
to make getting started easier (in opentelemetry.io docs), create an sdk-based metrics exporter. this means we can export all signals without needing a protobuf implementation. since console export is only useful for playing around, the change in format to a simpler human-readable output seems reasonable.


pdelewski:
LGTM

pdelewski (approved):

PR #1067, opened 2023-07-03
require google/protobuf (native)

brettmc:
We made protobuf implementation optional back in #949 however it seems to be tripping up a number of developers. Since native + extension can be installed at the same time (with extension taking priority), I think that users will be less surprised when starting out if there is a working implementation by default.

### DIFF
--- a/proto/otel/README.md
+++ b/proto/otel/README.md
@@ -8,20 +8,19 @@
 
 ## Protobuf Runtime library
 
+OTLP exporting requires a [protobuf runtime library](https://github.com/protocolbuffers/protobuf/tree/main/php).
+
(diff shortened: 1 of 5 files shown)


brettmc:
The downside I see to this PR is that users who do choose to use the extension must still install the native package. I have a PR against protobuf which I think will resolve this, but getting changes into protobuf is a long game, so we can improve things if and when the change is accepted and released.

pdelewski:
> We made protobuf implementation optional back in #949 however it seems to be tripping up a number of developers. Since native + extension can be installed at the same time (with extension taking priority), I think that users will be less surprised when starting out if there is a working implementation by default.

I think that it's step in right direction as I was asking myself several times why we don't install default.

pdelewski (approved):

Issue #1069, opened 2023-07-03
SchemaURL missing in emitted OTLP log


tigrannajaryan:
**Steps to reproduce**
Run the code:
```php
<?php

declare(strict_types=1);

use OpenTelemetry\API\Logs\LogRecord;
use OpenTelemetry\SDK\Sdk;
use OpenTelemetry\SDK\Common\Attribute\Attributes;
use OpenTelemetry\SDK\Common\Instrumentation\InstrumentationScopeFactory;
use OpenTelemetry\SDK\Common\Time\ClockFactory;
use OpenTelemetry\SDK\Logs\Exporter\ConsoleExporterFactory;
use OpenTelemetry\SDK\Logs\LoggerProvider;
use OpenTelemetry\SDK\Logs\Processor\SimpleLogsProcessor;
use OpenTelemetry\SDK\Logs\Processor\BatchLogsProcessor;
use OpenTelemetry\API\Common\Instrumentation\Globals;
use OpenTelemetry\Contrib\Otlp\OtlpHttpTransportFactory;
use OpenTelemetry\Contrib\Otlp\LogsExporter;

require 'vendor/autoload.php';


$transport = (new OtlpHttpTransportFactory())->create('http://localhost:4318/v1/logs', 'application/json');

$exporter = new LogsExporter($transport);

$loggerProvider = new LoggerProvider(
    new SimpleLogsProcessor($exporter),
    new InstrumentationScopeFactory(Attributes::factory())
);

Sdk::builder()
    ->setLoggerProvider($loggerProvider)
    ->setAutoShutdown(true)
    ->buildAndRegisterGlobal();


$logger = Globals::loggerProvider()->getLogger('demo', '1.0', 'https://opentelemetry.io/schemas/1.7.1', ['scopefoo' => 'scopebar']);

$record = (new LogRecord(['foo' => 'bar', 'baz' => 'bat', 'msg' => 'hello world']));

$logger->emit($record);
```

**What is the expected behavior?**
SchemaURL should be populated in the emitted logs.

**What is the actual behavior?**
SchemaURL is missing. This is what Collector reports:
```
Resource SchemaURL:
Resource attributes:
     -> host.arch: Str(x86_64)
     -> host.name: Str(C02Z1080LVDM)
     -> os.description: Str(22.5.0)
     -> os.name: Str(Darwin)
     -> os.type: Str(darwin)
[...]

PR #1071, opened 2023-07-04
set schemaUrl for logs output

brettmc:
Fixes: #1069

### DIFF
--- a/examples/logs/exporters/otlp_http.php
+++ b/examples/logs/exporters/otlp_http.php
@@ -35,7 +35,7 @@ $loggerProvider = new LoggerProvider(
         (new LogRecordLimitsBuilder())->build()->getAttributeFactory()
     )
 );
-$logger = $loggerProvider->getLogger('demo', '1.0', 'http://schema.url', ['foo' => 'bar']);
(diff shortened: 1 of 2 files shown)


examples/logs/exporters/otlp_http.php
```diff
@@ -35,7 +35,7 @@
         (new LogRecordLimitsBuilder())->build()->getAttributeFactory()
     )
 );
-$logger = $loggerProvider->getLogger('demo', '1.0', 'http://schema.url', ['foo' => 'bar']);
+$logger = $loggerProvider->getLogger('demo', '1.0', 'https://opentelemetry.io/schemas/1.7.1', ['foo' => 'bar']);
```
pdelewski, line 38: Will we update this link everytime version will change?

pdelewski (approved):

examples/logs/exporters/otlp_http.php
pdelewski: yes, right

brettmc: That's a really good question. According to the docs, published schemas _should_ be there: https://opentelemetry.io/docs/specs/otel/schemas/#opentelemetry-schema
It doesn't stop the example from working, but I'll follow up with the other maintainers.

PR #1072, opened 2023-07-05
refactor attribute validation

### LINKED ISSUE #1074: Complex attributes must be allowed for LogRecord
Log data model (unlike traces and metrics) allows complex attribute values, non-homogeneous arrays, nested maps: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any

The current implementation disallows such values, but should not.

**Steps to reproduce**
Use this code to create a log record:
```php
$record = (new LogRecord(['foo' => 'bar', 'baz' => 'bat', 'msg' => 'hello world']))
    ->setSeverityText('INFO')
    ->setSeverityNumber(9)
    ->setAttributes(['attr1' => 'val1','attr2' => 123, 'arrnum' => [1,2,3], 'complex' => [1, 'a'], 'nested' => [123, 'abc', ['sub'=>'val']]]);

$logger->emit($record);
```

All attributes should be output.

Complex attributes `complex` and `nested` are dropped:
```
                "attributes": {
                    "attr1": "val1",
                    "attr2": 123,
                    "arrnum": [
                        1,
                        2,
                        3
                    ]
                },
```

**Additional context**
Part of https://github.com/open-telemetry/community/issues/1537

brettmc:
moving private method into a testable class, introduce attribute validation class which applies to most attribute. LogRecord attributes are allowed to be complex, so introduce a validator for these which lets everything through.

Fixes #1074

### DIFF
--- a/tests/Unit/SDK/Common/Attribute/AttributeValidatorTest.php
+++ b/tests/Unit/SDK/Common/Attribute/AttributeValidatorTest.php
@@ -68,7 +68,7 @@ class AttributeValidatorTest extends TestCase
         return [
             'empty array' => [[], true],
             'array of strings' => [['one', 'two'], true],
-            'array of int' => [[1, 2], true],
+            'array of ints' => [[1, 2], true],
             'array of double' => [[1.1, 1.2], true],
             'mixed numerics' => [[2, 3.5, PHP_INT_MAX], true],
             'array of bool' => [[true, false], true],
--- a/tests/Unit/SDK/Common/Attribute/AttributeValidatorTest.php
+++ b/tests/Unit/SDK/Common/Attribute/AttributeValidatorTest.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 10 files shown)


tests/Unit/SDK/Common/Attribute/AttributeValidatorTest.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        $this->assertSame($expected, $this->validator->validate($value));
+    }
+
+    public static function arrayProvider(): array
+    {
+        return [
+            'empty array' => [[], true],
+            'array of strings' => [['one', 'two'], true],
+            'array of int' => [[1, 2], true],
```
pdelewski, line 71: nitpick: to be consistent with above name, should be `array of ints`

brettmc, line 71:
```suggestion
            'array of ints' => [[1, 2], true],
```

PR #1081, opened 2023-07-15
Always weaken asynchronous callbacks

Nevay:
Removes the `$weaken` parameter and always weakens callbacks.
Returned `ObservableCallback`s keep a strong reference to the `$this` of the callback to prevent preemptive gc-ing if the user wants to detach the callback manually.


brettmc (approved):

brettmc:
LGTM. Can you tidy up/correct the comment in `examples/metrics/weak-reference-observables.php` now that the behavior has changed?

PR #1090, opened 2023-07-21
fix: fiber observers by FFI are not safe on ZTS

morrisonlevi:
They aren't _truly_ safe on NTS builds either, as the functions used are only expected to be set at certain times of the lifecycle, but in practice it's safe enough on NTS.

However, for ZTS, these APIs are not safe to call at runtime at all.

I haven't tested these changes. I simply saw this issue while reading the code and thought I'd offer a starting point by opening a PR instead of just an issue.

### DIFF
--- a/src/Context/README.md
+++ b/src/Context/README.md
@@ -35,7 +35,7 @@ It is recommended to use a `try-finally` statement after `::activate()` to ensur
 
 ### Fiber support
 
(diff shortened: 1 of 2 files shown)


brettmc (approved):

brettmc:
This looks like a sensible starting point. Would we be able to replicate this in a test against a ZTS build? Or is it more of an instability/"weird things might happen" issue?
I'm wondering whether switching our build over to use a ZTS build is worth considering?

morrisonlevi:
You might get "lucky" and it might work some times, but other times you might crash or otherwise corrupt memory. The APIs to add observers are not thread safe.

PR #1094, opened 2023-08-01
Fix for maxRetries=0

### LINKED ISSUE #1095: PsrTransport maxRetries off by one
The `$maxRetries` parameter of [`PsrTransport::_construct()`](https://github.com/open-telemetry/opentelemetry-php/blob/ff15cc7037cbdb58641785434b61fd1aa9f27085/src/SDK/Common/Export/Http/PsrTransport.php#L58) appears to be off by one - it's being used as "tries" rather than "retries".

This has the effect of going into an infinite loop if you try and disable retries by passing `0` for `$maxRetries`.

I've created a PR, but this would be a small backwards compatibility break.

stickeegreg:
Fixes #1095

### DIFF
--- a/src/SDK/Common/Export/Http/PsrTransport.php
+++ b/src/SDK/Common/Export/Http/PsrTransport.php
@@ -110,7 +110,7 @@ final class PsrTransport implements TransportInterface
                 return new ErrorFuture($e);
             }
 
-            if ($retries + 1 === $this->maxRetries) {
+            if ($retries === $this->maxRetries) {
                 return new ErrorFuture(new RuntimeException('Export retry limit exceeded', 0, $e));
             }
 
--- a/src/SDK/Common/Export/Http/PsrTransport.php
+++ b/src/SDK/Common/Export/Http/PsrTransport.php
@@ -110,7 +110,7 @@ final class PsrTransport implements TransportInterface
                 return new ErrorFuture($e);
             }
 
-            if ($retries === $this->maxRetries) {
+            if ($retries >= $this->maxRetries) {
                 return new ErrorFuture(new RuntimeException('Export retry limit exceeded', 0, $e));
             }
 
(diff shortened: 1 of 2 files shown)


src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -92,7 +92,7 @@ public function send(string $payload, ?CancellationInterface $cancellation = nul
             $request = $request->withAddedHeader($header, $value);
         }
 
-        for ($retries = 0;; $retries++) {
+        for ($tries = 0;; $tries++) {
```
brettmc, line 95: I don't think the variable name change is needed.

brettmc (changes requested):
There's a failing test with this change. Can you look into why, and make that green? I assume the problem here is that if max retire = 0 then it never tries. Does the spec say anything about valid values for maxRetries?

stickeegreg:
Thanks, I've fixed the test and added a new one, also undone renaming the variable.

I had a look at the spec, and as far as I can see it doesn't mention a retry limit at all.
Looking at the Go implementation it seems like they use a maximum time rather than number of tries.

src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -110,7 +110,7 @@ public function send(string $payload, ?CancellationInterface $cancellation = nul
                 return new ErrorFuture($e);
             }
 
-            if ($retries + 1 === $this->maxRetries) {
+            if ($retries === $this->maxRetries) {
```
brettmc, line 113: what happens if retries is negative? perhaps this should be a `>=` test...

stickeegreg: Tempted to say "you get what you deserve" if you pass a negative maxRetries 😂 
I've pushed up a change so it's `>=` now.

brettmc (approved):

PR #1107, opened 2023-08-31
adding 8.2 + 8.3 to build matrix

brettmc:
8.3 is in RC, so add it to the matrix.
bump 8.2 to non-experimental, which should have happened some time ago.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -13,11 +13,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


brettmc:
converting back to draft, to be continued after GA and when 8.3 is closer to GA

pdelewski:
> converting back to draft, to be continued after GA and when 8.3 is closer to GA

I was thinking about that too, otherwise we will have failing tests.

bobstrecansky (approved):

PR #1110, opened 2023-09-04
fixing some psalm 5 complaints

brettmc:
Whilst investigating upgrading to psalm 5, I notice that it generates a lot of new complaints. This fixes the ones that looked legit and require an API change, or were trivial. Does not upgrade to psalm 5 yet, though. That's a bigger job for another day.


src/API/Signals.php
```diff
@@ -4,9 +4,7 @@
 
 namespace OpenTelemetry\API;
 
-use InvalidArgumentException;
-
-class Signals
+interface Signals
```
pdelewski, line 7: Does psalm complain about class keyword? I must say that it's very strange pattern for me to use interface to store constants only

brettmc: No, but it did complain about unused private constructors in a few classes (where a class had been used to store constants, and private constructor used to stop it being instantiated).
We have a couple of interfaces already which are used just to store constants, so I've updated these ones to be consistent.

PR #1126, opened 2023-10-13
update semconv to 1.22.0

brettmc:
Since 1.19.0, the source has been split out of opentelemetry-specifiction into semantic-conventions. Update the build script accordingly. Update to latest generator.
Add 2 new deprecations for semconvs which were renamed since 1.19


brettmc:
Prior to merging this, I'd like to discuss whether we should tag semconv with the upstream version? i.e, publish our current version as 1.19.0 and this next one as 1.22.0.
Since the repo moved, I wasn't getting notifications of new releases, so we missed a couple. Meh.

bobstrecansky (approved):

PR #1136, opened 2023-11-02
otel-sig-security - adding SBOM to builds

bobstrecansky:
#otel-sig-security asked for SBOM to be added to builds.  This PR adds it to our php actions.  It may be too frequent, but we can ensure we have this for when we do our builds.


brettmc:
It ran in 4 seconds and looks to have found some PHP packages, which is a good start. The next question is: how do we surface this? (particularly given we don't do releases in this repo or in contrib)

brettmc:
@bobstrecansky It looks like this is successfully generating an SBOM, and it _claims_ to be uploading it as an artifact, but I can't see one.

Issue #1045, opened 2023-06-21
Auto opentelemetry configuration not working for laravel and php 8.x


abhisheksurve45:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your php version (`php -v` will tell you your current version), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on master.

Laravel: 7.2x
PHP: 8.x
Guzzle: 6.3
Composer: 2.x

**Steps to reproduce**
Describe exactly how to reproduce the error. Include a code sample if applicable.

Hi, trying to auto instrument a laravel application, following the documentation https://opentelemetry.io/docs/instrumentation/php/automatic/

Describing the steps taken for auto instrumenting the project:

1. Base docker image

Using base image `php:8.0-fpm-alpine`

Installed required packages

```dockerfile
RUN sudo apt-get install gcc make autoconf
```

2. Install `opentelemetry-beta` in dockerfile

```dockerfile
RUN pecl install opentelemetry-beta
```

3. Add `ini` file in php config

telemetry.ini

```
[opentelemetry]
extension=opentelemetry.so
```

```dockerfile
COPY config/telemetry.ini $PHP_INI_DIR/conf.d/
```

Post this I can confirm that package is installed in docker by

```bash
php -m | grep opentelemetry
opentelemetry
```

4. Set `discovery` to true

```bash
composer config allow-plugins.php-http/discovery true
```

composer.json

```json
"config": {
        "preferred-install": "dist",
        "platform-check": false,
        "allow-plugins": {
            "php-http/discovery": true
        }
    }
```

5. Composer install `guzzle7-adapter`, `open-telemetry/sdk` and `open-telemetry/opentelemetry-auto-slim`

`guzzle7-adapter` does not install due to required php 7.x error, however `open-telemetry/sdk` and `open-telemetry/opentelemetry-auto-slim` gets installed successfully

6. Add environment

`.env` file

```env
OTEL_SERVIE_NAME=test
OTEL_PHP_AUTOLOAD_ENABLED=true
OTEL_TRACES_EXPORTER=otlp
OTEL_EXPORTER_OTLP_ENDPOINT=http://xxx.xx.xx.xx:xxxxx
OTEL_PHP_DETECTORS=all
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
```

**What is the expected behavior?**
What did you expect to see?
[...]

PR #1144, opened 2023-11-17
Added Sdk::isDeveloperModeEnabled() helper.

ChrisLightfootWild:
Use case in mind:
 - allow extra functionality to be hooked during developer mode

Alternative:
 - create additional packages that can be added as composer dev-dependencies, which can register via autoloading in the usual way


brettmc:
Can you elaborate on what sort of extra functionality you're thinking of?

src/SDK/Sdk.php
```diff
@@ -43,6 +44,23 @@ public static function isInstrumentationDisabled(string $name): bool
...
+    /**
+     * Tests whether developer mode has been enabled by config.
+     */
+    public static function isDeveloperModeEnabled(): bool
+    {
+        if (Configuration::has(Variables::OTEL_PHP_DEVELOPER_MODE_ENABLED)) {
+            return Configuration::getBoolean(Variables::OTEL_PHP_DEVELOPER_MODE_ENABLED);
+        }
+
+        try {
```
brettmc, line 56: I think it should only be enabled by variable (and default to off). Is there any benefit other than convenience to also checking InstalledVersions?

ChrisLightfootWild: No, purely the convenience. I will go with your lead here, will remove this dependency/complexity.

ChrisLightfootWild:
> Can you elaborate on what sort of extra functionality you're thinking of?

It was along the lines of hooking things which may only be available as a dev-dependency, without having to split that into a separate package.

The split may be additional overhead, but happy either way. There may not be a strong enough use-case for this and that is fine too!

ChrisLightfootWild:
> Can you elaborate on what sort of extra functionality you're thinking of?

With https://github.com/open-telemetry/opentelemetry-php/issues/1045 in mind, to be able to hook additional classes which may or may not be present!

With that particular example, something akin to the below but without the overhead of the `hook` call where it is not required:
```php
if ($devMode && isset(\Illuminate\Foundation\Console\ServeCommand::$passthroughVariables)) {
    foreach ($_ENV as $key => $value) {
        if (str_starts_with($key, 'OTEL_')) {
            \Illuminate\Foundation\Console\ServeCommand::$passthroughVariables[] = $key;
        }
    }
}
```

brettmc:
Gotcha. My understanding of the observer API (mostly from reading https://www.datadoghq.com/blog/engineering/php-8-observability-baked-right-in/#target-functions-and-methods) is that it's very low overhead for methods that aren't called. There's no issue with observing non-existent methods, either (eg if the class is not installed). So, I'm not particularly worried about the overhead of extra hooks.

Even so, I think you could wrap the hook with an `if class_exists(DevDependency::class)` and that would achieve what you're after?

PR #1145, opened 2023-11-20
generate semconv for 1.23.1

brettmc:
a bug in 1.23.0 meant that cloud, container and oci resource attributes were incorrectly published under trace attributes. sort the attributes, so that future diffs are easier to read.


brettmc:
I've checked that the only changes to 1.23.0 are:
cloud + container + oci attributes moved from TraceAttribute to ResourceAttributes

The diff is large because I've chosen to sort the output now, which should make future semconv changes easier to comprehend.

bobstrecansky (approved):

PR #1149, opened 2023-12-05
Set default value for SchemaURL in cached instrumentation

cedricziel:
Based on our conversation [here](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/218#discussion_r1416347123), we probably want to apply the schema url as a default value.

### DIFF
--- a/src/API/Instrumentation/CachedInstrumentation.php
+++ b/src/API/Instrumentation/CachedInstrumentation.php
@@ -38,11 +38,11 @@ final class CachedInstrumentation
     /** @var ArrayAccess<LoggerProviderInterface, LoggerInterface>|null */
     private ?ArrayAccess $loggers;
 
-    public function __construct(string $name, ?string $version = null, string $schemaUrl = TraceAttributes::SCHEMA_URL, iterable $attributes = [])
+    public function __construct(string $name, ?string $version = null, ?string $schemaUrl = null, iterable $attributes = [])
     {
         $this->name = $name;
         $this->version = $version;
-        $this->schemaUrl = $schemaUrl;
+        $this->schemaUrl = ($schemaUrl !== null && $schemaUrl !== '') ? $schemaUrl : TraceAttributes::SCHEMA_URL;
         $this->attributes = $attributes;
         $this->tracers = self::createWeakMap();
         $this->meters = self::createWeakMap();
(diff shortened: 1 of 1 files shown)


brettmc (approved):

Nevay:
I dislike this change:
- it disallows the usage of `CachedInstrumentation` without schema url
- IMO instrumentation should specify the schema url as string literal instead of relying on the `TraceAttributes::SCHEMA_URL` constant, its value can change if the `open-telemetry/sem-conv` package is not pinned to a minor version; pinning it to a minor version would result in unnecessary dependency conflicts between instrumentation packages (related: https://github.com/open-telemetry/opentelemetry-php/pull/965#discussion_r1164001585); not pinning it would result in reporting an incorrect / newer schema url which might prevent [schema transformations](https://opentelemetry.io/docs/specs/otel/schemas/file_format_v1.1.0/#order-of-transformations).

(Iff we still want to do this: `open-telemetry/sem-conv` dependency is missing in `composer.json`)

brettmc:
Excellent points @Nevay and thanks for the feedback. I do agree that not setting the schemaUrl should be allowed (and probably the default).
I'm not entirely convinced about using a string from the libraries themselves; since we can't use more than one semconv composer dependency at a time (without perhaps rearchitecting semconv to contain historical versions), then it needs to work with any semconv version that fits its version constraints. The spec points to https://github.com/open-telemetry/opentelemetry-go/blob/8ba6da8f3e5380629fcd72057fe0f827a9e23493/sdk/resource/os.go#L50 as an example of how to do it, and that example uses a value from the semconv package.

Nevay:
> since we can't use more than one semconv composer dependency at a time (without perhaps rearchitecting semconv to contain historical versions), then it needs to work with any semconv version that fits its version constraints.

Our semconv package should be fully backwards compatible (besides the changed value of `SCHEMA_URL`) as we deprecate removed attributes instead of removing them.

> The spec points to https://github.com/open-telemetry/opentelemetry-go/blob/8ba6da8f3e5380629fcd72057fe0f827a9e23493/sdk/resource/os.go#L50 as an example of how to do it, and that example uses a value from the semconv package.

I am not familiar with go, but they do seem to provide a new namespace for every version and import a specific version [import (semconv "go.opentelemetry.io/otel/semconv/v1.4.0")](https://github.com/open-telemetry/opentelemetry-go/blob/8ba6da8f3e5380629fcd72057fe0f827a9e23493/sdk/resource/os.go#L22) / https://github.com/open-telemetry/opentelemetry-go/tree/main/semconv -> their `semconv.SchemaURL` won't change on semconv updates.

bobstrecansky:
Are we still working on this?  It can probably be closed otherwise.

PR #1186, opened 2023-12-09
Add instrument advisory parameter

Nevay:
See [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.28.0/specification/metrics/api.md#instrument-advisory-parameters), adds an optional `$advisory` parameter to all `Meter::createXXX()` methods.

Use case [`http.server.request.duration`](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.1/docs/http/http-metrics.md#metric-httpserverrequestduration):
> This metric SHOULD be specified with [ExplicitBucketBoundaries](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.26.0/specification/metrics/api.md#instrument-advisory-parameters) of [ 0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10 ].

### DIFF
--- a/src/SDK/Metrics/Meter.php
+++ b/src/SDK/Metrics/Meter.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\SDK\Metrics;
 
+use function array_unshift;
 use ArrayAccess;
+use function is_callable;
 use OpenTelemetry\API\Metrics\CounterInterface;
 use OpenTelemetry\API\Metrics\HistogramInterface;
 use OpenTelemetry\API\Metrics\MeterInterface;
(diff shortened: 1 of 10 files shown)


src/SDK/Metrics/Meter.php
```diff
@@ -133,25 +147,31 @@ public function createObservableGauge(string $name, ?string $unit = null, ?strin
...
+            $advisory,
         );
 
         return new UpDownCounter($this->writer, $instrument, $referenceCounter);
     }
 
-    public function createObservableUpDownCounter(string $name, ?string $unit = null, ?string $description = null, callable ...$callbacks): ObservableUpDownCounterInterface
+    public function createObservableUpDownCounter(string $name, ?string $unit = null, ?string $description = null, $advisory = [], callable ...$callbacks): ObservableUpDownCounterInterface
     {
+        if (is_callable($advisory)) {
```
brettmc, line 165: this is for BC, right? If so, is it worth documenting that it could be removed in a future breaking version?

Nevay: IMO we should wait with marking this for future removal until we actually plan to break BC / prepare for an API 2.0 release that changes the type to `array $advisory` (which I assume won't happen in the near future). Removal in the SDK can then happen in a minor version.

brettmc (approved):

PR #1210, opened 2024-01-12
optimize resource creation

brettmc:
Profiling the SDK autoloader startup with xdebug, I could see that we re-detect resources multiple times which adds overhead.

- instead of creating new resources from various factories, create them once from sdk autoloader and
pass them to the factories (most of which already supported this).
- cache header from OtlpUtil, to save a couple of constructor calls to Sdk detector
- check env before php.ini for config, since this is the more popular approach and saves some cycles
- note as a todo that we could save some further cycles around resource merging by not re-validating attributes

### DIFF
--- a/src/SDK/Resource/ResourceInfo.php
+++ b/src/SDK/Resource/ResourceInfo.php
@@ -50,7 +50,7 @@ class ResourceInfo
         $copyOfAttributesAsArray = array_slice($this->attributes->toArray(), 0); //This may be overly cautious (in trying to avoid mutating the source array)
         ksort($copyOfAttributesAsArray); //sort the associative array by keys since the serializer will consider equal arrays different otherwise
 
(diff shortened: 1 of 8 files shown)


bobstrecansky (approved):

src/SDK/Resource/ResourceInfo.php
```diff
@@ -62,6 +62,7 @@ public function serialize(): string
      * resource, the value of the updating resource MUST be picked (even if the updated value is empty)
      *
      * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.20.0/specification/resource/sdk.md#merge
+     * @todo can we optimize this to avoid re-validating the attributes on merge?
```
bobstrecansky, line 65: Created an issue for this: https://github.com/open-telemetry/opentelemetry-php/issues/1212

PR #1216, opened 2024-01-17
Fix proto 1.1.0 link trace flags

Nevay:


src/Contrib/composer.json
```diff
@@ -15,7 +15,7 @@
         "ext-json": "*",
         "open-telemetry/api": "^1.0",
         "open-telemetry/context": "^1.0",
-        "open-telemetry/gen-otlp-protobuf": "^1.0",
+        "open-telemetry/gen-otlp-protobuf": "^1.1",
```
Nevay, line 18: Can we remove this `composer.json`? Doesn't seem to be used https://github.com/open-telemetry/opentelemetry-php/blob/main/.gitsplit.yml.

brettmc: I think we can - it was once used for https://packagist.org/packages/open-telemetry/sdk-contrib but that's now abandoned.

brettmc (approved):

PR #1226, opened 2024-02-01
adding OTEL_PHP_EXCLUDED_URLS

brettmc:
Modeled on python's feature to turn off all instrumentation for specific URLs, https://opentelemetry.io/docs/languages/python/automatic/agent-config/#excluded-urls this feature will effectively disable SDK autoloading if there is a request URI that matches one of the provided excluded URLs, and no-op implementations will be used for tracer/logger/meter providers.


brettmc:
The primary use-case for this would be for shared-nothing PHP setups (apache, nginx) to not generate 1000s of traces for URLs such as `/healthcheck`...

bobstrecansky (approved):

PR #1237, opened 2024-02-23
feat: Allow OTEL_PHP_DEBUG_SCOPES_DISABLED as en environment variable

PROFeNoM:
Hi :wave:

While doing some tests and trying to set `OTEL_PHP_DEBUG_SCOPES_DISABLED` to `1` as an **environment variable** to disable the use of `DebugScope`, it wouldn't work. This is because it wasn't considered.

I guess it could have also made sense to use some kind of a `ResolverInterface`, considering both INIs and env var.

Also, my personal opinion on this env var is that it could make sense to have it enabled by default (in the next major, being a breaking change), considering that the use of `debug_backtrace` has a non-negligible performance impact.

### DIFF
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -7,7 +7,6 @@ namespace OpenTelemetry\Context;
 use function assert;
 use const FILTER_VALIDATE_BOOLEAN;
 use function filter_var;
-use function ini_get;
 use function spl_object_id;
 
 /**
@@ -15,6 +14,8 @@ use function spl_object_id;
  */
 final class Context implements ContextInterface
 {
+    private const OTEL_PHP_DEBUG_SCOPES_DISABLED = 'OTEL_PHP_DEBUG_SCOPES_DISABLED';
+
     /** @var ContextStorageInterface&ExecutionContextAwareInterface */
     private static ContextStorageInterface $storage;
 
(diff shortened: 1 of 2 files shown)


Nevay:
Disabling via environment variable should work as expected (unless `variables_order` is configured to not populate `$_SERVER`):
```shell
OTEL_PHP_DEBUG_SCOPES_DISABLED=1 php -r 'require "vendor/autoload.php"; echo OpenTelemetry\Context\Context::getCurrent()->activate()::class, PHP_EOL;'
```

---

Assuming that you are using `putenv()` to set the environment variable as done in the testcase - `putenv()` is not threadsafe, for example `vlucas/phpdotenv` and `symfony/dotenv` won't populate it by default:
> Using getenv() and putenv() is strongly discouraged due to the fact that these functions are not thread safe, however it is still possible to instruct PHP dotenv to use these functions.

> Beware that `putenv()` is not thread safe, that's why it's not enabled by default
> // don't check existence with getenv() because of thread safety issues

If we want to support `putenv()` / `getenv()`, we should IMO use the order `$_SERVER` > `getenv()` > `ini_get()`.

brettmc (changes requested):
To be consistent with how the SDK checks for variables, can you change the order to be $_SERVER, getenv, ini_get

src/Context/Context.php
```diff
@@ -91,9 +91,15 @@ public function activate(): ScopeInterface
 
     private static function debugScopesDisabled(): bool
     {
-        $disabled = $_SERVER['OTEL_PHP_DEBUG_SCOPES_DISABLED'] ?? ini_get('OTEL_PHP_DEBUG_SCOPES_DISABLED');
+        if (isset($_SERVER['OTEL_PHP_DEBUG_SCOPES_DISABLED']) && filter_var($_SERVER['OTEL_PHP_DEBUG_SCOPES_DISABLED'], FILTER_VALIDATE_BOOLEAN)) {
```
brettmc, line 94: since the `OTEL_PHP_DEBUG_SCOPES_DISABLED` string is used a lot, can you move it into a class constant?

tests/Unit/Context/ContextTest.php
```diff
@@ -177,4 +178,19 @@ public function test_attach_and_detach_set_current_ctx(): void
             $scope->detach();
         }
     }
+
+    public function test_debug_scopes_disabled_env_var(): void
+    {
+        \putenv('OTEL_PHP_DEBUG_SCOPES_DISABLED=1');
```
brettmc, line 184: I'm worried this might interfere with other tests, since it modifies the environment and doesn't reset it.

PROFeNoM: My bad; it would have interfered with other tests!

src/Context/Context.php
```diff
@@ -91,9 +92,15 @@ public function activate(): ScopeInterface
 
     private static function debugScopesDisabled(): bool
     {
-        $disabled = $_SERVER['OTEL_PHP_DEBUG_SCOPES_DISABLED'] ?? ini_get('OTEL_PHP_DEBUG_SCOPES_DISABLED');
+        if (isset($_SERVER[self::OTEL_PHP_DEBUG_SCOPES_DISABLED]) && filter_var($_SERVER[self::OTEL_PHP_DEBUG_SCOPES_DISABLED], FILTER_VALIDATE_BOOLEAN)) {
+            return true;
```
brettmc, line 96: I think that this means that if `OTEL_PHP_DEBUG_SCOPES_DISABLED` is `false` in any source, it will then look to the next entries. But, if it's actually set as false (as opposed to not set at all), the user presumably means they don't want it disabled?

brettmc:
> Also, my personal opinion on this env var is that it could make sense to have it enabled by default (in the next major, being a breaking change), considering that the use of debug_backtrace has a non-negligible performance impact.

I think the idea is that because it's inside an assertion, when run with a [production configuration](https://github.com/php/php-src/blob/master/php.ini-production#L1527) that code will not be run. So in effect, you couldn't use DebugScope in a non-assertion environment even if you wanted to.

brettmc:
Can you add a test to prove what happens if the variable is falsey? I'm assuming that disabled=false === enabled, so there DebugScope should be used in this case.

brettmc (approved):

Issue #1239, opened 2024-02-26
URL encoded values for `OTEL_EXPORTER_OTLP_HEADERS` are not supported


pyohannes:
**Describe your environment** 
```
PHP 8.1.2-1ubuntu2.14 (cli) (built: Aug 18 2023 11:41:11) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.2, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.2-1ubuntu2.14, Copyright (c), by Zend Technologies
```

**Steps to reproduce**
```sh
$ export OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic%20NzAx[...]"
$ php examples/traces/exporters/otlp_http_from_factory.php                                                                                                                    
OTLP http/protobuf example complete!
OpenTelemetry: [error] Export failure [exception] Unauthorized [previous]
#0 /home/johannes/Source/opentelemetry-php/src/Contrib/Otlp/SpanExporter.php(41): OpenTelemetry\SDK\Common\Export\Http\PsrTransport->send()
#1 /home/johannes/Source/opentelemetry-php/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php(255): OpenTelemetry\Contrib\Otlp\SpanExporter->export()
#2 /home/johannes/Source/opentelemetry-php/src/SDK/Trace/SpanProcessor/BatchSpanProcessor.php(198): OpenTelemetry\SDK\Trace\SpanProcessor\BatchSpanProcessor->flush()
#3 /home/johannes/Source/opentelemetry-php/src/SDK/Trace/TracerSharedState.php(98): OpenTelemetry\SDK\Trace\SpanProcessor\BatchSpanProcessor->shutdown()
#4 /home/johannes/Source/opentelemetry-php/src/SDK/Trace/TracerProvider.php(92): OpenTelemetry\SDK\Trace\TracerSharedState->shutdown()
#5 /home/johannes/Source/opentelemetry-php/examples/traces/exporters/otlp_http_from_factory.php(21): OpenTelemetry\SDK\Trace\TracerProvider->shutdown()
#6 {main}
$ export OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic NzAx[...]"
$ php examples/traces/exporters/otlp_http_from_factory.php 

OTLP http/protobuf example complete!  
$ 
```
**What is the expected behavior?**

[...]

PR #1242, opened 2024-02-28
support url-encoded OTEL_EXPORTER_OTLP_HEADERS values

brettmc:
Per a recent spec clarification in https://github.com/open-telemetry/opentelemetry-specification/issues/3832 OTLP headers should support url-encoded header values.

Fixes: #1239

### DIFF
--- a/src/Contrib/Otlp/OtlpUtil.php
+++ b/src/Contrib/Otlp/OtlpUtil.php
@@ -44,7 +44,7 @@ class OtlpUtil
             Configuration::getMap(Variables::OTEL_EXPORTER_OTLP_HEADERS);
         $headers += self::getUserAgentHeader();
 
-        return array_map('urldecode', $headers);
+        return array_map('rawurldecode', $headers);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


bobstrecansky (approved):

src/Contrib/Otlp/OtlpUtil.php
```diff
@@ -30,6 +37,16 @@ public static function method(string $signal): string
...
     }
 
+    public static function getHeaders(string $signal): array
+    {
+        $headers = Configuration::has(self::HEADER_VARS[$signal]) ?
+            Configuration::getMap(self::HEADER_VARS[$signal]) :
+            Configuration::getMap(Variables::OTEL_EXPORTER_OTLP_HEADERS);
+        $headers += self::getUserAgentHeader();
+
+        return array_map('urldecode', $headers);
```
Nevay, line 47:
```suggestion
        return array_map('rawurldecode', $headers);
```

https://www.w3.org/TR/baggage/#value specifies RFC3986 encoding.

Issue #1243, opened 2024-02-28
drop PHP 7.4 support


brettmc:
In line with our [versioning policy](https://github.com/open-telemetry/opentelemetry-php?tab=readme-ov-file#versioning), PHP 7.4 is overdue to be de-supported. This has been discussed at SIG a couple of times over the past few months.

PR #1244, opened 2024-02-29
Drop PHP 7.4 support

brettmc:
* update composer.json minimums, and run rector over the code to update what it can. changed some static analysis config to fix some rector-induced failures.
* adding some union type-hints which rector didn't, and remove the corresponding `@param` phpdoc
* update typehint for metric aggregation, and note a todo for 8.1+ to convert to an enum

To do:
* [x] fix `TypeError: WeakMap key must be an object` test failures with batch exporter self-diagnostics
* [x] remove 7.4 from opentelemetry-php/gh-workflows/.github/workflows/validate-packages.yml

Fixes: #1243


brettmc:
The `WeakMap` issue I was hitting is actually a [bug](https://github.com/php/php-src/pull/8995) fixed in 8.0.22, 8.1.9 and 8.2.0 - found and fixed by otel's own Nevay.

agoallikmaa (approved):
I suppose `symfony/polyfill-php80` dependency can also be removed now?

src/API/Baggage/Entry.php
```diff
@@ -6,21 +6,10 @@
...
-
-    /**
-     * @param mixed $value
-     * @param MetadataInterface $metadata
-     */
     public function __construct(
-        $value,
-        MetadataInterface $metadata
+        private mixed $value,
+        private MetadataInterface $metadata
```
Nevay, line 11:
```suggestion
        private MetadataInterface $metadata,
```

Should add trailing comma to changed constructors to reduce diff if we add more parameters.

src/Context/Propagation/MultiTextMapPropagator.php
```diff
@@ -32,10 +25,12 @@ final class MultiTextMapPropagator implements TextMapPropagatorInterface
...
      * @param list<TextMapPropagatorInterface> $propagators
      */
-    public function __construct(array $propagators)
-    {
-        $this->propagators = $propagators;
-        $this->fields = $this->extractFields($propagators);
+    public function __construct(/**
+     * @readonly
+     */
+    private array $propagators
```
Nevay, line 31:
```suggestion
    public function __construct(
        /** @readonly */
        private array $propagators,
```

src/Contrib/Otlp/ProtobufSerializer.php
```diff
@@ -29,100 +28,69 @@
...
-            case self::JSON:
-            case self::NDJSON:
-                return base64_decode(bin2hex($traceId));
-            default:
-                throw new AssertionError();
-        }
+        return match ($this->contentType) {
+            ContentTypes::PROTOBUF => $traceId,
+            ContentTypes::JSON, ContentTypes::NDJSON => base64_decode(bin2hex($traceId)),
+            default => throw new AssertionError(),
```
Nevay, line 60:
```suggestion
```

Same for the matches below.

Nevay (commented):
Some more methods where union types can be added:
- `SpanBuilderInterface::setParent()` and `::setAttribute()`
- `MeterInterface` `$advisory` parameter
- `ObserverInterface::observe()`
- `LogRecord::setAttribute()`

We can use `WeakMap` directly now and can remove:
- `OpenTelemetry\API\Instrumentation\CachedInstrumentation::createWeakMap()`
- `OpenTelemetry\SDK\Common\Util\WeakMap`

src/API/Baggage/Entry.php
brettmc: added a php-cs-fixer rule to apply this everywhere

src/Context/ContextKey.php
```diff
@@ -9,11 +9,8 @@
  */
 final class ContextKey implements ContextKeyInterface
 {
-    private ?string $name;
-
-    public function __construct(?string $name=null)
+    public function __construct(private ?string $name=null)
```
ChrisLightfootWild, line 12:
```suggestion
    public function __construct(private ?string $name = null)
```

ChrisLightfootWild (approved):
Nice :broom:  :muscle:

bobstrecansky (approved):

Issue #1245, opened 2024-02-29
Move SDK configuration into own package


brettmc:
Before opening a feature request against this repo, consider whether the feature should/could be implemented in the [other OpenTelemetry client libraries](https://github.com/open-telemetry). If so, please [open an issue in opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/new) first.

**Is your feature request related to a problem?**
Per [otel spec](https://opentelemetry.io/docs/specs/otel/overview/#sdk), most components should only depend on the API, and not the SDK. Because all of our configuration-management code is in the SDK, it's not usable by contrib modules etc which need to deal with config. So, we end up with duplicate code and inconsistent handling of config.

**Describe the solution you'd like**
If we break config related code (`OpenTelemetry\SDK\Common\Configuration\*`) into its own small package that other packages can use, then they can use config without requiring the SDK.

**Describe alternatives you've considered**
I once thought about creating a configuration-resolver interface, and writing a minimal config resolver in the API (matching the interface of the SDK version). The implementation was very similar to the SDK one. But, now that we also have config management in the `Context` package, that would create a circular dependency between API and Context.

brettmc:
Closing, replaced by #1304 which is an improved (and draft spec-compliant) way to configure all of the things.

PR #1248, opened 2024-03-05
split config into own package

brettmc:
Moving configuration code out of the SDK means that it can be used by components that should not depend on the SDK.
- move config into own package
- update all core components which were self-managing config to use the config package
- add some tests for config package to get coverage to 100%
- add missing type-hints to config package
- add missing config variables (OTEL_PHP_DEBUG_SCOPES_DISABLED and OTEL_PHP_FIBERS_ENABLED)

Prior to merge:
- [ ] create gitsplit target repo
- [ ] document that some components which formerly allowed truthy/falsey values will (or soon will) only accept `true` or `false`

Closes: #1245


bobstrecansky (approved):

Nevay:
IMO we shouldn't move everything out of the SDK, I would keep `Variables` (and `Defaults` + `KnownValues` if needed) in the component that uses them.

`Configuration::validateVariableType()` + `ValueTypes` + `VariableTypes` can be removed. They are only used to verify that code calls the correct method; no need to check this during runtime / should be verified by tests.

> Because all of our configuration-management code is in the SDK, it's not usable by contrib modules etc which need to deal with config

We should also take file configuration and configuration via code into account. The configuration formats do not match 1:1 -> an intermediate instrumentation configuration model would make sense instead of using `Configuration::getXXX()` directly. This would be especially useful for complex configuration options that cannot be expressed as environment variable, e.g. configuration for a SQL sanitizer.

Example: HTTP instrumentation needs to know which request and response headers should be recorded as attributes. Env/ini configuration currently uses `otel.instrumentation.http.request_headers` and `otel.instrumentation.http.response_headers`, while file configuration might use:
```yaml
instrumentation:
  config:
  - http_client:
      request_headers: []
      response_headers: []
```
Both formats could then populate the following intermediate model that is made available via a registry `InstrumentationConfigurationRegistry::get(HttpClientInstrumentation::class)`.
```php
class HttpClientInstrumentation implements InstrumentationConfiguration {
    public function __construct(
        public readonly array $requestHeaders = [],
        public readonly array $responseHeaders = [],
    ) {}
}
```

brettmc:
@Nevay

> Configuration::validateVariableType() + ValueTypes + VariableTypes can be removed

Removed.

> IMO we shouldn't move everything out of the SDK, I would keep Variables (and Defaults + KnownValues if needed) in the component that uses them.

I'm hesitant to do this, because then contrib packages need to depend on SDK to fetch config (unless we start splitting config up into `SDK`, `general`, and possibly also for any packages that have custom config)?

> Both formats could then populate the following intermediate model that is made available via a registry

I started playing around with this. I thought perhaps factories could accept an `InstrumentationConfiguration`. It felt like double-handling if a ComponentProvider created an intermediate configuration, and as it stands the various factories build their own config as they go. Maybe this is worth a follow-up PR?

Nevay:
> I'm hesitant to do this, because then contrib packages need to depend on SDK to fetch config (unless we start splitting config up into SDK, general, and possibly also for any packages that have custom config)?

`Variables` & co. contain only constants, not the code to actually fetch config.

> I started playing around with this. I thought perhaps factories could accept an InstrumentationConfiguration. It felt like double-handling if a ComponentProvider created an intermediate configuration, and as it stands the various factories build their own config as they go. Maybe this is worth a follow-up PR?

I've created a prototype that utilizes the idea of https://github.com/open-telemetry/opentelemetry-php/pull/1274 to inject configuration into instrumentation: https://github.com/Nevay/opentelemetry-php/commit/2ecf4b22c0c8d76f07d6238cd75c27167d396b15

brettmc:
Closing, replaced by #1304

PR #1249, opened 2024-03-05
File configuration

Nevay:
Implements an extensible file configuration format to support [opentelemetry-configuration](https://github.com/open-telemetry/opentelemetry-configuration). It uses `symfony/config` for creation and validation of the configuration schema.

New components can be added by implementing `ComponentProvider` and `ServiceLoader::register()`ing them. An example that provides additional samplers can be seen at https://github.com/Nevay/otel-sdk-contrib-sampler (not compatible with this SDK).

---

Opening this now as  #1248 should take future file configuration support into account.

### DIFF
--- a/.gitsplit.yml
+++ b/.gitsplit.yml
@@ -16,6 +16,8 @@ splits:
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/api.git"
   - prefix: "src/SDK"
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/sdk.git"
+  - prefix: "src/Config/SDK"
+    target: "https://${GH_TOKEN}@github.com/opentelemetry-php/config-sdk.git"
   - prefix: "src/Contrib/Otlp"
     target: "https://${GH_TOKEN}@github.com/opentelemetry-php/exporter-otlp.git"
   - prefix: "src/Contrib/Grpc"
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -319,6 +319,7 @@ return [
     'exclude_analysis_directory_list' => [
         'vendor/',
         'proto/',
+        'src/Config/SDK',
     ],
 
     // Enable this to enable checks of require/include statements referring to valid paths.
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "require": {
         "php": "^8.1",
         "google/protobuf": "^3.22",
-        "open-telemetry/sdk-configuration": "^1.0-dev",
+        "open-telemetry/sdk-configuration": "^0.1-dev",
         "php-http/discovery": "^1.14",
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,6 @@
     "require": {
         "php": "^8.1",
         "google/protobuf": "^3.22",
-        "open-telemetry/sdk-configuration": "@dev",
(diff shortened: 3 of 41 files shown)


brettmc:
I really like this. Could you add an example of how a component could be configured using environment variables? If we can do that, then it looks like this could completely replace the existing SDK configuration code.

I understand that this is a draft, but does it make sense for the various component providers to live alongside the component and be registered (via `spi`) in the appropriate composer.json, rather than globally registering everything?

P.S. I'm proposing at SIG this week to soon drop 8.0 support, given its [very low usage](https://packagist.org/packages/open-telemetry/api/php-stats) and being out of official support. That might make some further improvements possible (and I notice that `spi` and `otel-sdk-configuration` require `^8.1`)

Nevay:
I added a configuration example that uses env substitution. We cannot fully replicate the environment variable configuration:
- variables with list/map value cannot be used, injecting YAML structures is disallowed by spec, e.g. `OTEL_PROPAGATORS`, `OTEL_RESOURCE_ATTRIBUTES`
- variables that select a specific implementation do not map to the configuration format, e.g. `OTEL_TRACES_SAMPLER`, `OTEL_{signal}_EXPORTER`
- variables with fallback behavior are not supported by the environment variable substitution definition, e.g. `OTEL_EXPORTER_OTLP_{signal}_*`

> does it make sense for the various component providers to live alongside the component and be registered (via spi) in the appropriate composer.json, rather than globally registering everything?

I'm leaning towards yes; for extensions that are not defined by `opentelemetry-configuration` definitely. I mainly kept it all in a separate package for now due to the PHP8.1 requirement.

FWIW I have been using an `Env\Loader` interface for env configuration which could replace the current `Registry` + `SpanExporterFactory`/`MetricExporterFactory` etc. if we want to use the same registration mechanism for both formats.
```php
interface Loader {

    public function load(EnvResolver $env, LoaderRegistry $registry, Context $context): mixed;

    public function name(): string;
}
```

brettmc:
I think we can merge this. Let's annotate it as `experimental`, and then I'll try to apply my config PR over the top.

Nevay:
We should wait until #1256 is merged, I think most workflows will fail otherwise.

> could the important parts of otel-sdk-configuration come into the opentelemetry org?

Yes (but it should stay independent from the SDK); should it be added to this repository?

brettmc:
> Yes (but it should stay independent from the SDK); should it be added to this repository?

I think it should be in this repository, it doesn't feel right for a core component to depend on a contrib one. I think it can either be bundled into the `Config` directory, or a new one if you feel it doesn't belong in config.

#1256 has been merged, now.

Nevay:
> Could you add an example of how a component could be configured using environment variables?

https://github.com/open-telemetry/opentelemetry-configuration/pull/76 provides a starter template that contains all env vars that map cleanly to file config.

> are you still planning on working on this PR?

Yes; but not sure when I will find time to move `otel-sdk-configuration` to this repository.
Currently the `packages:validate:installation` check is failing: we can either copy all data from the package `composer.json` to the root `composer.json` and remove the `path` repository or try to fix the validation check. There are some TODOs that require SDK changes -> we should likely disable these configuration options for now. Besides that this PR should be ready for review.

brettmc:
> Currently the packages:validate:installation check is failing

@Nevay submitted a PR against your branch which makes this command happy (per your first suggestion)

brettmc (approved):
LGTM, I'm happy to merge it as-is. I'll create an issue to port otel-sdk-configuration in.

PR #1253, opened 2024-03-08
Fix conversion of non-utf8 sequences to `AnyValue`

Nevay:
[Converting to AnyValue](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-type-mapping.md#string-values):
> String values which are not valid Unicode sequences SHOULD be converted to AnyValue's bytes_value with the bytes representing the string in the original order and format of the source string.

### DIFF
--- a/src/Contrib/Otlp/AttributesConverter.php
+++ b/src/Contrib/Otlp/AttributesConverter.php
@@ -46,10 +46,21 @@ final class AttributesConverter
             $result->setDoubleValue($value);
         }
         if (is_string($value)) {
-            $result->setStringValue($value);
+            if (self::isUtf8($value)) {
+                $result->setStringValue($value);
+            } else {
+                $result->setBytesValue($value);
+            }
         }
 
         return $result;
+    }
+
+    private static function isUtf8(string $value): bool
+    {
+        return \extension_loaded('mbstring')
+            ? \mb_check_encoding($value, 'UTF-8')
+            : (bool) \preg_match('//u', $value);
     }
 
     /**
--- a/tests/Unit/Contrib/Otlp/AttributesConverterTest.php
+++ b/tests/Unit/Contrib/Otlp/AttributesConverterTest.php
@@ -36,6 +36,13 @@ class AttributesConverterTest extends TestCase
         ];
     }
 
+    public function test_convert_bytes(): void
+    {
+        $anyValue = AttributesConverter::convertAnyValue("\xe2");
+        $this->assertTrue($anyValue->hasBytesValue());
+        $this->assertSame("\xe2", $anyValue->getBytesValue());
+    }
+
     /**
      * @dataProvider arrayProvider
      */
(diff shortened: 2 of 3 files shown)


weslenteche (approved):

src/Contrib/Otlp/AttributesConverter.php
```diff
@@ -46,12 +46,23 @@ public static function convertAnyValue($value): AnyValue
...
+                $result->setBytesValue($value);
+            }
         }
 
         return $result;
     }
 
+    private static function isUtf8(string $value): bool
+    {
+        return \extension_loaded('mbstring')
```
agoallikmaa, line 61: Technically this condition might not be necessary since `symfony/polyfill-mbstring` is pulled in transitively from the SDK. Not sure if it's best practice to rely on that though.

agoallikmaa (approved):

brettmc (approved):

bobstrecansky (approved):

src/Contrib/Otlp/AttributesConverter.php
Nevay: We should leave it as-is. The polyfill is slower and exports can contain a large number of attributes.
Creating and exporting spans with 10 string attributes (16 bytes ea.; 8 valid and 2 invalid utf8 sequences) to local collector w/ default batch size was around 10% slower with the polyfill compared to the `preg_match()` fallback (fallback was incorrect until latest push).


Benchmark results for creating `AnyValue` from a 16 bytes string (polyfill is esp. costly for invalid utf8 sequences due to [its `iconv` fallback](https://github.com/symfony/polyfill-mbstring/blob/v1.29.0/Mbstring.php#L427)):
```
+------------------------------------+---------+---------+--------+---------+
| subject                            | memory  | mode    | rstdev | stdev   |
+------------------------------------+---------+---------+--------+---------+
| benchCheckEncoding (valid)         | 1.959mb | 0.245μs | ±1.95% | 0.005μs |
| benchCheckEncoding (invalid-begin) | 1.959mb | 0.248μs | ±1.68% | 0.004μs |
| benchCheckEncoding (invalid-end)   | 1.959mb | 0.250μs | ±1.94% | 0.005μs |
| benchPregMatch (valid)             | 1.959mb | 0.307μs | ±2.06% | 0.006μs |
| benchPregMatch (invalid-begin)     | 1.959mb | 0.262μs | ±1.53% | 0.004μs |
| benchPregMatch (invalid-end)       | 1.959mb | 0.270μs | ±1.62% | 0.004μs |
| benchIsUtf8 (valid)                | 1.959mb | 0.250μs | ±1.62% | 0.004μs |
| benchIsUtf8 (invalid-begin)        | 1.959mb | 0.252μs | ±1.64% | 0.004μs |
| benchIsUtf8 (invalid-end)          | 1.959mb | 0.253μs | ±1.31% | 0.003μs |
| benchPolyfill (valid)              | 1.959mb | 0.383μs | ±1.67% | 0.006μs |
| benchPolyfill (invalid-begin)      | 1.959mb | 0.830μs | ±1.51% | 0.013μs |
| benchPolyfill (invalid-end)        | 1.959mb | 0.945μs | ±1.51% | 0.014μs |
+------------------------------------+---------+---------+--------+---------+
```

PR #1256, opened 2024-03-13
drop php 8.0 support

brettmc:
these are mostly automated changes from rector, applying 8.1 syntax updates.
also, apply some manual type-hint updates and readonly properties that rector missed.

after this is merged, we can separate create PRs to:
- upgrade phpunit to v10
- replace some interfaces/classes with enums

### DIFF
--- a/src/SDK/Trace/Sampler/ParentBased.php
+++ b/src/SDK/Trace/Sampler/ParentBased.php
@@ -29,22 +29,31 @@ use OpenTelemetry\SDK\Trace\Span;
  */
 class ParentBased implements SamplerInterface
 {
+    private readonly SamplerInterface $remoteParentSampler;
+    private readonly SamplerInterface $remoteParentNotSampler;
+    private readonly SamplerInterface $localParentSampler;
+    private readonly SamplerInterface $localParentNotSampler;
+
     /**
      * ParentBased sampler delegates the sampling decision based on the parent context.
      *
      * @param SamplerInterface $root Sampler called for the span with no parent (root span).
-     * @param SamplerInterface $remoteParentSampler Sampler called for the span with the remote sampled parent. When null, `AlwaysOnSampler` is used.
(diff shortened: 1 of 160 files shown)


bobstrecansky (approved):

agoallikmaa (approved):
Should be able to drop `symfony/polyfill-php81` dependency.

src/SDK/Trace/Sampler/ParentBased.php
```diff
@@ -29,34 +29,22 @@
...
-     * @param SamplerInterface|null $remoteParentNotSampler Sampler called for the span with the remote not sampled parent. When null, `AlwaysOffSampler` is used.
-     * @param SamplerInterface|null $localParentSampler Sampler called for the span with local the sampled parent. When null, `AlwaysOnSampler` is used.
-     * @param SamplerInterface|null $localParentNotSampler Sampler called for the span with the local not sampled parent. When null, `AlwaysOffSampler` is used.
+     * @param SamplerInterface $remoteParentSampler Sampler called for the span with the remote sampled parent. When null, `AlwaysOnSampler` is used.
+     * @param SamplerInterface $remoteParentNotSampler Sampler called for the span with the remote not sampled parent. When null, `AlwaysOffSampler` is used.
+     * @param SamplerInterface $localParentSampler Sampler called for the span with local the sampled parent. When null, `AlwaysOnSampler` is used.
+     * @param SamplerInterface $localParentNotSampler Sampler called for the span with the local not sampled parent. When null, `AlwaysOffSampler` is used.
```
Nevay, line 39: Changing to non-nullable parameters is a breaking change.
If we still want to change it: should remove the `When null, ...` part from doc

brettmc: Reverted to the original code, and disabled a rector rule for this file.

PR #1262, opened 2024-03-21
Align internal no-op implementations into Noop namespaces.

ChrisLightfootWild:
I noticed the noop namespace for metrics (`\OpenTelemetry\API\Metrics\Noop`) and had seen the other signals did not follow the same pattern.

Some IDE magic moved these around quickly so I thought I'd open a PR.

These are marked as internal, so by my understanding we *could* theoretically move them without breaking our public API. Whether we *should* is a different question, so I leave that to feedback.


src/API/Trace/Noop/NonRecordingSpan.php
brettmc, line 1: I don't think this should be moved into `Noop`, since it's not directly related to noop traces (it is used by `Span` too)

brettmc:
> These are marked as internal

Many _aren't_ marked as internal (and somebody doing manual configuration might have cause to, for example, create a noop tracer provider). So, I think it's a BC break and we might need to plan for it (or, at least document it in the release notes)

ChrisLightfootWild:
> > These are marked as internal
> 
> Many _aren't_ marked as internal (and somebody doing manual configuration might have cause to, for example, create a noop tracer provider). So, I think it's a BC break and we might need to plan for it (or, at least document it in the release notes)

Given that is the case, it is probably best to close this one off for now. 👍

PR #1267, opened 2024-03-31
Fix tracestate implementation

Nevay:
Found some issues while running the `TraceState` implementation against my old test cases:
- member limit of 32 was only applied on `::parse()`, was not applied in `::with()`
- `::parse()` discarded tracestates with length > 512; [W3C spec](https://www.w3.org/TR/trace-context/#tracestate-limits) specifies this as lower limit, not upper limit:
  > Vendors SHOULD propagate at least 512 characters of a combined header.
- `::parse()` did not allow empty / whitespace-only members, [W3C spec](https://www.w3.org/TR/trace-context/#tracestate-header-field-values):
  > Empty and whitespace-only list members are allowed. Vendors MUST accept empty tracestate headers but SHOULD avoid sending them. Empty list members are allowed in tracestate because it is difficult for a vendor to recognize the empty value when multiple tracestate headers are sent.

This PR also changes the behavior for duplicate keys in `::parse()`; previously the last value was used, now the first (=newer) value is used; we might consider discarding the tracestate instead. `key=value,key2=value2,key=value3` resulted previously in `key=value3,key2=value2`, now results in `key=value,key2=value2`.


tests/Unit/API/Trace/TraceStateTest.php
```diff
@@ -210,4 +213,58 @@ public function test_validate_value(): void
...
+        foreach (array_reverse($entries) as $key => $value) {
+            $traceState = $traceState->with($key, $value);
+        }
+
+        $this->assertSame($expected, $traceState->toString(limit: $limit));
+    }
+
+    public static function toStringProvider(): array
+    {
+        return [
```
agoallikmaa, line 230: Perhaps add test for value ending with space here too (which would be rejected), as that doesn't seem to be covered by other tests.

tests/Unit/API/Trace/TraceStateTest.php
```diff
@@ -210,4 +213,58 @@ public function test_validate_value(): void
...
+            ['key=value', 'key=value'],
+            ['key=value,key2=value2', 'key=value,key2=value2'],
+            ['key=value,key2=value2,key=value3', 'key=value,key2=value2'],
+
+            ['', ''],
+            ['  ', ''],
+            ['key=value,,key2=value2', 'key=value,key2=value2'],
+            ['key=value,   ,key2=value2', 'key=value,key2=value2'],
+
+            ['0', null],
```
agoallikmaa, line 266: Doesn't seem like there is any reason to differentiate between `''` and `null` here, can change to `''` and remove nullability from test method.

agoallikmaa (approved):

tests/Unit/API/Trace/TraceStateTest.php
Nevay: Will add an assertion that a warning is triggered if expected is `null`.

tests/Unit/API/Trace/TraceStateTest.php
Nevay: Will readd https://github.com/open-telemetry/opentelemetry-php/pull/1267/commits/bace440ea13a17c0bd2f08da99f0e728a474f7d7#diff-48f55048bb7253924533e9078b9eda9994babffa4ef746e8a16a65d65a0dc319L162-L181, which includes a test for value with trailing space.

Issue #1111, opened 2023-09-04
$timout in PsrTransportFactory is not used


xosofox:
Check https://github.com/open-telemetry/opentelemetry-php/blob/73ff5adcb8f1db348bedb422de760e475df16841/src/SDK/Common/Export/Http/PsrTransportFactory.php#L41

The `create` function has a parameter `$timeout`.

However, this value is not used anymore further - thus rendering the expectation of passing a timeout to have any effect useless.

Is there a reason why `PsrTransport` does not have this paramter?

I'm willing to contribute to fix this - as my app is hanging terribly in case the metrics server is not available - however I would like to understand any considerations to be taken into account

Or am I missing something? Is there another way to pass a timout to my `ClientInterface`?

brettmc:
The `timeout` param is there because the spec says it should be: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.24.0/specification/performance.md#shutdown-and-explicit-flushing-could-block
And, it doesn't work because we code to the PSR http client/request interfaces, and I believe there is not a way to do something useful with that timeout value, either as part of the request or the client. Timeouts do work for some transports though (notably, gRPC).

The only way I'm aware of to set a timeout is to pre-configure your http client of choice before giving it to the PsrTransportFactory.

xosofox:
Thanks for getting back that quickly

I will try to investigate around Psr...

In the meantime, can you give me a pointer on how to pre-configure the Client when I'm useing this scenario:

        $transport = (new OtlpHttpTransportFactory())->create(.........);

How would that have to be expanded?

brettmc:
Today, that scenario won't work, and you'd need to instantiate a PsrTransport directly rather than use the factory.
With guzzle and symfony's client, it's an option in the Client constructor.

So, something like:
```php
$client = new HttpClient(['timeout' => 1.5]); //symfony
$client = new Client(['timeout' => 1.5]); //guzzle

$transport = new PsrTransport($client, /* other params */);
```
But I also note from symfony's docs that you can set `default_socket_timeout` and that will be used as the default http request timeout, which I think means you could use Psr18 discovery and hence the OtlpHttpTransportFactory.

All that said, I think that `OtlpHttpTransportFactory` could accept an http client as a constructor argument, and only use discovery as a fallback. I don't think that it helps your use-case though...

xosofox:
I worked around it by using the `PsrTransportFactory`directly, configuring my Client first.

[...]

PR #1275, opened 2024-04-12
enable timeout for http transports

brettmc:
Attempt to do our own discovery for some well-known PSR-18 clients, which allows for configuring timeout (and in future certificates, keys etc). If we can't create a client, fall back to PSR-18 discovery which was the previous behaviour.

Fixes: #1111


examples/traces/exporters/otlp_http_from_factory.php
```diff
@@ -11,6 +11,7 @@
  */
 putenv('OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4318');
 putenv('OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf');
+putenv('OTEL_EXPORTER_OTLP_TIMEOUT=1500'); //1.5s
```
bobstrecansky, line 14: Is 1.5s an arbitrary measurement?  Seems just as good as any; was just curious.

bobstrecansky (commented):
This feels like a good direction and a net positive to me.

examples/traces/exporters/otlp_http_from_factory.php
brettmc: This is just an example to show setting it to a non-default value does work. The official default in spec is 10s.

bobstrecansky (approved):

Issue #1280, opened 2024-04-18
Move ClockInterface / ClockFactory from SDK to API


ChrisLightfootWild:
**Is your feature request related to a problem?**
When dealing with timing within a contrib package, there's no other obvious mechanism to mock time other than using [SDK/Common/Time](https://github.com/open-telemetry/opentelemetry-php/tree/1.0.0beta2/src/SDK/Common/Time) components.

As an example, [this commit](https://github.com/open-telemetry/opentelemetry-php-contrib/pull/250/commits/a5a72620924a717d5da91f77a9106c05a3567b19) uses the ClockFactory from the SDK to allow for reliable testing of the functionality.

**Describe the solution you'd like**
Ideally, port the time components into the API package so that contrib would not have a hard dependency on the SDK. 

**Describe alternatives you've considered**
Adding a ClockFactory directly to the contrib package, but it is likely something that other packages would have a use for.

**Additional context**
I'd be happy to assist with porting this, if required.

PR #1281, opened 2024-04-19
migrate Clock classes to API

brettmc:
- move Clock* + Util into API
- add a deprecated wrapper for ClockFactory and Util
- remove unused StopWatch* classes

Fixes: #1280


src/SDK/Common/Time/ClockFactory.php
```diff
@@ -4,27 +4,16 @@
...
-    {
-        return new SystemClock();
-    }
+use OpenTelemetry\API\Common\Time\ClockInterface;
 
+/**
+ * @deprecated Use OpenTelemetry\API\Common\Time\ClockFactory
+ * @codeCoverageIgnore
+ */
+class ClockFactory
```
ChrisLightfootWild, line 13: :thinking: Should the old interface should remain too and also marked as deprecated?
```suggestion
class ClockFactory implements ClockFactoryInterface
```

then maybe something like:
```suggestion
interface ClockFactoryInterface extends \OpenTelemetry\API\Common\Time\ClockFactoryInterface
```

brettmc: Retained old SDK interfaces and Util class, deprecated.

ChrisLightfootWild (approved):
Amazing, thanks for this! :stopwatch:

weslenteche (approved):

agoallikmaa (approved):

src/API/Common/Time/Util.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Time;
+
+class Util
```
Nevay, line 7: Would prefer if we drop this class. It's currently only needed in `Zipkin\SpanConverter`. Not sure how useful the `nanosToXxx` methods are, some(/most?) use-cases might need rounded values instead of floored values.

If we want to keep the class: should probably add `nanosToSeconds` and `microsToNanos`.

src/API/Common/Time/ClockInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Common\Time;
+
+interface ClockInterface
+{
+    public const MILLIS_PER_SECOND = 1_000;
+    public const MICROS_PER_SECOND = 1_000_000;
```
Nevay, line 10: Do we need these two constants? Only conversion from/to nanos should be relevant.

src/API/Common/Time/Util.php
brettmc: dropped

src/API/Common/Time/ClockInterface.php
brettmc: we don't. removed.

src/API/Common/Time/ClockFactory.php
```diff
@@ -0,0 +1,30 @@
...
+final class ClockFactory implements ClockFactoryInterface
+{
+    private static ?ClockInterface $default = null;
+
+    public static function create(): self
+    {
+        return new self();
+    }
+
+    public function build(): ClockInterface
```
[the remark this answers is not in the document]
brettmc: removed ClockFactory + interface, and replaced the functionality with a `Clock` class.

bobstrecansky (approved):

src/SDK/Common/Time/ClockFactory.php
brettmc: I'm prepared to gamble on nobody having done that, and wear the BC break. There'll be a beta of the next version.

brettmc: I think that in the unlikely event that somebody has implemented their own clock factory, then it can get picked up when we release a beta. I did try to keep it around, but deptrac was unhappy and I didn't think it was worth the effort.

Nevay (approved):

PR #1291, opened 2024-04-26
implement events v1.32

brettmc:
implement the events api + sdk per spec v1.32:
- event logger is now only retrievable via an event logger provider
- domain attribute for events is removed
- events accept a subset of LogRecord fields, rather than an entire LogRecord

todo: the spec is a little unclear on exactly how an EventLoggerProvider should access the Logger that it uses. Since the only way a Logger should be obtained is through a LoggerProvider (at least, that's true for other signals, but not expressly forbidden by the [logs bridge API](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/bridge-api.md#loggerprovider)), I've gone with an EventLoggerProvider containing a LoggerProvider, from which it gets its logger.

The [event logger API](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/event-api.md#eventloggerprovider) suggests that the above is correct ("EventLoggers can be accessed with a EventLoggerProvider"), however the [event logger SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/event-sdk.md#overview) says that "The SDK MAY be implemented on top of the [Logs Bridge API](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/bridge-api.md)".


src/API/Logs/Severity.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Logs;
+
+class Severity
```
ChrisLightfootWild, line 7: Now we are in 8.1, do you have any thoughts on making something like this an enum? :thinking:

brettmc: great suggestion, now it's an enum. Currently, `LogRecord` accepts int|enum, but I'm certainly open to accepting just the enum.

src/API/Logs/EventLoggerProviderInterface.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Logs;
+
+/**
+ * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/event-api.md#get-an-eventlogger
+ */
+interface EventLoggerProviderInterface
+{
+    public function __construct(LoggerProviderInterface $loggerProvider);
```
Nevay, line 12:
```suggestion
```

src/API/Logs/EventLoggerInterface.php
```diff
@@ -4,10 +4,24 @@
...
+        ?int $timestamp = null,
+        ?ContextInterface $context = null,
+        Severity|int|null $severityNumber = null,
+        ?array $attributes = [],
+    ): void;
+
+    /**
+     * @deprecated Use `emit`
+     */
     public function logEvent(string $eventName, LogRecord $logRecord): void;
```
Nevay, line 26: IMO it doesn't make sense to keep this method, the definition of the `event.name` attribute was changed in SemConv 1.24 -> previous calls to `::logEvent()` might populate the attribute with an incorrect value:
> BREAKING: Change event.name definition to include namespace and remove event.domain from log event attributes.

src/SDK/Logs/EventLogger.php
```diff
@@ -0,0 +1,53 @@
...
+    ): void {
+        $logRecord = new LogRecord();
+        $attributes += ['event.name' => $name];
+        $logRecord->setAttributes($attributes);
+        $payload && $logRecord->setBody($payload);
+        $logRecord->setTimestamp($timestamp ?? (int) (microtime(true)*LogRecord::NANOS_PER_SECOND));
+        $context && $logRecord->setContext($context);
+        $logRecord->setSeverityNumber($severityNumber ?? Severity::INFO);
+
+        $this->logger->emit($logRecord);
```
Nevay, line 40: Should emit directly to `LogRecordProcessor` instead of `LoggerInterface`:
> The [Observed Timestamp](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-observedtimestamp) MUST not be set. (NOTE: [emit a logRecord](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#emit-a-logrecord) will set ObservedTimestamp to the current time when unset.)

src/API/Logs/Severity.php
ChrisLightfootWild: I can't see a concrete restriction on the  SeverityNumber (see: [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/data-model.md#field-severitynumber)), but we could potentially restrict to known values only `Severity::tryFrom($value)` etc.

In that case, maybe internally we could work with an enum throughout for this new functionality? :thought_balloon:

brettmc: Let's try it out. I've updated to only use an enum.

src/SDK/Logs/EventLogger.php
```diff
@@ -0,0 +1,42 @@
...
+        ?Severity $severityNumber = null,
+        ?array $attributes = [],
+    ): void {
+        $logRecord = new LogRecord();
+        $attributes += ['event.name' => $name];
+        $logRecord->setAttributes($attributes);
+        $payload && $logRecord->setBody($payload);
+        $logRecord->setTimestamp($timestamp ?? (int) (microtime(true)*LogRecord::NANOS_PER_SECOND));
+        $context && $logRecord->setContext($context);
+        $severityNumber && $logRecord->setSeverityNumber($severityNumber);
```
Nevay, line 38: Severity must default to `INFO` if not provided.
> If not provided, SeverityNumber MUST be set to SEVERITY_NUMBER_INFO=9.

src/API/Logs/LogRecord.php
```diff
@@ -42,9 +42,9 @@ public function setContext(?ContextInterface $context = null): self
     /**
      * @see https://opentelemetry.io/docs/reference/specification/logs/data-model/#field-severitynumber
      */
-    public function setSeverityNumber(int $severityNumber): self
+    public function setSeverityNumber(Severity $severityNumber): self
```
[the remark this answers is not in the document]
brettmc: Since we've removed `emitEvent(LogRecord)`, there shouldn't be any direct usage of this class outside of log appenders (the monolog one in contrib is the only one I'm aware of). So, I don't think this is a critical BC to maintain...?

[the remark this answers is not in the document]
Nevay: It has very likely low/no impact (as `LogRecord::setSeverityNumber(Psr::serverityNumber($...))` stays compatible); but it is also quite trivial to keep BC. IMO we should preserve BC for the low chance that someone is using the logs API w/ a non-PSR-3 logger.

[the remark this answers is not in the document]
brettmc: Fair enough, reverted to maintain BC. I also deprecated `Psr::severityNumber` in favour of moving that logic into the enum itself.

src/SDK/Logs/EventLogger.php
```diff
@@ -0,0 +1,42 @@
...
+    public function emit(
+        string $name,
+        mixed $payload = null,
+        ?int $timestamp = null,
+        ?ContextInterface $context = null,
+        ?Severity $severityNumber = null,
+        ?array $attributes = [],
+    ): void {
+        $logRecord = new LogRecord();
+        $attributes += ['event.name' => $name];
```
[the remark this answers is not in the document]
brettmc: done, and added a test for this.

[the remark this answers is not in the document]
Nevay: Seems like the `array $attributes = []` -> `iterable $attributes = []` change https://github.com/open-telemetry/opentelemetry-php/pull/1291#discussion_r1582091353 was reverted while implementing this. Can use the following to reserve space for the `event.name` attribute:
```php
$logRecord->setAttribute('event.name', $name);
$logRecord->setAttributes($attributes);
$logRecord->setAttribute('event.name', $name);
```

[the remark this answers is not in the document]
brettmc: Fixed. I actually deliberately changed it back because I couldn't find a way to make it work with iterable. But, your suggestion works.

examples/logs/exporters/otlp_grpc.php
```diff
@@ -30,19 +29,18 @@
...
-        ->setSeverityNumber(SeverityNumber::SEVERITY_NUMBER_INFO)
+$eventLogger->emit(
+    name: 'foo',
+    payload: ['foo' => 'bar', 'baz' => 'bat', 'msg' => 'hello world'],
+    severityNumber: Severity::INFO
 );
 
-$eventLogger->logEvent(
+$eventLogger->emit(
     'foo',
```
Nevay, line 42:
```suggestion
    'bar',
```

 `foo` is used on L36 with different payload schema.
> All events with the same `name` have `Payloads` that conform to the same schema

Nevay (approved):

bobstrecansky (approved):

PR #1293, opened 2024-04-27
fix deptrac config to allow http clients

brettmc:
This quietens some deptrac warnings about uncovered dependencies for recently-added http clients.


deptrac.yaml
```diff
@@ -77,6 +77,14 @@ deptrac:
           collectors:
               - type: className
                 regex: ^Composer\\*
+      -   name: HttpClients
+          collectors:
+            - type: className
+              value: ^Symfony\\*
```
ChrisLightfootWild, line 83: Could/should this be more specific? 💭 
```suggestion
              value: ^Symfony\\Component\\HttpClient\\*
```

ChrisLightfootWild (approved):

bobstrecansky (approved):

PR #1302, opened 2024-05-02
move test tools into vendor-bin

brettmc:
To reduce (hopefully, eliminate) the situations where our testing tools have conflicting dependencies, use bamarni/composer-bin-plugin to manage them as separate vendored dirs. Update Makefile and github actions to use the self-contained versions. Remove the dependencies from root composer.json
Rejigger dockerfile to add opcache, and some config required for latest version of grpc

Replace: #1300


tools/README.md
```diff
@@ -0,0 +1 @@
+This directory contains symlinks to binaries for some of our testing tools that are managed by https://github.com/bamarni/composer-bin-plugin
```
dkarlovi, line 1: IIRC you can have the plugin add the executable into your regular vendor/bin so you don't need to maintain these links manually.

brettmc: I saw that feature, but it's usage [is discouraged](https://github.com/bamarni/composer-bin-plugin?tab=readme-ov-file#bin-links-bin-links) so I didn't try it.

dkarlovi: Ah, got it.

Maybe it makes sense to just use the path from `vendor-bin/bin` directly? It's not that much more complex than the tools one and it seems like less "paperwork" with the tools. :)

brettmc:
> Maybe it makes sense to just use the path from vendor-bin/bin directly?

I don't see a `vendor-bin/bin`. It would be handy if there was one. I thought `vendor-bin/<package>/vendor/bin/<executable>` was a bit verbose - not something I'd want to remember if, for example, I want to `make bash` then `XDEBUG_MODE=debug phpunit /path/to/some/test --filter SpecificTest` - which I personally frequently do.

dkarlovi (approved):

bobstrecansky (approved):

PR #1304, opened 2024-05-06
auto-instrumentation registration

brettmc:
- porting Nevay's PoC registering auto-instrumentation packages via SPI with file-based config.
- extend to support auto-loading:
  - use the not-yet-released v0.3 of file-based configuration (which combines SDK and instrumentation config)
  - The existence of the `OTEL_EXPERIMENTAL_CONFIG_FILE` defines where the SDK config file resides (relative to CWD), and also acts as a trigger to autoload via SPI rather than our current env-based version
  - move single-use code (create hook manager + config parsing + actual registration) into classes
  - support general instrumentation config (which is described in the file-based config spec)
  - support map-based array-like config (which is used by instrumentation configuration)

todo:
- [x] SDK autoloader should use noop hook manager if opentelemetry extension not available


src/SDK/Sdk.php
```diff
@@ -22,6 +23,7 @@ public function __construct(
         private readonly LoggerProviderInterface $loggerProvider,
         private readonly EventLoggerProviderInterface $eventLoggerProvider,
         private readonly TextMapPropagatorInterface $propagator,
+        private readonly HookManager $hookManager,
```
brettmc, line 26: This might just be a temporary home for hook manager, I'm not convinced that it belongs in the SDK.

src/API/Globals.php
```diff
@@ -34,6 +35,7 @@ public function __construct(
         private readonly LoggerProviderInterface $loggerProvider,
         private readonly EventLoggerProviderInterface $eventLoggerProvider,
         private readonly TextMapPropagatorInterface $propagator,
+        private readonly HookManager $hookManager,
```
brettmc, line 38: hook manager may not need to be globally registered, since all auto-instrumentations are registered at the same time and we probably don't need to access the hook manager again.

src/API/Instrumentation/AutoInstrumentation/Instrumentation.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace OpenTelemetry\API\Instrumentation\AutoInstrumentation;
+
+use OpenTelemetry\Config\SDK\Configuration\Context;
+use OpenTelemetry\Context\ContextStorageInterface;
+
+interface Instrumentation
+{
+    /**
+     * @todo context is nullable in order to support autoloading (and retrieving lazy-loaded tracers), but auto and non-auto
```
brettmc, line 13: there's a race condition when autoloading/auto-registering an instrumentation vs loading it later (see the two examples, configure_instrumentation.php and configure_instrumentation_global.php). For the moment, this allows a context or will use Globals - we should choose one approach.

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
```diff
@@ -0,0 +1,74 @@
...
+namespace OpenTelemetry\API\Instrumentation\AutoInstrumentation;
+
+use function assert;
+use Closure;
+use function extension_loaded;
+use OpenTelemetry\Context\Context;
+use OpenTelemetry\Context\ContextKeyInterface;
+use ReflectionFunction;
+
+// #[Nevay\SPI\ServiceProviderDependency\ExtensionDependency('opentelemetry', '^1.0')]
```
brettmc, line 14: @Nevay this class doesn't exist in any released SPI version

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
```diff
@@ -0,0 +1,74 @@
...
+            return null;
+        }
+
+        $contextKey = $this->contextKey;
+        $reflection = new ReflectionFunction($closure);
+
+        // TODO Add an option flag to ext-opentelemetry `hook` that configures whether return values should be used?
+        if (!$reflection->getReturnType() || (string) $reflection->getReturnType() === 'void') {
+            return static function (mixed ...$args) use ($closure, $contextKey): void {
+                /*if (!Context::getCurrent()->get($contextKey)) {
```
brettmc, line 58: commented out to make it work, haven't investigated why that key is not in context. same story with post function.

src/SDK/Common/Configuration/Variables.php
```diff
@@ -140,4 +140,6 @@ interface Variables
     public const OTEL_PHP_INTERNAL_METRICS_ENABLED = 'OTEL_PHP_INTERNAL_METRICS_ENABLED'; //whether the SDK should emit its own metrics
     public const OTEL_PHP_DISABLED_INSTRUMENTATIONS = 'OTEL_PHP_DISABLED_INSTRUMENTATIONS';
     public const OTEL_PHP_EXCLUDED_URLS = 'OTEL_PHP_EXCLUDED_URLS';
+    public const OTEL_PHP_SDK_CONFIG_FILE = 'OTEL_PHP_SDK_CONFIG_FILE';
```
Nevay, line 143:
```suggestion
    public const OTEL_PHP_SDK_CONFIG_FILE = 'OTEL_EXPERIMENTAL_CONFIG_FILE';
```

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#file-configuration

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
brettmc: Sorry, so it it. My IDE was having trouble with it.

src/SDK/Common/Configuration/Variables.php
brettmc: Done, and I've set `sdk-config.yaml` as the default, which the spec encourages.

src/SDK/Sdk.php
brettmc: removed

src/API/Globals.php
brettmc: removed from sdk and globals

examples/instrumentation/configure_instrumentation_global.php
```diff
@@ -0,0 +1,20 @@
...
+
+use OpenTelemetry\Example\Example;
+use const PHP_EOL;
+
+/**
+ * This example uses SPI (see root composer.json extra.spi) to configure an example auto-instrumentation from a YAML file
+ */
+// EXAMPLE_INSTRUMENTATION_SPAN_NAME=test1234 php examples/instrumentation/configure_instrumentation.php
+putenv('OTEL_PHP_AUTOLOAD_ENABLED=true');
+putenv(sprintf('OTEL_EXPERIMENTAL_CONFIG_FILE=%s/%s', __DIR__, 'otel-sdk.yaml'));
```
Nevay, line 15: Might make sense to add `getcwd()` as path to the configuration file locator to allow loading relative paths https://github.com/open-telemetry/opentelemetry-php/blob/500f6ee1bd32a2fd289125aecd0d38a123c57596/src/Config/SDK/Configuration/ConfigurationFactory.php#L101

brettmc: Agreed. Done.

src/API/Instrumentation/AutoInstrumentation/Instrumentation.php
brettmc: I've gone with using globals, as we currently do. Autoloading happens too early for providers to be available to pass in via context.

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
brettmc: I think this is fixed now (using context + activate, coupled with a shutdown handler to detach)

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
```diff
@@ -0,0 +1,78 @@
...
+        if (!$closure) {
+            return null;
+        }
+
+        $contextKey = $this->contextKey;
+        $reflection = new ReflectionFunction($closure);
+
+        // TODO Add an option flag to ext-opentelemetry `hook` that configures whether return values should be used?
+        if (!$reflection->getReturnType() || (string) $reflection->getReturnType() === 'void') {
+            return static function (mixed ...$args) use ($closure, $contextKey): void {
```
brettmc, line 61: question from SIG meeting: do we understand the difference between these two variants? does it affect a method's return type in unexpected ways?

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,107 @@ public static function autoload(): bool
...
-            return $configurator
-                ->withTracerProvider($tracerProvider)
-                ->withMeterProvider($meterProvider)
-                ->withLoggerProvider($loggerProvider)
-                ->withEventLoggerProvider($eventLoggerProvider)
-                ->withPropagator($propagator)
-            ;
-        });
+        if (Configuration::has(Variables::OTEL_EXPERIMENTAL_CONFIG_FILE)) {
+            $sdk = self::createAndRegisterFileBasedSdk();
```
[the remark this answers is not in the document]
brettmc: I originally did this, but I'm coming to the conclusion that:
- if we use SPI for everything (SDK, 3rd-party + contrib implementations of APIs, and auto-instrumentations), then they are all registered in `vendor/composer/GeneratedServiceProviderData.php` as part of composer install/update
- if all services are known about in advance, we no longer have a race condition because of the undefined order of composer's `autoload.files`, where an auto-instrumentation may be registered before the SDK or some other component
- so, we can be confident that when we register instrumentations, we have already completely set up TracerProvider & friends
- so, we don't need to lazy-load them, and can pass them straight in (like in https://github.com/Nevay/opentelemetry-php/commit/2ecf4b22c0c8d76f07d6238cd75c27167d396b15#diff-c524b1bb7970cf80c06c8356ca6d3d05fb905a2b23a46aa5733489822952b556R38 - which I've retained except for creating a new class in API because of dependencies)

TL;DR: using SPI means that Globals can be immediately registered.

WDYT?

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
```diff
@@ -0,0 +1,78 @@
...
+                    return;
+                }
+
+                $closure(...$args);
+            };
+        }
+
+        return static function (mixed ...$args) use ($closure, $contextKey): mixed {
+            if (!Context::getCurrent()->get($contextKey)) {
+                return null;
```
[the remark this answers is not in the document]
brettmc: It is a bug. Added a test and fix.

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
brettmc: I think this is resolved (and covered by tests) now.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,107 @@ public static function autoload(): bool
...
+            self::configureInstrumentation($sdk);
```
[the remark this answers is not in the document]
brettmc: yeah, makes sense (assuming that anything registered with SPI has default values or will otherwise successfully load without config). done.

src/SDK/SdkAutoloader.php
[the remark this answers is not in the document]
brettmc: I think my postscript above (coexisting with anything that might be using globals initializers) is what you meant here. So, I've added an initializer and a exercise adding an initializer after autoload has created an SDK, then accessing Globals.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,121 @@ public static function autoload(): bool
...
-            return false;
+        if (!Configuration::has(Variables::OTEL_PHP_INSTRUMENTATION_CONFIG_FILE)) {
+            return;
         }
-        $url = $_SERVER['REQUEST_URI'] ?? null;
-        if (!$url) {
-            return false;
+        $file = Configuration::getString(Variables::OTEL_PHP_INSTRUMENTATION_CONFIG_FILE);
+        $configuration = Instrumentation::parseFile($file)->create();
+        $storage = Context::storage();
```
Nevay, line 117: We cannot use `Globals::storage()` here. `initialize_fiber_handlers.php` as well as users of the swoole context storage might `Context::setStorage()` a different context storage during startup.

Either we drop the `$storage` parameter from `Instrumentation::register()` or we inject a context storage that delegates to the global storage.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,121 @@ public static function autoload(): bool
...
+     * @phan-suppress PhanUndeclaredClassMethod
      */
-    public static function isIgnoredUrl(): bool
+    private static function registerInstrumentations(?Sdk $sdk): void
     {
-        $ignoreUrls = Configuration::getList(Variables::OTEL_PHP_EXCLUDED_URLS, []);
-        if ($ignoreUrls === []) {
-            return false;
+        if (!Configuration::has(Variables::OTEL_PHP_INSTRUMENTATION_CONFIG_FILE)) {
+            return;
```
Nevay, line 113: IMO we should load instrumentations with empty configuration registry if the config file is not specified.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,121 @@ public static function autoload(): bool
...
+        $storage = Context::storage();
+        $hookManager = self::getHookManager();
+        $context = new InstrumentationContext(
+            $sdk?->getTracerProvider() ?? Globals::tracerProvider(),
+            $sdk?->getMeterProvider() ?? Globals::meterProvider(),
+            $sdk?->getLoggerProvider() ?? Globals::loggerProvider(),
+        );
+        foreach (ServiceLoader::load(\OpenTelemetry\API\Instrumentation\AutoInstrumentation\Instrumentation::class) as $instrumentation) {
+            /** @var \OpenTelemetry\API\Instrumentation\AutoInstrumentation\Instrumentation $instrumentation */
+            $instrumentation->register($hookManager, $context, $configuration, $storage);
```
Nevay, line 126: Should probably log errors instead of failing.

src/SDK/SdkAutoloader.php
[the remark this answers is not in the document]
Nevay: > TL;DR: using SPI means that Globals can be immediately registered.

Yes, if we remove `Registry::registerXxxFactory()` etc. we can replace `Globals::registerInitializer()` with a setter.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,121 @@ public static function autoload(): bool
...
-                ->withTracerProvider($tracerProvider)
-                ->withMeterProvider($meterProvider)
-                ->withLoggerProvider($loggerProvider)
-                ->withEventLoggerProvider($eventLoggerProvider)
-                ->withPropagator($propagator)
-            ;
-        });
+        $sdk = null;
+        if (Configuration::has(Variables::OTEL_EXPERIMENTAL_CONFIG_FILE)) {
+            $sdk = self::createSdkFromFileConfig();
```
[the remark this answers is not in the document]
brettmc: > `self::registerInstrumentation()` can use the providers from `Globals`

I think that's logically the same thing - if we register an initializer but then immediately reference `Globals`, then we haven't provided any opportunity for anything else that might run later from `autoload.files` to register anything else of interest - eg propagators.
So I think the discussion here is whether instrumentation packages in future should always get their providers from Globals. If we stick with Globals, which I have no issue with, then I'll just drop the `OpenTelemetry\API\Instrumentation\AutoInstrumentation\Context` class that was allowing providers to be passed in to an instrumentation's `register` method. That does have the benefit of having automatic and manual instrumentation get providers in the same way. Globals is the only way.

src/SDK/SdkAutoloader.php
```diff
@@ -30,80 +36,121 @@ public static function autoload(): bool
...
+    }
+
+    /**
+     * @phan-suppress PhanUndeclaredClassMethod
+     */
+    private static function getHookManager(): HookManager
+    {
+        /** @var HookManager $hookManager */
+        foreach (ServiceLoader::load(HookManager::class) as $hookManager) {
+            $scope = $hookManager->enable(Context::getCurrent())->activate();
```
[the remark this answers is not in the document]
brettmc: > We could add a method that allows changing hook managers from default disabled to default enabled which would make this scope obsolete.

I do think it would be preferable if the hook manager was enabled by default. If you've installed auto-instrumentation packages, you should get auto-instrumentation (which is the current behaviour). Plus, I don't want developers to need to write code to enable hooks to run; the most common usage I think would be to instrument.

src/SDK/SdkAutoloader.php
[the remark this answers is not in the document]
brettmc: Added a todo to `Registry` for a future enhancement, once we fully commit to SPI and get contrib converted...

src/SDK/SdkAutoloader.php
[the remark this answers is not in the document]
brettmc: I've switched `ExtensionHookManager` to being enabled by default. How one might access the hook manager to disable it isn't very clear if it was autoloaded - maybe it needs to be a singleton or registered in context?

src/SDK/SdkAutoloader.php
brettmc: Dropped storage from `Instrumentation::register()`

src/SDK/SdkAutoloader.php
brettmc: It _seems_ to work. `autoload.files` should not be an issue since everything continues to use globals initializers.
I've wrapped registering instrumentation in a try/catch/log, since some will not work without config. If we could work out a way to provide the default config rather than empty (ie, by processing the `InstrumentationConfiguration` without input), it might work a bit better?

src/SDK/SdkAutoloader.php
brettmc: done.

src/SDK/SdkAutoloader.php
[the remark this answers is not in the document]
brettmc: Ok. I made that change yesterday, so now SDK config does happen in an initializer.

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
```diff
@@ -0,0 +1,78 @@
...
+use function extension_loaded;
+use Nevay\SPI\ServiceProviderDependency\ExtensionDependency;
+use OpenTelemetry\Context\Context;
+use OpenTelemetry\Context\ContextInterface;
+use OpenTelemetry\Context\ContextKeyInterface;
+use ReflectionFunction;
+
+/** @phan-file-suppress PhanUndeclaredClassAttribute */
+
+#[ExtensionDependency('opentelemetry', '^1.0')]
```
brettmc, line 18: @Nevay I don't think SPI is checking this. If I disable opentelemetry extension, I'd expect to not get this back from `ServiceLoader::load(HookManager::class)`, but I do.

Nevay: Dependencies are only checked during autoload dump in `v0.2.0`.
`v0.2.1` revalidates extension dependencies during runtime -> should work without `composer dump-autoload` after updating to the newest version.

examples/src/ExampleInstrumentation.php
```diff
@@ -0,0 +1,55 @@
...
+     * @todo can we pass in just the config for _this_ instrumentation, rather than all?
+     */
+    public function register(HookManager $hookManager, ConfigurationRegistry $configuration): void
+    {
+        $config = $configuration->get(ExampleConfig::class) ?? throw new Exception('example instrumentation must be configured');
+        if (!$config->enabled) {
+            return;
+        }
+
+        $tracer = Globals::tracerProvider()->getTracer('example-instrumentation');
```
Nevay, line 28: This triggers initialization during sdk autoloading; iff we want to use the `Globals` providers to delay initialization, we should specify that `Globals` must not be accessed outside of hook callbacks. Otherwise I don't see an advantage over injecting the providers.

Apart from that I would generally prefer if we inject the providers. It allows initializing telemetry instances outside of hook callbacks¹ and might allow us to change `Globals` to be a truly global state instead of being `Context` dependent.

¹ For example reusing a counter instance and calling `::add()` directly is around three times faster than refetching the internal instrument from a `CachedInstrumentation` instance. (initial counter creation w/ single metric reader/metric stream takes ~33μs):

```
+---------------------------------+---------+----------+--------+---------+
| subject                         | memory  | mode     | rstdev | stdev   |
+---------------------------------+---------+----------+--------+---------+
| benchCounterAdd (empty)         | 2.716mb | 0.902μs  | ±1.95% | 0.018μs |
| benchCounterAddRecreate (empty) | 2.780mb | 3.319μs  | ±1.86% | 0.062μs |
| benchCounterAdd (http)          | 2.727mb | 1.954μs  | ±1.63% | 0.032μs |
| benchCounterAddRecreate (http)  | 2.783mb | 4.539μs  | ±1.73% | 0.079μs |
+---------------------------------+---------+----------+--------+---------+
```

src/API/Instrumentation/AutoInstrumentation/ExtensionHookManager.php
brettmc: tested against 0.2.1, and it works as expected.

examples/src/ExampleInstrumentation.php
brettmc: That makes sense, and I prefer the method of passing in the providers when registering an instrumentation. I don't think we can commit to file-based config and deprecate `Registry` just yet, given the experimental state of config.

But, I think we can start to move in that direction. How does this sound?
1. update existing instrumentation classes to only retrieve providers from hook callbacks
2. document this new restriction (probably in `Globals`)
3. add a todo to Registry to deprecate in a future release (in case anybody's reading the source)
4. pass in Noop providers to `Instrumentation`s via `::register()`, and document that these will become real implementations in a future release but to keep using `Globals` for now

We then wait for config to stabilise.

Nevay: > I don't think we can commit to file-based config and deprecate Registry just yet, given the experimental state of config.

The restriction / issue with `Registry` is not related to file-based config; any usage of `Globals` within an `autoload.files` file might trigger this issue already, for example [`open-telemetry/opentelemetry-auto-openai-php`](https://github.com/open-telemetry/opentelemetry-php-contrib/blob/f9b3ca8c83c74a19b62ce0bf96ba8dacf19ca1a7/src/Instrumentation/OpenAIPHP/src/OpenAIPHPInstrumentation.php#L45-L49).

4. If we have to stay compatible with `Registry::register{Xxx}Factory` for now:
Instead of noop providers we could inject providers that resolve `Globals` on the first call to `::spanBuilder()` / `::create{Instrument}()` / `::emit()`. This would allow us to use the injected providers immediately (with the same restriction as `CachedInstrumentation` that `Tracer`/`Meter`/`Logger` methods must only be called within hook callbacks for now).

<details>
<summary>TracerProvider example</summary>

```php
$tracerProvider = new LateBindingTracerProvider(static function(): TracerProviderInterface {
    $scope = Context::getRoot()->activate();
    try {
        return Globals::tracerProvider();
    } finally {
        $scope->detach();
    }
});

final class LateBindingTracerProvider implements TracerProviderInterface {

    private ?TracerProviderInterface $tracerProvider = null;

    /** @param Closure(): TracerProviderInterface $factory */
    public function __construct(
        private readonly Closure $factory,
    ) {}

    public function getTracer(string $name, ?string $version = null, ?string $schemaUrl = null, iterable $attributes = []): TracerInterface {
        return $this->tracerProvider?->getTracer($name, $version, $schemaUrl, $attributes)
            ?? new LateBindingTracer(fn(): TracerInterface => ($this->tracerProvider ??= ($this->factory)())->getTracer($name, $version, $schemaUrl, $attributes));
    }
}

final class LateBindingTracer implements TracerInterface {

    private ?TracerInterface $tracer = null;

    /** @param Closure(): TracerInterface $factory */
    public function __construct(
        private readonly Closure $factory,
    ) {}

    public function spanBuilder(string $spanName): SpanBuilderInterface {
        return ($this->tracer ??= ($this->factory)())->spanBuilder($spanName);
    }
}
```
</details>

brettmc: Implemented late-binding providers based on the above example (thanks!)

examples/instrumentation/configure_instrumentation_global.php
```diff
@@ -0,0 +1,21 @@
...
+use const PHP_EOL;
+
+/**
+ * This example uses SPI (see root composer.json extra.spi) to configure an example auto-instrumentation from a YAML file.
+ * The YAML file paths are relative to the current working directory.
+ */
+// EXAMPLE_INSTRUMENTATION_SPAN_NAME=test1234 php examples/instrumentation/configure_instrumentation_global.php
+putenv('OTEL_PHP_AUTOLOAD_ENABLED=true');
+putenv('OTEL_EXPERIMENTAL_CONFIG_FILE=examples/instrumentation/otel-sdk.yaml');
+putenv('OTEL_PHP_INSTRUMENTATION_CONFIG_FILE=examples/instrumentation/otel-instrumentation.yaml');
```
brettmc, line 17: NB that I'm working on a PR over in the config repo to incorporate instrumentation into the general file-based config. I did a quick test against, and combining them was a pretty minor change.

src/SDK/SdkAutoloader.php
ChrisLightfootWild: > If we could work out a way to provide the default config rather than empty (ie, by processing the `InstrumentationConfiguration` without input), it might work a bit better?

This is the bit I'm watching out for too, as I'm trying to keep existing behaviour locally without providing config files.

brettmc: One solution could be to extend the `InstrumentationConfiguration` interface so that each config class can provide its own defaults:
```php
interface InstrumentationConfiguration
{
    public static function default(): self;
}

class ExampleInstrumentation implements InstrumentationConfiguration
{
    public static function default(): self
    {
        return new self('default1', 'another-default');
    }
}

public function register(HookManager $hookManager, ConfigurationRegistry $configuration, InstrumentationContext $context): void
    {
        $config = $configuration->get(ExampleConfig::class) ?? ExampleConfig::default();
        //snip
    }
}
```

brettmc: I still haven't solved this, however I wonder if we could actually set the defaults in the config provider's `getConfig()` method? That said, I guess defaults could be handled in the instrumentation code like in the example above...

brettmc:
@open-telemetry/php-approvers @Nevay  I think I'm happy with this and it's ready for review!

examples/instrumentation/configure_instrumentation_global.php
brettmc: this has been merged and included in this PR now...

src/SDK/SdkAutoloader.php
```diff
@@ -19,91 +33,164 @@
...
-            return $configurator
-                ->withTracerProvider($tracerProvider)
-                ->withMeterProvider($meterProvider)
-                ->withLoggerProvider($loggerProvider)
-                ->withEventLoggerProvider($eventLoggerProvider)
-                ->withPropagator($propagator)
-            ;
-        });
+        if (Configuration::has(Variables::OTEL_EXPERIMENTAL_CONFIG_FILE)) {
+            Globals::registerInitializer(fn ($configurator) => self::fileBasedInitializer($configurator));
```
Nevay, line 51: We'll need to disable the hook manager during initialization to avoid autoinstrumenting the SDK.

brettmc: Discussed at SIG, and we couldn't think of a scenario where SDK would be instrumented only during initialization?

Nevay: The SDK might use classes/methods that are autoinstrumented by a user (http clients etc.); SDK exports use the context that was active during the SDK initialization (related: https://github.com/open-telemetry/opentelemetry-php/pull/1304#discussion_r1599039084) -> disabling the hooks during initialization disables them for all SDK exports.

(AFAIK Java avoids this issue by loading autoinstrumentation classes in a separate classloader)

brettmc: Understood (and I now remember this happening in #878 ). I think I've done this (by making hookmanager enable/disable static methods which seemed the cleanest solution) - WDYT?

Nevay: If we want to make enabling/disabling static: methods should be moved out of `ExtensionHookManager`.
Alternatively we could provide an accessor to the global hook manager and still allow enabling / disabling hook managers independently.

brettmc: ok, re-jiggered things a little by moving the static methods into `HookManager`. How does this approach look?

The main wart I see is that I renamed the existing interface, `HookManager` to `HookManagerInterface` which looks untidy in composer.json's extra.spi. But, `HookManager::disable()` seemed like the most obvious approach.

bobstrecansky (approved):

PR #1306, opened 2024-05-07
Add infection test tool

brettmc:
- run unit tests in random order
running in random order highlights some interference between tests (mostly logging being enabled), and is a requirement for
mutation testing (ie, all tests must pass when executed in a random order). Fix the failing tests so that everything consistently passes when tests are run in random order.
- adding infection mutation testing
this adds the ability to run infection. It generates a lot of output, which is an exercise for another PR...

PR #1308, opened 2024-05-09
upgrade to phpunit 11

brettmc:
- fix `make rector` so that it applies to all configured dirs, not just `src`
- use rector to upgrade tests to phpunit 11 (mostly changing annotations to attributes)
- replace php-global-state package (from which we use the `EnvironmentVariables` trait), with a local trait which manages env vars plus cleaning up our global state after each test
- update to phpunit 11


brettmc:
In draft because it's built on #1306 which needs to be merged first...

tests/Unit/SDK/Common/Util/WeakenTest.php
```diff
@@ -7,13 +7,14 @@
...
+use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\TestCase;
 use WeakReference;
 
 /**
- * @covers \OpenTelemetry\SDK\Common\Util\closure
- * @covers \OpenTelemetry\SDK\Common\Util\weaken
+ * @todo configure phpunit coverage with attributes
  */
+#[CoversNothing]
```
brettmc, line 17: help wanted: how can you use attributes to cover namespaced functions?

brettmc: fixed. It wasn't that hard with fresh eyes the next morning.

bobstrecansky (approved):

PR #1309, opened 2024-05-13
Adds experimental service.instance.id resource detector

matt-hensley:
Adds experimental service.instance.id resource detector per [experimental semantic conventions](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#service-experimental).

This detector has been added to the list of well known detectors with the ID `service` and can be enabled via the `OTEL_PHP_DETECTORS` environment variable.


brettmc:
@matt-hensley looking good. phan and deptrac issues are config-related:
- add uuid to .phan/config.php/directory_list
- add uuid to deptrac.yaml and allow SDK to depend on it

src/SDK/Resource/Detectors/Service.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+final class Service implements ResourceDetectorInterface
+{
+    public function getResource(): ResourceInfo
+    {
+        static $serviceInstanceId = null;
+
+        if ($serviceInstanceId === null) {
+            $serviceInstanceId = \Ramsey\Uuid\Uuid::uuid4()->toString();
+        }
```
brettmc, line 23:
```suggestion
        static $serviceInstanceId;
        $serviceInstanceId ??= Uuid::uuid4()->toString();
```
plus, a `use Ramsey\Uuid\Uuid;` up above

deptrac.yaml
```diff
@@ -85,6 +85,10 @@ deptrac:
               value: ^GuzzleHttp\\*
             - type: className
               value: ^Buzz\\*
+      -   name: RamseyUuid
```
brettmc, line 88: indentation looks wonky here... tabs vs spaces?

src/SDK/Resource/Detectors/Service.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Resource\Detectors;
+
+use OpenTelemetry\SDK\Common\Attribute\Attributes;
+use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
+use OpenTelemetry\SDK\Resource\ResourceInfo;
+use OpenTelemetry\SemConv\ResourceAttributes;
+use \Ramsey\Uuid\Uuid;
```
brettmc, line 11:
```suggestion
use Ramsey\Uuid\Uuid;
```

brettmc:
Couple of minor nits from me, but otherwise looks good.

composer.json
```diff
@@ -10,11 +10,12 @@
         "php": "^8.1",
         "google/protobuf": "^3.22",
         "php-http/discovery": "^1.14",
+        "psr/http-client": "^1.0",
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
-        "psr/http-client": "^1.0",
         "psr/http-message": "^1.0.1|^2.0",
         "psr/log": "^1.1|^2.0|^3.0",
+        "ramsey/uuid": "^4.7",
```
brettmc, line 18: I think it'd be worth checking what is the oldest version this will work with, and making that the minimum.

deptrac.yaml
matt-hensley: indentation in this file is strange, made this and the previous section match the other `layers` children.

composer.json
matt-hensley: 4.3.0 was released March 26, 2022 and looks to work correctly

composer.json
```diff
@@ -10,11 +10,12 @@
         "php": "^8.1",
         "google/protobuf": "^3.22",
         "php-http/discovery": "^1.14",
+        "psr/http-client": "^1.0",
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
-        "psr/http-client": "^1.0",
         "psr/http-message": "^1.0.1|^2.0",
         "psr/log": "^1.1|^2.0|^3.0",
+        "ramsey/uuid": "^4.3",
```
brettmc, line 18: I think we can support all the way back to `3.0.0`: https://github.com/ramsey/uuid/blob/3.0.0/src/Uuid.php#L647

brettmc (approved):

composer.json
```diff
@@ -10,11 +10,12 @@
         "php": "^8.1",
         "google/protobuf": "^3.22",
         "php-http/discovery": "^1.14",
+        "psr/http-client": "^1.0",
         "psr/http-client-implementation": "^1.0",
         "psr/http-factory-implementation": "^1.0",
-        "psr/http-client": "^1.0",
         "psr/http-message": "^1.0.1|^2.0",
         "psr/log": "^1.1|^2.0|^3.0",
+        "ramsey/uuid": "^3.0 || ^4.0",
```
ChrisLightfootWild, line 18: These changes should be added to `src/SDK/composer.json` as well.

ChrisLightfootWild (changes requested):

tests/Unit/SDK/Resource/Detectors/ServiceTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $resource = $this->detector->getResource();
+
+        $this->assertNotEmpty($resource->getAttributes());
+
+        $id = $resource->getAttributes()->get(ResourceAttributes::SERVICE_INSTANCE_ID);
+        $this->assertMatchesRegularExpression(self::UUID_REGEX, $id);
+    }
+
+    public function test_service_get_resource_multiple_calls_same_service_instance_id(): void
```
ChrisLightfootWild, line 43: This is multiple calls within the same process, but typically, will most things be running as shared-nothing and be creating a new ID per request?

I think if that is the case, maybe this detector should be opted into rather than available as a default (perhaps more so as it is still marked experimental?)

agoallikmaa (approved):

tests/Unit/SDK/Resource/Detectors/ServiceTest.php
matt-hensley: No problem making this opt-in since the PHP process model could cause incorrect values to be generated.

Is there a better approach for sharing an instance ID across the processes, or a more idiomatic way to derive an instance ID in PHP?

Random UUID v4 is the recommended default, but any useful value can be used to generate a UUID v5 per the spec.

brettmc: Yeah, let's make it opt-in for now. re: sharing an id across processes - I think that would need to be worked out per-runtime, if they provide something we can use.
https://www.php.net/php_sapi_name could be useful, and if not, maybe instance id could be _disabled_ for shared-nothing based on sapi?

brettmc:
The failure is because we just moved from `EnvironmentVariables` test trait to `OpenTelemetry\Tests\TestState` - might have been a mis-merge for that to not come across when you merged in main...

brettmc:
@matt-hensley you can ignore the psalm errors. I'm aware of them: https://github.com/mockery/mockery/issues/1421
If we get desperate, a workaround is to go back a minor version of mockery.

brettmc:
@matt-hensley if you merge in main now, those psalm errors should be gone.

ChrisLightfootWild (approved):

PR #1316, opened 2024-05-22
update semconv to 1.26.0

brettmc:
note that messaging.client_id was renamed to messaging.client.id, which causes a conflict in const name. It hasn't been resolved upstream yet, but manually removing the duplicate seems like the best way to go for us.

### DIFF
--- a/script/semantic-conventions/templates/trace_deprecations.php.partial
+++ b/script/semantic-conventions/templates/trace_deprecations.php.partial
@@ -6,27 +6,7 @@
     /**
      * @deprecated
      */
-    public const HTTP_HOST = 'http.host';
-
-    /**
-     * @deprecated
-     */
-    public const HTTP_REQUEST_CONTENT_LENGTH_UNCOMPRESSED = 'http.request_content_length_uncompressed';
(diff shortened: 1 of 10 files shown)


script/semantic-conventions/templates/trace_deprecations.php.partial
```diff
@@ -3,31 +3,11 @@
      */
     public const FAAS_EXECUTION = 'faas.execution';
 
-    /**
```
brettmc, line 6: all these guys were re-added to semconv as deprecated, so no longer needed here.

ChrisLightfootWild (approved):

agoallikmaa (approved):

brettmc:
Converting back to draft, awaiting agreement in https://github.com/open-telemetry/semantic-conventions/issues/1031 on how to handle `messaging.client.id`

brettmc:
> The recommendation for messaging.client_id -> messaging.client.id would be to drop the old attribute.

The recommendation from semconv maintainers is to clobber the old, which is what this PR does.

Issue #1328, opened 2024-06-12
OpenTelemetry Specification 1.34.0 Release - Events - Add specification for EventLogger and EventLoggerProvider


bobstrecansky:
Add specification for EventLogger and EventLoggerProvider

We need to follow the following specification change:
Add specification for EventLogger and EventLoggerProvider
https://github.com/open-telemetry/opentelemetry-specification/pull/4031

Issue #1327, opened 2024-06-12
OpenTelemetry Specification 1.34.0 Release - Events -  Rename event payload to body


bobstrecansky:
Rename event payload to body

We need to follow the following specification change:
Rename event payload to body.
https://github.com/open-telemetry/opentelemetry-specification/pull/4035

PR #1335, opened 2024-06-14
event logger spec compliance

brettmc:
update event logger to conform to spec 1.34.0, specifically:
- implement ForceFlush, which delegates to the LoggerProvider
- rename `payload` to `body`

Fixes: #1328 
Fixes: #1327


weslenteche (approved):

src/API/Logs/EventLoggerInterface.php
```diff
@@ -13,7 +13,7 @@ interface EventLoggerInterface
 {
     public function emit(
         string $name,
-        mixed $payload = null,
+        mixed $body = null,
```
ChrisLightfootWild, line 16: This is technically a breaking change if you are using named parameters; should we have any concerns around this?

agoallikmaa (approved):

src/API/Logs/EventLoggerInterface.php
brettmc: Only minor concerns. Since [that part of the spec is still in-development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md), changes should be expected. In fact, the whole logs signal seems to still be in-development/experimental.

ChrisLightfootWild (approved):

Issue #1323, opened 2024-06-12
OpenTelemetry Specification 1.34.0 Release - Traces - Clarify the trace SDK should log discarded events and links


bobstrecansky:
Clarify the trace SDK should log discarded events and links.

We need to follow the following specification change:
https://github.com/open-telemetry/opentelemetry-specification/pull/4064

PR #1336, opened 2024-06-15
log discarded span attributes, events, links

brettmc:
The SDK should log once if there are any dropped span attributes, links or events.
The SDK should log once if there are any dropped LogRecord attributes.

Fixes: #1323


src/SDK/Trace/Span.php
```diff
@@ -249,6 +251,15 @@ public function end(int $endEpochNanos = null): void
             return;
         }
 
+        $spanData = $this->toSpanData();
```
agoallikmaa, line 254: Would it perhaps make sense to skip logging for not sampled spans? All built-in processor implementations discard them anyway. Another option could be having this whole check in the processors since there's no additional `toSpanData` overhead then.

src/SDK/Trace/Span.php
```diff
@@ -249,6 +251,15 @@ public function end(int $endEpochNanos = null): void
...
+        if ($spanData->getTotalDroppedLinks() || $spanData->getTotalDroppedEvents() || $spanData->getAttributes()->getDroppedAttributesCount()) {
+            self::logWarning('Dropped span attributes, links or events', [
```
agoallikmaa, line 256: Having the trace and span IDs also logged could help people find the actual traces/spans where this happened. While the actual dropped links/events/attributes would obviously not be recorded there, at least it gives more context.

src/SDK/Trace/Span.php
brettmc: I wasn't really happy with needing the extra `toSpanData` call. Perhaps we could move the check into `toSpanData` itself (which would not be called on a non-sampled span). That effectively moves the check into the processors, without duplicating the code into each sampler? (although, it could be in a trait to remove that duplication)...

agoallikmaa (approved):

src/SDK/Trace/Span.php
```diff
@@ -290,6 +292,18 @@ public function toSpanData(): SpanDataInterface
             $this->endEpochNanos,
             $this->hasEnded
         );
+
+        if ($spanData->getTotalDroppedLinks() || $spanData->getTotalDroppedEvents() || $spanData->getAttributes()->getDroppedAttributesCount()) {
```
Nevay, line 296: Should also check for dropped event or link attributes.

Nevay (changes requested):
> Would it perhaps make sense to skip logging for not sampled spans?

IMO we should log discarded data for all recorded spans, not only for sampled spans. Usage of a sampler that records a span indicates that a user is interested in the span data (the built-in samplers return DROP or RECORD_AND_SAMPLE, not RECORD_ONLY).

> Outside of tests, this function is only called from span processors once

Every span processor calls `::toSpanData()` -> currently would be logged multiple times if multiple span processors are registered.


(We'll need the same for `LogRecord` attributes [spec](https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits))

src/SDK/Trace/Span.php
brettmc: Updated to check+log on LogRecord dropped attributes, span event+link dropped attributes.

Moved the check into span's `onEnd`, so that it is called once per span. This does add an extra call to `toSpanData`. I didn't profile the difference, but it looks like a computationally light function...

PR #1358, opened 2024-08-13
add minimal auto-instrumentation hook functions

brettmc:
add OpenTelemetry\API\Instrumentation\Handler, which will be the default pre/post hook callbacks for attribute-based auto-instrumentation.

Related: https://github.com/open-telemetry/opentelemetry-php-instrumentation/pull/146

### DIFF
--- a/src/API/Instrumentation/Handler.php
+++ b/src/API/Instrumentation/Handler.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API\Instrumentation;
+
+use OpenTelemetry\API\Trace\Span;
+use OpenTelemetry\API\Trace\SpanKind;
+use OpenTelemetry\API\Trace\StatusCode;
+use OpenTelemetry\Context\Context;
+use Throwable;
+
+class Handler
+{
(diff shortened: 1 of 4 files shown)


src/API/Instrumentation/Handler.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace OpenTelemetry\API\Instrumentation;
+
+use OpenTelemetry\API\Trace\Span;
+use OpenTelemetry\API\Trace\SpanKind;
+use OpenTelemetry\API\Trace\StatusCode;
+use OpenTelemetry\Context\Context;
+use Throwable;
+
+class Handler
```
agoallikmaa, line 13: Maybe the name of this class could reflect that this is the handler specifically for creating spans? It is the most common purpose why functions are instrumented, but would still be nice for it to be explicit.

agoallikmaa: Latter seems to make more sense to me.

brettmc: renamed to `WithSpanHandler`

agoallikmaa (approved):

PR #1371, opened 2024-08-28
adding auto-instrumentation attribute to API

brettmc:
adding WithSpan and SpanAttribute attributes to the API. These attributes are used by an upcoming feature in auto-instrumentation which allows users to add attributes to their code to enable auto-instrumentation.


ChrisLightfootWild (approved):
:star:

PR #1395, opened 2024-10-02
Fix HookManagerInterface annotation 

ChrisLightfootWild:
Align OpenTelemetry\API\Instrumentation\AutoInstrumentation\HookManagerInterface::hook postHook Closure signature with OpenTelemetry\Instrumentation\hook.

Phan complains in open-telemetry/opentelemetry-php-contrib#269. Upon checking, I think the annotated signature of the post-hook is possibly incorrect?

This is an example of what I am seeing:.

In pre-hook:
```
src/Hooks/Illuminate/Contracts/Http/Kernel.php:48 PhanTypeMismatchArgumentSuperType Argument 3 ($preHook) is (function) of type Closure(\Illuminate\Contracts\Http\Kernel,array,string,string,?string,?int):array{0:?\Illuminate\Http\Request} but \OpenTelemetry\API\Instrumentation\AutoInstrumentation\HookManagerInterface::hook() takes Closure(null|object|string,array,string,string,null|string,int|null):void|null defined at vendor/open-telemetry/api/Instrumentation/AutoInstrumentation/HookManagerInterface.php:17 (expected type to be the same or a subtype, but saw a supertype instead)
```

In post-hook:
```
\OpenTelemetry\API\Instrumentation\AutoInstrumentation\HookManagerInterface::hook() takes Closure(null|object|string,array,mixed,\Throwable|null):void|null defined at vendor/open-telemetry/api/Instrumentation/AutoInstrumentation/HookManagerInterface.php:17 (expected type to be the same or a subtype, but saw a supertype instead
```

Edit: it was late and I pasted an example of pre-hook instead of post-hook. Both were throwing phan errors.

### DIFF
--- a/src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
+++ b/src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
@@ -9,10 +9,9 @@ use Throwable;
 
 interface HookManagerInterface
 {
-
     /**
-     * @param Closure(object|string|null,array,string,string,string|null,int|null):void|null $preHook
-     * @param Closure(object|string|null,array,mixed,Throwable|null,string,string,string|null,int|null):void|null $postHook
+     * @param ?Closure(object|string|null, array, ?string, string, ?string, ?int): (array|null|void) $preHook
+     * @param ?Closure(object|string|null, array, mixed, ?Throwable): mixed $postHook
      */
     public function hook(?string $class, string $function, ?Closure $preHook = null, ?Closure $postHook = null): void;
 }
--- a/src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
+++ b/src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
@@ -11,7 +11,7 @@ interface HookManagerInterface
 {
     /**
      * @param ?Closure(object|string|null, array, ?string, string, ?string, ?int): (array|null|void) $preHook
-     * @param ?Closure(object|string|null, array, mixed, ?Throwable): mixed $postHook
+     * @param ?Closure(object|string|null, array, mixed, ?Throwable): (mixed|void) $postHook
      */
     public function hook(?string $class, string $function, ?Closure $preHook = null, ?Closure $postHook = null): void;
 }


src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
```diff
@@ -9,10 +9,9 @@
 
 interface HookManagerInterface
 {
-
     /**
      * @param Closure(object|string|null,array,string,string,string|null,int|null):void|null $preHook
-     * @param Closure(object|string|null,array,mixed,Throwable|null,string,string,string|null,int|null):void|null $postHook
+     * @param Closure(object|string|null,array,mixed,Throwable|null):void|null $postHook
```
brettmc, line 14: Definitely those extra args in post are wrong, but it looks wonky in other ways too. I _think_ this is correct?
```
     * @param ?Closure(object|string|null,array,?string,string,?string,?int):array $preHook
     * @param ?Closure(object|string|null,array,mixed,?Throwable):mixed $postHook
```

ChrisLightfootWild: Yeah, thanks for following up with that. I put the PR back into draft as I thought return type also looked wonky.

I shall take a further look and see if I can get phan to be happy with the changes 👍

ChrisLightfootWild: @brettmc I've updated the PR with those changes.

I was still getting complaints with the preHook expecting an array, where it actually appears to permit void|null too. Marking the return of pre as `?array` suppressed the phan complaints for me - so hoping this is okay? :thinking:

ChrisLightfootWild: (Fixing the build first, oops!)

ChrisLightfootWild: Phan seemed a bit precious around parsing a mixed return type of `@param` - but seems happier with wrapping it in parentheses.

I have seen pre-hooks without an explicit return (thus void) which I thought would might benefit to add?

brettmc: Yes you are right, it's definitely allowed to return nothing from hooks.

brettmc (approved):

src/API/Instrumentation/AutoInstrumentation/HookManagerInterface.php
```diff
@@ -9,10 +9,9 @@
 
 interface HookManagerInterface
 {
-
     /**
-     * @param Closure(object|string|null,array,string,string,string|null,int|null):void|null $preHook
-     * @param Closure(object|string|null,array,mixed,Throwable|null,string,string,string|null,int|null):void|null $postHook
+     * @param ?Closure(object|string|null, array, ?string, string, ?string, ?int): (array|null|void) $preHook
+     * @param ?Closure(object|string|null, array, mixed, ?Throwable): mixed $postHook
```
Nevay, line 14:
```suggestion
     * @param ?Closure(object|string|null, array, mixed, ?Throwable): (mixed|void) $postHook
```

PR #1398, opened 2024-10-11
use weaver to generate semconv

brettmc:
update semconv generation to use the new tooling. This is fairly close to a 1:1 of what we have now, although I noted a couple of TODOs which we might consider in future to align with what Java is doing: split attribute up into file-per-type, and split stable/unstable.

The diffs are fairly small (mostly related to formatting of comments, which weaver improves). I've removed our manually-maintained deprecations, since the semconv SIG is a lot more careful now about removing stuff - that accounts for most removals.

Fixes: #1396


script/semantic-conventions/semconv.sh
```diff
@@ -9,20 +9,18 @@
 # Source repositories:
 #  - https://github.com/open-telemetry/semantic-conventions/releases
 #  - https://github.com/open-telemetry/build-tools/releases
-set -e
+set -ex
```
bobstrecansky, line 12: Do we want to write a trace for each command before execution?

bobstrecansky (approved):
One minor non-blocking nit, but the rest looks kosher.  Thanks @brettmc

script/semantic-conventions/semconv.sh
brettmc: It doesn't hurt, and is quite helpful for debugging. Other SIGs which I took inspiration from also have this set...

Issue #1397, opened 2024-10-10
EACCESS in host detectors crash opentelemetry initialization "trim(): Argument #1 ($string) must be of type string, bool given"


pkramme:
**Describe your environment** 

Custom hosting environment with a limited environment in which PHP apps are run. Recently did an upgrade from opentelemetry-php/sdk 1.0.8 to 1.1.0.

**Steps to reproduce**
1. Prevent read access to `/etc/machine-id`.

**What is the expected behavior?**
opentelemetry-php/sdk's host detectors should not crash the application, if for any reasons any of the host identifiers cannot be read.

**What is the actual behavior?**
```
OpenTelemetry: [warning] Error during opentelemetry initialization: trim(): Argument #1 ($string) must be of type string, bool given
#0 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/Resource/Detectors/Host.php(56): trim()
#1 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/Resource/Detectors/Host.php(42): OpenTelemetry\SDK\Resource\Detectors\Host->getLinuxId()
#2 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/Resource/Detectors/Host.php(33): OpenTelemetry\SDK\Resource\Detectors\Host->getMachineId()
#3 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/Resource/Detectors/Composite.php(24): OpenTelemetry\SDK\Resource\Detectors\Host->getResource()
#4 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/Resource/ResourceInfoFactory.php(38): OpenTelemetry\SDK\Resource\Detectors\Composite->getResource()
#5 /var/www/hubble_101/dc9f7037-0cdb-4e9c-b723-daf0b6a02eb4/releases/167/vendor/open-telemetry/sdk/SdkAutoloader.php(84): OpenTelemetry\SDK\Resource\ResourceInfoFactory::defaultResource()
[...]

PR #1400, opened 2024-10-13
check for readable host id files before read

brettmc:
Check if files are readable before trying to read them. Also check whether `kenv` exists before executing it, since it complains when I run it locally on a non-BSD system.

Closes: #1397

### DIFF
--- a/src/SDK/Resource/Detectors/Host.php
+++ b/src/SDK/Resource/Detectors/Host.php
@@ -52,8 +52,9 @@ final class Host implements ResourceDetectorInterface
         $paths = [self::PATH_ETC_MACHINEID, self::PATH_VAR_LIB_DBUS_MACHINEID];
 
         foreach ($paths as $path) {
-            if (file_exists($this->dir . $path)) {
-                return trim(file_get_contents($this->dir . $path));
+            $file = $this->dir . $path;
+            if (file_exists($file) && is_readable($file)) {
+                return trim(file_get_contents($file));
             }
         }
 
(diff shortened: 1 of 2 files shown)


tests/Unit/SDK/Resource/Detectors/HostTest.php
```diff
@@ -72,4 +73,35 @@ public static function hostIdData(): array
             ['BSD', $etc_hostid, '1234567890'],
         ];
     }
+
+    /**
+     * @dataProvider nonReadableFileProvider
+     */
```
ChrisLightfootWild, line 79:
```suggestion
     #[DataProvider('nonReadableFileProvider')]
```

ChrisLightfootWild: I think you made a start at switching these across now 💪

src/SDK/Resource/Detectors/Host.php
```diff
@@ -52,8 +52,9 @@ private function getLinuxId(): ?string
         $paths = [self::PATH_ETC_MACHINEID, self::PATH_VAR_LIB_DBUS_MACHINEID];
 
         foreach ($paths as $path) {
-            if (file_exists($this->dir . $path)) {
-                return trim(file_get_contents($this->dir . $path));
+            $file = $this->dir . $path;
+            if (file_exists($file) && is_readable($file)) {
```
Nevay, line 56:
```suggestion
            if (is_file($file) && is_readable($file)) {
```
Currently fails if `$file` is a readable directory.

Alternatively use only `file_get_contents()` and check return value to handle all possible failure scenarios:
```php
set_error_handler(static fn() => true);
try {
    if (($content = file_get_contents($file)) !== false) {
        return trim($content);
    }
} finally {
    restore_error_handler();
}
```

ChrisLightfootWild (approved):

PR #1412, opened 2024-10-18
use SPI to manage components

brettmc:
To resolve our long-standing race conditions stemming from using composer's autoload->files to register SDK components at runtime, this changes things so that:
- components can be registered in various composer.json files, under `extra.spi`
- existing `_register.php` files manually register through SPI, but will be removed from `autoload_files` if SPI is allowed to run
- SDK fetches factories from SPI instead of maintaining static arrays

If we move ahead with this approach, a follow-up PR could remove our various late-binding providers.


Nevay:
An alternative idea if we want to avoid duplicating the "load implementations from `composer.json`" approach:

The main difference between the solution in this PR and an SPI-based approach is the addition of a name/key for each implementation within the `composer.json` configuration. This information could instead be added as additional method in the interfaces[^1] to allow loading them using SPI.

[^1]: See also [Java ServiceLoader - Designing Services](https://cr.openjdk.org/~mr/jigsaw/spec/api/java/util/ServiceLoader.html):
    > A service should declare as many methods as needed to allow service providers to communicate their domain-specific properties and other quality-of-implementation factors. An application which obtains a service loader for the service may then invoke these methods on each instance of a service provider, in order to choose the best provider for the application.


`Registry` is currently used for two different types of implementations:
- factories used to initialize the SDK from environment variables
- transport factory implementations, which are also used outside of environment variables configuration

#### Transport factory implementations

We can extend `TransportFactoryInterface` with the necessary information to populate the transport factories:
```php
$factories = iterator_to_array(ServiceLoader::load(TransportFactoryServiceInterface::class));
array_multisort(
    array_map(static fn($factory) => $factory->priority(), $factories),
    SORT_DESC,
    $factories,
);
$factoriesByProtocol = [];
foreach ($factories as $factory) {
    $factoriesByProtocol[$factory->protocol()] ??= $factory;
}

self::$transportFactories = $factoriesByProtocol;
```
```php
interface TransportFactoryServiceInterface extends TransportFactoryInterface {

    public function protocol(): string;
    
    public function priority(): int;
}
```

#### Factories to initialize the SDK from environment variables

We could align the implementation with the file-based implementation by adding a new, from the current implementation independent interface[^2] similar to [`Configuration\ComponentProvider`] and deprecate the current registry methods and factory interfaces.
Very basic interface definition:
```php
namespace OpenTelemetry\Config\SDK\Environment;

/**
 * @template T
 */
interface ComponentProvider {

    /**
     * @return T
     */
    public function createPlugin(): mixed;

    public function name(): string;
}
```

Alternatively we could follow the approach mentioned for transport factory implementations and add a name(/priority) method to the factory interfaces and continue using the `Registry`.

[^2]: FWIW I've been using the following [`Env\Loader`] interface for env-based configuration, its implementations are [loaded via SPI](https://github.com/Nevay/otel-sdk-config/blob/8eff71317b04c28989cafcc22db64425a55b1411/Env.php#L36-L39):
    ```php
    /**
     * @template T
     */
    interface Loader {

        /**
         * @return T
         */
        public function load(EnvResolver $env, LoaderRegistry $registry, Context $context): mixed;

        public function name(): string;
    }
    ```

[`Configuration\ComponentProvider`]: https://github.com/open-telemetry/opentelemetry-php/blob/93f2068d0bee0301c663d783d688d61911fdb6f4/src/Config/SDK/Configuration/ComponentProvider.php
[`Env\Loader`]: https://github.com/Nevay/otel-sdk-config/blob/8eff71317b04c28989cafcc22db64425a55b1411/Env/Loader.php

brettmc:
updated to use SPI, and implemented the idea of having factories declare their key & priority.

haad:
- [ ]   The Vneno i

src/SDK/Common/Export/Http/PsrTransportFactory.php
```diff
@@ -43,6 +43,9 @@ public function create(
             throw new InvalidArgumentException(sprintf('Invalid endpoint url "%s"', $endpoint));
         }
         assert(!empty($endpoint));
+        if ($compression === 'none') {
+            $compression = null;
+        }
```
ChrisLightfootWild, line 48: Would this be better suited within `PsrUtils::compression()`?

src/SDK/Trace/SpanExporter/SpanExporterFactoryInterface.php
```diff
@@ -9,4 +9,6 @@
 interface SpanExporterFactoryInterface
 {
     public function create(): SpanExporterInterface;
+    public function type(): string;
+    public function priority(): int;
```
ChrisLightfootWild, line 13: As above with an [example](https://github.com/opentelemetry-php/exporter-zipkin/blob/main/_register.php#L5) from the Zipkin exporter.

src/SDK/Common/Export/TransportFactoryInterface.php
```diff
@@ -29,4 +29,7 @@ public function create(
         ?string $cert = null,
         ?string $key = null,
     ): TransportInterface;
+
+    public function type(): string;
+    public function priority(): int;
```
ChrisLightfootWild, line 34: Should we be considering this to be a breaking change? 🤔

Example in [exporter-otlp](https://github.com/opentelemetry-php/exporter-otlp/blob/main/OtlpHttpTransportFactory.php).

src/SDK/Logs/LogRecordExporterFactoryInterface.php
```diff
@@ -7,4 +7,6 @@
 interface LogRecordExporterFactoryInterface
 {
     public function create(): LogRecordExporterInterface;
+    public function type(): string;
```
ChrisLightfootWild, line 10: Should we be considering this to be a breaking change? 🤔

src/SDK/Metrics/MetricExporterFactoryInterface.php
```diff
@@ -7,4 +7,6 @@
 interface MetricExporterFactoryInterface
 {
     public function create(): MetricExporterInterface;
+    public function type(): string;
+    public function priority(): int;
```
ChrisLightfootWild, line 11: As above.

src/Contrib/Zipkin/SpanExporterFactory.php
```diff
@@ -20,4 +20,13 @@ public function create(): SpanExporterInterface
 
         return new Exporter($transport);
     }
+
+    public function type(): string
+    {
+        return 'zipkin';
+    }
+    public function priority(): int
```
ChrisLightfootWild, line 28:
```suggestion
    }

    public function priority(): int
```

Sorry :grimacing: 
Hopefully there's a lint setting somewhere so I don't look so pedantic :sweat_smile:

ChrisLightfootWild (commented):
This is an exciting set of changes!

Just left a few bits for consideration/discussion.

src/Extension/Propagator/B3/B3MultiPropagatorFactory.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Extension\Propagator\B3;
+
+use OpenTelemetry\Context\Propagation\TextMapPropagatorFactoryInterface;
+use OpenTelemetry\Context\Propagation\TextMapPropagatorInterface;
+
+class B3MultiPropagatorFactory implements TextMapPropagatorFactoryInterface
+{
+
```
xvilo, line 12:
```suggestion
```

xvilo (commented):
One small code-style comment was placed. 

I understand that the Registry stuff will not be used when using `extra.spi`. Is it correct that third-party libraries (in my case, proprietary code/packages within a company) can also add their instrumentation through the SPI way, as shown in `composer.json`? 

It would be nice to pair this with some documentation regardless, e.g. a 'How to add instrumentation into your composer package' section. It would be a good way to demo/reflect/document the preferred way of including AutoInstrumentations.

brettmc:
> I understand that the Registry stuff will not be used when using extra.spi. Is it correct that third-party libraries (in my case, proprietary code/packages within a company) can also add their instrumentation through the SPI way, as shown in composer.json?

@xvilo That's right. It's still up for debate exactly how registry/SPI should interact, and which one would be the default. The SPI mechanism will definitely allow custom components like we have now.

src/Contrib/Zipkin/SpanExporterFactory.php
brettmc: there is a fixer that can do this:
`'class_attributes_separation' => ['elements' => ['method' => 'one'],]`
It changes a lot of files, so I won't do it in this PR, but did fix the blanks b/w `priority()`

src/SDK/Common/Export/TransportFactoryInterface.php
brettmc: This is a good point, we should consider trying to do this in a non-breaking way. Perhaps another level of abstraction like a `TransportFactoryProvider` & friends which has these extra methods and returns a factory :thinking:

brettmc: Noted with a label this this is a BC break. I've recently created a PR to get things going for 2.x versions of packages, and when that is done I'll rebase this to the 2.x branch

src/Contrib/Zipkin/SpanExporterFactory.php
brettmc: this fixer has since been applied to the codebase.

src/SDK/Common/Export/TransportFactoryInterface.php
brettmc: Have rebased onto 2.x branch, and note that this is a BC.

composer.json
```diff
@@ -123,7 +122,53 @@
...
+                "src/Contrib/Otlp/_register.php",
+                "src/Contrib/Grpc/_register.php",
+                "src/Contrib/Zipkin/_register.php",
+                "src/Extension/Propagator/B3/_register.php",
+                "src/Extension/Propagator/CloudTrace/_register.php",
+                "src/Extension/Propagator/Jaeger/_register.php",
+                "src/SDK/Logs/Exporter/_register.php",
+                "src/SDK/Metrics/MetricExporter/_register.php",
+                "src/SDK/Trace/SpanExporter/_register.php"
+            ]
```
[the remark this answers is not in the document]
brettmc: Reading the SPI docs, this is related to supporting registering service providers via autoload.files (which is a fallback for when the SPI plugin is not allowed to run), right?

For this to work, we would:
1. register service providers "properly" via `extra.spi`
2. update all `_register.php` files to manually/directly register via SPI
3. add those `_register.php` files to `extra.spi-config.autoload-files`
4. set `extra.spi-config.prune_autoload_files` to true

Does that sound right?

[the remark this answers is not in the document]
brettmc: I got this working, except for in SDK's composer.json since it has other autoload files that we wouldn't want to prune.

brettmc:
@open-telemetry/php-approvers @Nevay I think this is ready for review: NB that it targets 2.x branch and has some BC breakage, which I've started to document in `docs/upgrading.md`

tests/Unit/SDK/Common/Services/LoaderTest.php
```diff
@@ -88,15 +95,13 @@ public static function logRecordExporterProvider(): array
...
-        $propagator = Registry::textMapPropagator($name);
+        $propagator = Loader::textMapPropagator($name);
         $this->assertInstanceOf(TextMapPropagatorInterface::class, $propagator);
     }
 
     public static function textMapPropagator(): array
     {
         return [
-            ['tracecontext'],
-            ['baggage'],
```
Nevay, line 99: Why were these removed?

src/SDK/Common/Export/Http/PsrUtils.php
```diff
@@ -117,7 +117,7 @@ public static function compression($compression): array
         if (is_array($compression)) {
             return $compression;
```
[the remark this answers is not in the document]
brettmc: do you mean check for 'none' in an array of compression? Which takes precedence if found, eg for `gzip,none` is the result `[gzip]` or `[]`?

tests/Unit/SDK/Common/Services/LoaderTest.php
brettmc: Lost in translation, added back and registered through SPI

composer.json
```diff
@@ -123,7 +122,56 @@
...
+        },
         "spi": {
```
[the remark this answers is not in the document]
brettmc: I've moved most of these to an appropriate `_register.php`, excepting some test instances and the API's ExtensionHookManager...

src/SDK/Common/Export/Http/PsrUtils.php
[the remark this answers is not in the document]
brettmc: done: now removes 'none' compression from the list.

composer.json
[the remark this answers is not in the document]
ChrisLightfootWild: Are the leftover ones there to serve as an example? Or does this simplify development by negating the need for dump-autoload during development? :thinking:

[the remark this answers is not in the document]
brettmc: ExtensionHookManager is in API, and we shouldn't have an SDK dependency. The rest are text/example implementations and while I could use autoload-dev.files for those, it would require adding them to spi-config which didn't seem right.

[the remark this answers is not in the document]
brettmc: Moved all spi config into various _register.php files

src/Contrib/Zipkin/composer.json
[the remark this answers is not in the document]
brettmc: Good idea. I have made SPI a direct dependency of all packages which register service providers. This allowed removing some `class_exists`, and now everything works consistently (I think).

src/Extension/Propagator/B3/_register.php
```diff
@@ -2,18 +2,12 @@
...
-use OpenTelemetry\SDK\Registry;
-
-if (!class_exists(Registry::class)) {
-    return;
+use Nevay\SPI\ServiceLoader;
+use OpenTelemetry\Context\Propagation\TextMapPropagatorFactoryInterface;
+use OpenTelemetry\Extension\Propagator\B3\B3MultiPropagatorFactory;
+use OpenTelemetry\Extension\Propagator\B3\B3PropagatorFactory;
+
+if (class_exists(ServiceLoader::class)) {
```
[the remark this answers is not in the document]
brettmc: Resolved by making SPI a dependency of this and other extension packages.

composer.json
```diff
@@ -123,51 +126,19 @@
...
+                "src/API/_register.php",
+                "src/Contrib/Otlp/_register.php",
+                "src/Contrib/Grpc/_register.php",
+                "src/Contrib/Zipkin/_register.php",
+                "src/Extension/Propagator/B3/_register.php",
+                "src/Extension/Propagator/CloudTrace/_register.php",
+                "src/Extension/Propagator/Jaeger/_register.php",
+                "src/SDK/_register.php",
+                "tests/_register.php"
             ]
```
Nevay, line 142:
```suggestion
            "autoload-files": [
                "examples/src/_register.php",
                "src/API/_register.php",
                "src/Contrib/Otlp/_register.php",
                "src/Contrib/Grpc/_register.php",
                "src/Contrib/Zipkin/_register.php",
                "src/Extension/Propagator/B3/_register.php",
                "src/Extension/Propagator/CloudTrace/_register.php",
                "src/Extension/Propagator/Jaeger/_register.php",
                "src/SDK/_register.php",
                "tests/_register.php"
            ],
            "prune-autoload-files": true
```

src/API/composer.json
```diff
@@ -37,9 +39,10 @@
...
             "dev-2.x": "2.x-dev"
         },
-        "spi": {
-            "OpenTelemetry\\API\\Instrumentation\\AutoInstrumentation\\HookManagerInterface": [
-                "OpenTelemetry\\API\\Instrumentation\\AutoInstrumentation\\ExtensionHookManager"
+        "spi-config": {
+            "autoload-files": true,
+            "prune-autoload-files": [
+                "_register.php"
             ]
```
Nevay, line 46:
```suggestion
            "autoload-files": [
                "_register.php"
            ],
            "prune-autoload-files": true
```

src/Context/Propagation/TextMapPropagatorFactoryInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\Context\Propagation;
+
+interface TextMapPropagatorFactoryInterface
```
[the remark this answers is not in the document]
brettmc: Moved things around in https://github.com/open-telemetry/opentelemetry-php/pull/1412/commits/0571d3b965d827d04458d63a2be99d845dc80dd8 - WDYT?

brettmc:
@open-telemetry/php-approvers this is ready for another review. I think all feedback from Nevay addressed and OK'd.

bobstrecansky (approved):

PR #1421, opened 2024-10-30
Support google protobuf v4

roxblnfk:
If google protobuf v4 is not allowed, the package version 1.0.0beta5 will be installed, which breaks the application.

The same value is in the root composer.json:
https://github.com/open-telemetry/opentelemetry-php/blob/1aae9dd61930ac02d2936e1eda795c39c4f7a959/composer.json#L11

### DIFF
--- a/proto/otel/composer.json
+++ b/proto/otel/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": "^8.0",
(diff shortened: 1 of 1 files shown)


brettmc (approved):

brettmc:
tagged 1.2.1 with this change, @roxblnfk

PR #1430, opened 2024-11-07
guard against possible false value from file_get_contents in Host.php

jspaetzel:
There's other circumstances besides is_file && is_readable when file_get_contents may return false. This is a defensive change to prevent TypeErrors when calling trim()


jspaetzel:
Will follow-up on CLA asap

brettmc (approved):

brettmc:
@jspaetzel can you run php-cs-fixer to resolve its complaint? Otherwise LGTM and just waiting on CLA to be signed :)

ChrisLightfootWild (approved):

jspaetzel:
:heavy_check_mark: Thanks y'all!

Issue #1431, opened 2024-11-08
"OTEL_PHP_AUTOLOAD_ENABLED=true" not working in php.ini


Willhang23:
"true" in php.ini are evaluated to "1" by get_cfg_var() (OpenTelemetry\SDK\Common\Configuration\Resolver\PhpIniAccessor) while the OpenTelemetry\SDK\Common\Configuration\Parser\BooleanParser only accepts "true" as true value.

brettmc:
Discussed a little in the linked PR... are you sure you're correctly quoting "true" in php.ini? The opentelemetry spec says we have to use "true" and not other things that could also mean true, and in my testing double-quoting resolves this. We might need a documentation patch to clarify this, though.

brettmc:
Related to #1442 and fixed by https://github.com/open-telemetry/opentelemetry.io/pull/5849 - that is, the required format for php.ini would be `OTEL_PHP_AUTOLOAD_ENABLED="true"`, noting that the value is double-quoted

PR #1435, opened 2024-11-20
A string with a number could be passed as a boolean value

stickeerehan:
get_cfg_var() evaluates "true" in php.ini to "1" which is not correctly parsed by the boolean parser.

Fixes issue raised on [https://github.com/open-telemetry/opentelemetry-php/issues/1431](https://github.com/open-telemetry/opentelemetry-php/issues/1431).

### DIFF
--- a/src/SDK/Common/Configuration/Parser/BooleanParser.php
+++ b/src/SDK/Common/Configuration/Parser/BooleanParser.php
@@ -28,11 +28,11 @@ class BooleanParser
             return false;
         }
 
-        if ((int)($value) === 1) {
+        if ((int) ($value) === 1) {
             return true;
         }
 
-        if ((int)($value) === 0) {
+        if ((int) ($value) === 0) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


brettmc:
Hi @stickeerehan thanks for the contribution. We also need to be compliant with the OpenTelemetry spec as much as possible, and here we run into this part: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value

> Any value that represents a Boolean MUST be set to true only by the case-insensitive string "true", meaning "True" or "TRUE" are also accepted, as true. An implementation MUST NOT extend this definition and define additional values that are interpreted as true. Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false.

Having a quick play with this:
`php.ini`:
```ini
my_bool = true
my_quoted_bool = "true"
```

```
$ php -a
Interactive shell

php > var_dump(get_cfg_var('my_bool'));
string(1) "1"
php > var_dump(get_cfg_var('my_quoted_bool'));
string(4) "true"
```

So, quoting boolean values in php.ini seems to work, and maybe we need to document that?

bobstrecansky (approved):

brettmc:
Docs updated in https://github.com/open-telemetry/opentelemetry.io/pull/5849 - boolean values need to be double-quoted in php.ini

PR #1453, opened 2024-12-12
Do not validate instrumentation, if the whole SDK is disabled.

profuel:
Do not validate instrumentation, if the whole SDK is disabled.

### DIFF
--- a/src/SDK/Sdk.php
+++ b/src/SDK/Sdk.php
@@ -35,6 +35,9 @@ class Sdk
      */
     public static function isInstrumentationDisabled(string $name): bool
     {
+        if (static::isDisabled()) {
+            return true;
+        }
         $disabledInstrumentations = Configuration::getList(Variables::OTEL_PHP_DISABLED_INSTRUMENTATIONS);
 
         return [self::OTEL_PHP_DISABLED_INSTRUMENTATIONS_ALL] === $disabledInstrumentations || in_array($name, $disabledInstrumentations);


brettmc:
This needs some thought. Per spec, disabling the SDK should still allow propagators: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration
We have auto-instrumentations for http clients, and would expect that they would still propagate distributed trace headers if the SDK is disabled. I think this would break that?

profuel:
That's an interesting detail, I was missing.
Could you please help me understand - does it mean that this code is not correct?
https://github.com/spryker/otel-rabbit-mq-instrumentation/blob/master/_register.php
And we have to use similar logic to this :
```$disabledInstrumentations = Configuration::getList(Variables::OTEL_PHP_DISABLED_INSTRUMENTATIONS);

        return [self::OTEL_PHP_DISABLED_INSTRUMENTATIONS_ALL] === $disabledInstrumentations || in_array($name, $disabledInstrumentations);```.

Thank you for your help!

brettmc:
> does it mean that this code is not correct

I think it probably doesn't matter for rabbitmq (it's not performing distributed tracing). Our current thinking is that if the SDK is disabled, then when you fetch tracers/meters/loggers from the various providers, you'll get a no-op instance.

brettmc:
@profuel thanks for your contribution, however we won't progress with this change due to the propagator thing - it would break some of our instrumentations according to spec.

Issue #1450, opened 2024-12-11
File(/etc/machine-id) is not within the allowed path(s)


ZebraNorth:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your php version (`php -v` will tell you your current version), version numbers of installed dependencies, information about your cloud hosting provider, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on master.

Linux, PHP 8.3.11, Laravel 11, opentelemetry-sdk v1.1.2

**Steps to reproduce**
Describe exactly how to reproduce the error. Include a code sample if applicable.

Edit `php.ini` to include `open_basedir=/var/www`

**What is the expected behavior?**
What did you expect to see?

The site should continue to work as before.

**What is the actual behavior?**
What did you see instead?

During page load, the following exception is thrown:

`is_file(): open_basedir restriction in effect. File(/etc/machine-id) is not within the allowed path(s):`

```
OpenTelemetry\SDK\Resource\Detectors\Host
:58
getLinuxId
 
OpenTelemetry\SDK\Resource\Detectors\Host
:42
getMachineId
 
OpenTelemetry\SDK\Resource\Detectors\Host
:33
getResource
 
OpenTelemetry\SDK\Resource\Detectors\Composite
:24
getResource
 
OpenTelemetry\SDK\Resource\ResourceInfoFactory
:38
defaultResource
 
Stickee\Instrumentation\Laravel\ServiceProvider
:158
Stickee\Instrumentation\Laravel\{closure}
 
Illuminate\Container\Container
:937
build
 
Illuminate\Container\Container
:819
resolve
 
Illuminate\Foundation\Application
:1048
resolve
 
Illuminate\Container\Container
:755
make
 
Illuminate\Foundation\Application
:1030
make
 
PlunkettScott\LaravelOpenTelemetry\Otel
:39
start
 
PlunkettScott\LaravelOpenTelemetry\OtelServiceProvider
:26
boot
 
Illuminate\Container\BoundMethod
:36
Illuminate\Container\{closure}
 
Illuminate\Container\Util
:43
unwrapIfClosure
 
Illuminate\Container\BoundMethod
:95
callBoundMethod
 
Illuminate\Container\BoundMethod
:35
call
 
Illuminate\Container\Container
:694
call
 
Illuminate\Foundation\Application
:1121
bootProvider
 
Illuminate\Foundation\Application
:1102
array_walk
 
[...]

PR #1451, opened 2024-12-11
Don't throw if open_basedir is enabled

ZebraNorth:
Closes #1450 

Don't throw an unhandled exception if the file `/etc/machine-id` cannot be read.

PR #1454, opened 2024-12-13
handle host detector with custom error handler and open_basedir

brettmc:
if `open_basedir` is configured such that files that the host detector tries to open are denied with a php warning, and a custom error handler is installed which converts warnings to exceptions, an unhandled exception occurs.
add a test for this, and some try/catch/ignore blocks around the code that can trigger this.

Closes: #1450
Replace: #1451

### DIFF
--- a/src/SDK/Resource/Detectors/Host.php
+++ b/src/SDK/Resource/Detectors/Host.php
@@ -9,7 +9,6 @@ use OpenTelemetry\SDK\Resource\ResourceDetectorInterface;
 use OpenTelemetry\SDK\Resource\ResourceInfo;
 use OpenTelemetry\SemConv\ResourceAttributes;
 use function php_uname;
-use Throwable;
 
 /**
  * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/resource/semantic_conventions/host.md#host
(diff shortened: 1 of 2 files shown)


src/SDK/Resource/Detectors/Host.php
```diff
@@ -66,10 +72,15 @@ private function getLinuxId(): ?string
...
         $file = $this->dir . self::PATH_ETC_HOSTID;
-        if (is_file($file) && is_readable($file)) {
-            $contents = file_get_contents($file);
 
-            return $contents !== false ? trim($contents) : null;
+        try {
+            if (is_file($file) && is_readable($file)) {
+                $contents = file_get_contents($file);
+
+                return $contents !== false ? trim($contents) : null;
```
Nevay, line 80: Should not return here if `file_get_contents()` returns false, we still want to attempt the fallback.

Nevay (commented):
IMO we should disable the error handler instead.
```php
set_error_handler(static fn() => true);
try {
    if (($content = file_get_contents($file)) !== false) {
        return trim($content);
    }
} finally {
    restore_error_handler();
}
```

src/SDK/Resource/Detectors/Host.php
brettmc: updated to disable the error handler

src/SDK/Resource/Detectors/Host.php
```diff
@@ -47,16 +47,37 @@ private function getMachineId(): ?string
...
 
+    /**
+     * @phan-suppress PhanTypeMismatchArgumentInternal
+     */
+    private function readFile(string $file): string|false
+    {
+        set_error_handler(static fn () => true);
+
+        try {
+            if (is_file($file) && is_readable($file)) {
```
Nevay, line 58: These checks are no longer needed.

ChrisLightfootWild (approved):

tests/Integration/SDK/Resource/Detectors/test_host_detector_with_open_basedir.phpt
```diff
@@ -0,0 +1,34 @@
+--TEST--
+Host detector with custom error handler and open_basedir
+--DESCRIPTION--
+An error handler is installed which converts PHP warnings to exceptions, and open_basedir
+is configured such that /etc/machine-id and friends cannot be read
+--SKIPIF--
+<?php if (!in_array(PHP_OS_FAMILY, ['Linux', 'BSD'])) die('skip requires Linux or BSD'); ?>
+--ENV--
+OTEL_PHP_FIBERS_ENABLED=0
```
Nevay, line 9:
```suggestion
```

Nevay (approved):

PR #1456, opened 2024-12-13
update links to openmetrics to reference the v1.0.0 release

dashpole:
Related to https://github.com/prometheus/OpenMetrics/issues/287

The OM 2.0 effort is kicked off, and will start developing on main. This updates the links to OpenMetrics to reference the 1.0 release. The OM project has also been moved into the prometheus github org.


brettmc:
Hi! It looks like you've already updated the upstream proto definition in https://github.com/open-telemetry/opentelemetry-proto/pull/612 so what we need to do here is regenerate the proto definition from source. When a new version is tagged containing your PR's changes, let us know via an issue or on slack and I'll regenerate.

PR #1462, opened 2024-12-20
Instana exporter implementation.

kothiga:
This PR is to contribute an OpenTelemetry exporter which facilitates conversion and transmission of spans to Instana.

I have also opened two other PRs for documentation related to this work
* https://github.com/open-telemetry/opentelemetry-specification/pull/4346
* https://github.com/open-telemetry/opentelemetry.io/pull/5834

I hope that I have added everything correctly for the read-only git subtree split.


brettmc:
Hi @kothiga 
The opentelemetry-php repository is for our implementation of the opentelemetry specification, and none of this stuff is in the spec.
Taking a quick look through [instana's docs](https://www.ibm.com/docs/en/instana-observability/current?topic=apis-opentelemetry), I see that `Instana can seamlessly integrate with OpenTelemetry`. Native OTLP support for ingesting telemetry is the direction most if not all APM vendors are moving in, and that's what we're all about here.
So, if Instana already natively supports the OTLP protocol, why is this PR required? And if it is required for some use-cases, wouldn't it be better done in the opentelemetry collector, so that it would work for all languages?

kothiga:
Hi @brettmc!

Just to start off, we have customers that are using our native protocol that we still need to support. Our native protocol tracer supports discovery and autotracing, but there is quite a bit of disparity with Otel PHP on these features as well as frameworks supported. 

With this contribution we wanted to provide support for the Otel SDK to natively support proprietary APIs in OpenTelemetry.

> wouldn't it be better done in the opentelemetry collector

The idea here was to not have to require our native customers to install additional Otel collectors where they are already installing our native agent. 

It is our intention to provide more contributions to OpenTelemetry PHP in the future to close the existing differences.

Issue #1459, opened 2024-12-16
add Logger.emitEvent


brettmc:
spec [1.40.0](https://github.com/open-telemetry/opentelemetry-specification/releases/tag/v1.40.0) deprecates the event api+sdk in favour of using the log API: https://github.com/open-telemetry/opentelemetry-specification/pull/4319

Depends on proto 1.5.0 (for event_name top-level field)

edit: on-hold, EmitEvent has been removed from spec and no replacement yet

brettmc:
Closing, there is no replacement. Now, the log API is user-facing.

PR #1466, opened 2025-01-06
deprecate event logger

brettmc:
The event logger was a Development-status component of the logging signal. It has been removed in favour of adding emitEvent to the logger interface, see https://github.com/open-telemetry/opentelemetry-specification/pull/4319

Related: #1459

### DIFF
--- a/src/SDK/SdkAutoloader.php
+++ b/src/SDK/SdkAutoloader.php
@@ -74,6 +74,9 @@ class SdkAutoloader
         return true;
     }
 
+    /**
+     * @phan-suppress PhanDeprecatedClass,PhanDeprecatedFunction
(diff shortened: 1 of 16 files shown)


src/SDK/SdkAutoloader.php
```diff
@@ -74,6 +74,9 @@ public static function autoload(): bool
         return true;
     }
 
+    /**
+     * @phan-suppress PhanDeprecatedClass,PhanDeprecatedFunction
```
ChrisLightfootWild, line 78: Could this work if moved to `phan-suppress-next-line` around the event logger? 🤔

agoallikmaa (approved):

PR #1480, opened 2025-01-16
Remove event logger

brettmc:
Event Logger has been removed from spec, in favour of Logger.EmitEvent. We have deprecated EventLogger in `main`, this removes it from `2.x` and adds `EmitEvent` as its replacement.


src/API/Logs/NoopLogger.php
```diff
@@ -35,4 +36,16 @@ public function isEnabled(): bool
     {
         return false;
     }
+
+    public function emitEvent(
```
weslenteche, line 40: Add `@codeCoverageIgnore` to this method `emitEvent`

weslenteche (approved):

src/API/Logs/LoggerInterface.php
```diff
@@ -12,6 +14,20 @@ interface LoggerInterface
      */
     public function emit(LogRecord $logRecord): void;
 
+    /**
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.40.0/specification/logs/api.md#emit-an-event
+     */
+    public function emitEvent(
```
Nevay, line 20: We shouldn't add this method until it is defined by the spec.
(Was removed in https://github.com/open-telemetry/opentelemetry-specification/pull/4352)

brettmc: Good catch. I'll remove this portion for now.

brettmc:
Worth mentioning and discussion, is that in this PR I've removed some methods etc related to event logger. [The spec](https://opentelemetry.io/docs/specs/otel/versioning-and-stability/#extending-apisdk-abstractions) really wants us to not ever make breaking changes to the API. However:

> It should be documented that the newly added methods are in Development and are subject to breaking changes.

The event logger was `development` stability, just not documented as such by us. I think we're allowed to remove non-stable features in a breaking way...

ChrisLightfootWild (approved):

PR #1482, opened 2025-01-21
implement mutable ReadWriteLogRecord

brettmc:
https://github.com/open-telemetry/opentelemetry-specification/pull/3907 implements some new requirements for logging:
- ReadWriteLogRecord can mutate (eg by processors)
- mutated ReadWriteLogRecord can be seen by later processors

This is a breaking change because LogRecordProcessorInterface onEmit param changes to by-reference

### DIFF
--- a/src/API/Logs/LogRecord.php
+++ b/src/API/Logs/LogRecord.php
@@ -16,8 +16,6 @@ class LogRecord
     protected int $severityNumber = 0;
     protected ?string $severityText = null;
     protected array $attributes = [];
-    /** @experimental */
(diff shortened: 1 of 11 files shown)


bobstrecansky (commented):
Seems like a reasonable change to me.  We should chat about the breaking change tomorrow in the SIG meeting.

bobstrecansky:
Ahh, darn it - we didn't talk about this in the SIG meeting.  Let's chat about it.

bobstrecansky (approved):

PR #1483, opened 2025-01-23
implement SpanProcessor::onEnding

brettmc:
added to spec 1.36.0


src/SDK/Trace/Span.php
```diff
@@ -253,6 +253,7 @@ public function end(?int $endEpochNanos = null): void
 
         $this->endEpochNanos = $endEpochNanos ?? Clock::getDefault()->now();
         $this->hasEnded = true;
+        $this->spanProcessor->onEnding($this);
```
Nevay, line 256: > The Span object MUST still be mutable while `OnEnding` is called.

> The SDK MUST guarantee that the span can no longer be modified by any other thread before invoking OnEnding of the first SpanProcessor. From that point on, modifications are only allowed synchronously from within the invoked OnEnding callbacks.

brettmc: Which part are you querying? `$this` is a Span implementing `ReadWriteSpanInterface` so is still mutable (and there's a test to show this working). I don't think we need to worry about other threads, and I don't see any other opportunity for modifying the span...

Nevay: > $this is a Span implementing ReadWriteSpanInterface so is still mutable

Setting `$this->hasEnded = true` prevents all further writes to the span (besides `::setAttributes()`, which is a bug); currently `::onEnding()` cannot modify the span, the following example fails:

```php
$tracerProvider = (new TracerProviderBuilder())
    ->addSpanProcessor(new class implements SpanProcessorInterface {

        public function onEnding(ReadWriteSpanInterface $span): void {
            $span->updateName('updated');
            assert($span->getName() === 'updated');
        }

        public function onStart(ReadWriteSpanInterface $span, ContextInterface $parentContext): void {}
        public function onEnd(ReadableSpanInterface $span): void {}
        public function forceFlush(?CancellationInterface $cancellation = null): bool { return true; }
        public function shutdown(?CancellationInterface $cancellation = null): bool { return true; }
    })
    ->build();

$span = $tracerProvider->getTracer('test')->spanBuilder('span')->startSpan();
$span->end();
```

> I don't think we need to worry about other threads, and I don't see any other opportunity for modifying the span...

Most mentions of "other threads" also apply to fibers/coroutines, for example the following shouldn't throw an assertion error after fixing the issue above:

```php
$tracerProvider = (new TracerProviderBuilder())
    ->addSpanProcessor(new class implements SpanProcessorInterface {

        public function onEnding(ReadWriteSpanInterface $span): void {
            $spanName = $span->getName();

            delay(0);
            assert($span->getName() === $spanName);
        }

        public function onStart(ReadWriteSpanInterface $span, ContextInterface $parentContext): void {}
        public function onEnd(ReadableSpanInterface $span): void {}
        public function forceFlush(?CancellationInterface $cancellation = null): bool { return true; }
        public function shutdown(?CancellationInterface $cancellation = null): bool { return true; }
    })
    ->build();

$span = $tracerProvider->getTracer('test')->spanBuilder('span')->startSpan();
async($span->updateName(...), 'should-not-update');
$span->end();
```

---

One solution would be to pass a new span object to `::onEnding()` while still `hasEnded=true` ending the original span.
A basic implementation (which will create unnecessary array copies on events/links modification that could be avoided by storing the state in a separate object):

```php
public function end(?int $endEpochNanos = null): void
{
    if ($this->endEpochNanos) { // should be nullable and check for !== null instead
        return;
    }

    $this->endEpochNanos = $endEpochNanos ?? Clock::getDefault()->now();
    $span = clone $this;
    $this->hasEnded = true;
    $this->spanProcessor->onEnding($span);
    $span->hasEnded = true;

    $span->checkForDroppedElements();

    $this->spanProcessor->onEnd($span);
}
```

brettmc: Sorry, my bad. I am confusing this PR with #1482 which is similar but for logs. Back to draft while I work on this more.

brettmc: Updated to clone+end span, and I added amphp/amp as a dev dependency + an integration test for async mutation.

vendor-bin/psalm/composer.json
```diff
@@ -3,5 +3,8 @@
         "vimeo/psalm": "^5.24",
         "psalm/plugin-mockery": "^1",
         "psalm/plugin-phpunit": "^0.18.4"
+    },
+    "replace": {
+        "amphp/amp": "*"
```
brettmc, line 8: This avoids a conflict between `vendor-bin/psalm`'s version of apm conflicting with the dev dependency.

src/SDK/Trace/Span.php
```diff
@@ -244,19 +248,25 @@ public function setStatus(string $code, ?string $description = null): self
...
     }
 
-    /** @inheritDoc */
+    /**
+     * @inheritDoc
+     * @psalm-suppress NoInterfaceProperties
+     **/
     public function end(?int $endEpochNanos = null): void
     {
         if ($this->hasEnded) {
```
Nevay, line 257: Must check for `$this->endEpochNanos`, calls to `Span::end()` within `::onEnding()` must be no-ops; the following would currently cause an OOM error:
```php
$tracerProvider = (new TracerProviderBuilder())
    ->addSpanProcessor(new class implements SpanProcessorInterface {

        public function onEnding(ReadWriteSpanInterface $span): void {
            $span->end();
        }

        public function onStart(ReadWriteSpanInterface $span, ContextInterface $parentContext): void {}
        public function onEnd(ReadableSpanInterface $span): void {}
        public function forceFlush(?CancellationInterface $cancellation = null): bool { return true; }
        public function shutdown(?CancellationInterface $cancellation = null): bool { return true; }
    })
    ->build();

$span = $tracerProvider->getTracer('test')->spanBuilder('span')->startSpan();
$span->end();
```

src/SDK/Trace/Span.php
```diff
@@ -244,19 +248,25 @@ public function setStatus(string $code, ?string $description = null): self
...
 
         $this->endEpochNanos = $endEpochNanos ?? Clock::getDefault()->now();
-        $this->hasEnded = true;
-
-        $this->checkForDroppedElements();
+        $span = clone $this;
+        $this->hasEnded = true; // prevent further modifications to the span by async code
+        $this->spanProcessor->onEnding($span);
+        $span->hasEnded = true;
+        $span->end();
```
Nevay, line 266:
```suggestion
```

src/SDK/Trace/Span.php
```diff
@@ -244,19 +248,25 @@ public function setStatus(string $code, ?string $description = null): self
...
 
         $this->spanProcessor->onEnd($this);
+        $this->checkForDroppedElements();
```
Nevay, line 269: Must use `$span` instead of `$this` to make sure that changes from `::onEnding()` are visible; moving the second callback in `test_span_processor_onending_can_mutate_span` from `::onEnding()` to `::onEnd()` should currently result in a test failure.

src/SDK/Trace/SpanProcessorInterface.php
```diff
@@ -15,6 +15,12 @@ interface SpanProcessorInterface
      */
     public function onStart(ReadWriteSpanInterface $span, ContextInterface $parentContext): void;
 
+    /**
+     * @experimental
+     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.41.0/specification/trace/sdk.md#onending
+     */
+    public function onEnding(ReadWriteSpanInterface &$span): void;
```
Nevay, line 22:
```suggestion
    public function onEnding(ReadWriteSpanInterface $span): void;
```

Nevay (changes requested):

Nevay (approved):

PR #1490, opened 2025-01-28
handle empty meter observers

brettmc:
if no observers are registered for a meter's instruments, some tests were emitting php warnings:
- foreach() argument must be of type array|object, null given
- Undefined array key XXX

null coalesce to an empty array to make the test happy.


Nevay:
Same fix needed for the `foreach ($this->instruments->writers` loop.
```php
$mp = (new MeterProviderBuilder())
    ->addReader(new ExportingReader(new InMemoryExporter()))
    ->build();
$mp->getMeter('test')->createObservableCounter('c')->observe(static fn() => null);
$mp->updateConfigurator(Configurator::meter());
```

Nevay (approved):

ChrisLightfootWild (approved):

PR #1491, opened 2025-01-28
generate semconv 1.28

brettmc:


ChrisLightfootWild (commented):
Are we able to add the new version [here](https://github.com/open-telemetry/opentelemetry-php/blob/b24a30c31aec75f6ad906f89320f11df7e56a30f/src/SemConv/Version.php#L9) too? 👍

bobstrecansky (approved):

PR #1493, opened 2025-01-29
drop string options for Temporality, InstrumentType

brettmc:
Remove forwards-compatible Metrics features for 2.x:
- No longer accept strings for Temporality and InstrumentType
- `$advisory` added to `DefaultAggregationProviderInterface`

### DIFF
--- a/src/SDK/Metrics/MetricMetadataInterface.php
+++ b/src/SDK/Metrics/MetricMetadataInterface.php
@@ -22,7 +22,7 @@ interface MetricMetadataInterface
     /**
      * Returns the underlying temporality of this metric.
      *
-     * @return ?Temporality internal temporality
+     * @return Temporality internal temporality
      */
-    public function temporality(): ?Temporality;
+    public function temporality(): Temporality;
 }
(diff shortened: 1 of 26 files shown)


src/SDK/Metrics/MetricMetadataInterface.php
```diff
@@ -22,7 +22,7 @@ public function description(): ?string;
     /**
      * Returns the underlying temporality of this metric.
      *
-     * @return string|Temporality internal temporality
+     * @return ?Temporality internal temporality
      */
-    public function temporality();
+    public function temporality(): ?Temporality;
```
brettmc, line 27: I wasn't sure about this, however there is a `lowmemory` test that fails if null is not allowed.

Nevay: Non-nullable is correct here - test should be fixed.

brettmc: Updated the offending test. `lowmemory` = "it depends on the instrument", [per spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.41.0/specification/metrics/sdk_exporters/otlp.md#additional-environment-variable-configuration) so I set a default and confirm that's used.

brettmc:
@open-telemetry/php-approvers I think this is ready

bobstrecansky (approved):

PR #1499, opened 2025-01-31
fix gRPC timeout (#1498)

### LINKED ISSUE #1498: The gRPC exporter does not respect timeout settings
**Describe your environment** 

PHP 8.4, linux (in docker), php-fpm model
grpc module version => 1.69.0
latest version of `transport-grpc`

**Steps to reproduce**
There are 2 scenarios:

Scenario 1 (long timeout) 
1. Set the following env vars:
```
OTEL_EXPORTER_OTLP_PROTOCOL: grpc
OTEL_EXPORTER_OTLP_ENDPOINT: http://google.com:4310
OTEL_EXPORTER_OTLP_TIMEOUT: 20000
```
2. export a batch of spans
3. observe the time for the request to complete

Expected: Request completes after 10 seconds
Actual: timeout completes after about 6 seconds

Scenario 2: short timeout (under 1 second)
1. Set the following env vars:
```
OTEL_EXPORTER_OTLP_PROTOCOL: grpc
OTEL_EXPORTER_OTLP_ENDPOINT: http://google.com:4310
OTEL_EXPORTER_OTLP_TIMEOUT: 999
```
2. export a batch of spans
3. observe the time for the request to complete

Expected: Request completes after about a second
Actual: Request completes immediately because the gRPC call immediately fails in less than 1ms (this also happens when you specify a valid collector endpoint and a 999ms timeout)

spadger:
Fixes #1498 (grpc timeouts not being respected)

The timeout seems to work correctly for long and short variants of the bug with this code but  I wasn't able to create meaningful tests.  Whilst it was possible to get the calculated timeval from the transport instance using reflection,  actually calling functions on `\Grpc\Timeval` (such as [similar()](https://grpc.github.io/grpc/php/class_grpc_1_1_timeval.html#a6bc3609d25c19f9297749781efd1071e) just crashes the test runner

No additional unit tests fail compared to `main` (35 vs 35)
Happy to make any adjustments according to feedback

### DIFF
--- a/src/Contrib/Grpc/GrpcTransport.php
+++ b/src/Contrib/Grpc/GrpcTransport.php
@@ -38,15 +38,18 @@ final class GrpcTransport implements TransportInterface
     private readonly array $metadata;
     private readonly Channel $channel;
     private bool $closed = false;
+    private Timeval $exportTimeout;
 
     public function __construct(
         string $endpoint,
         array $opts,
         private readonly string $method,
         array $headers = [],
+        int $timeoutMillis = 500,
     ) {
         $this->channel = new Channel($endpoint, $opts);
         $this->metadata = $this->formatMetadata(array_change_key_case($headers));
+        $this->exportTimeout = new Timeval($timeoutMillis * 1000);
     }
 
     public function contentType(): string
@@ -60,7 +63,7 @@ final class GrpcTransport implements TransportInterface
             return new ErrorFuture(new BadMethodCallException('Transport closed'));
         }
 
-        $call = new Call($this->channel, $this->method, Timeval::infFuture());
+        $call = new Call($this->channel, $this->method, $this->exportTimeout);
 
         $cancellation ??= new NullCancellation();
         $cancellationId = $cancellation->subscribe(static fn (Throwable $e) => $call->cancel());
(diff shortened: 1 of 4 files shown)


src/Contrib/Grpc/GrpcTransport.php
```diff
@@ -38,15 +38,18 @@ final class GrpcTransport implements TransportInterface
...
     public function __construct(
         string $endpoint,
         array $opts,
         private readonly string $method,
         array $headers = [],
+        int $timeoutMillis = 500,
     ) {
         $this->channel = new Channel($endpoint, $opts);
         $this->metadata = $this->formatMetadata(array_change_key_case($headers));
+        $this->exportTimeout = new Timeval($timeoutMillis * 1000);
```
brettmc, line 52: Can we use a const here, to make it clear we're converting from milliseconds to microseconds? We don't have a const for this yet in `API\Common\Time\ClockInterface`, but I think that's the place to put it.

brettmc:
LGTM - with no collector available I was able to test that the timeout is accurate for <5s values. Beyond 5s, I got a host lookup error which occurred at 5s if the configured timeout doesn't kick in first.

Can you please fix the linting error (`make style` locally, if you're using our test image etc)

brettmc (approved):

PR #1501, opened 2025-02-04
 Improve Contributing guide

RichardChukwu:
This PR is part of [Add Contributing guide to repos](https://github.com/open-telemetry/sig-contributor-experience/issues/31#top)

This update aims to make it easier for contributors to get started with the project and follow best practices while contributing while following a consistent pattern for a setup guide across the Otel ecosystem.

Certain info was taken from the [DEVELOPMENT.md](https://github.com/open-telemetry/opentelemetry-php/blob/main/DEVELOPMENT.md) guide to create this

Please review and let me know if you have any suggestions or feedback.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,19 +71,11 @@ make all-checks
 
 ## Testing
 
-We use various tools for testing and static analysis:
+To ensure your changes meet our project's standards, simply run:
 
 ```bash
-# Run all tests
-make test
+make all
 
-# Run static analysis
-make phan
-make psalm
-make phpstan
-
-# Run style checks
-make style
 ```
 
 To run tests against different PHP versions:
@@ -110,7 +102,7 @@ Most of our communication is done on CNCF Slack in the channel:
  
 - **Slack**: [otel-php](https://cloud-native.slack.com/archives/C01NFPCV44V)
 - **GitHub Discussions**: [OpenTelemetry PHP Discussions](https://github.com/open-telemetry/opentelemetry-php/discussions)
-- **Issues**: If you encounter a bug, [open an issue](https://github.com/open-telemetry/opentelemetry-php-contrib/issues)
+- **Issues**: If you encounter a bug, [open an issue](https://github.com/open-telemetry/opentelemetry-php/issues)
 
 
 ## Troubleshooting Guide
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-Welcome to the OpenTelemetry PHPrepository! 🎉
+Welcome to the OpenTelemetry PHP repository! 🎉
 
 Thank you for considering contributing to this project. Whether you're fixing a bug, adding new features, improving documentation, or reporting an issue, we appreciate your help in making OpenTelemetry better. 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,178 @@
+# Contributing to OpenTelemetry PHP
+
+## Introduction
+
+Welcome to the OpenTelemetry PHPrepository! 🎉
+
+Thank you for considering contributing to this project. Whether you're fixing a bug, adding new features, improving documentation, or reporting an issue, we appreciate your help in making OpenTelemetry better. 
+
+This repository is part of the OpenTelemetry ecosystem, which provides observability tooling for distributed systems. Your contributions help enhance the PHP ecosystem for OpenTelemetry users worldwide.
+
+If you have any questions, feel free to ask in the community channels. We’re happy to help! 😊
+
+## Pre-requisites
+
+Before getting started, ensure you have the following installed:
+
+- **PHP** (8.1 or higher) – [Install PHP](https://www.php.net/downloads)
+- **Composer** – [Install Composer](https://getcomposer.org/)
+- **Docker & Docker Compose** – [Install Docker](https://docs.docker.com/engine/install/)
+- **Make** (for running development tasks)
+
+Additional Notes:
+- Windows users may need [Git Bash](https://gitforwindows.org/) for better compatibility.
+- Consider using [phpenv](https://github.com/phpenv/phpenv) for managing multiple PHP versions.
+
+## Workflow
+
+We follow a structured workflow to ensure smooth collaboration:
+
+### Branch Naming Convention
+- **Feature branches**: `feature/<short-description>`
+- **Bugfix branches**: `fix/<short-description>`
+- **Documentation updates**: `docs/<short-description>`
+
+### Commit Message Format
+- Use descriptive commit messages (e.g., `fix(tracing): resolve issue with span context`)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,178 @@
...
+```bash
+make all-checks
+```
+
+
+## Testing
+
+We use various tools for testing and static analysis:
+
+```bash
```
brettmc, line 76: I think it's worth mentioning that `make all` is the least you need to know as a new contributor, and covers everything that github actions will check.

CONTRIBUTING.md
```diff
@@ -34,10 +209,6 @@ Find more information about the member role in the [community repository](https:
 
 Find more about emeritus roles in the [community repository](https://github.com/open-telemetry/community/blob/main/community-membership.md#emeritus-maintainerapprovertriager)
 
-## Communication
```
brettmc, line 37: Is there a reason for this to not be mentioned?

CONTRIBUTING.md
```diff
@@ -1,3 +1,178 @@
...
+
+## Further Help
+
+Need help? Join our community:
+
+Most of our communication is done on CNCF Slack in the channel:
+ 
+- **Slack**: [otel-php](https://cloud-native.slack.com/archives/C01NFPCV44V)
+- **GitHub Discussions**: [OpenTelemetry PHP Discussions](https://github.com/open-telemetry/opentelemetry-php/discussions)
+- **Issues**: If you encounter a bug, [open an issue](https://github.com/open-telemetry/opentelemetry-php-contrib/issues)
```
ChrisLightfootWild, line 113:
```suggestion
- **Issues**: If you encounter a bug, [open an issue](https://github.com/open-telemetry/opentelemetry-php/issues)
```

CONTRIBUTING.md
RichardChukwu: It is mentioned actually in the guide, see under "Further Help" and the last information in the doc. @brettmc

CONTRIBUTING.md
```diff
@@ -1,3 +1,170 @@
...
+- Follow [Conventional Commits](https://www.conventionalcommits.org/) where possible.
+
+### Pull Request Guidelines
+- Fork the repository and create a new branch.
+- Follow the coding guidelines before submitting your PR.
+- Ensure tests pass locally before pushing.
+- Link relevant issues in the PR description.
+
+
+## Local Run/Build
```
spadger, line 46: Hi

I did my first work on this repo a few days ago, and the makefile doesn't work because of this line
```
include .env
```

I see what it does now, but some guidance on how to provision one would have been helpful

CONTRIBUTING.md
```diff
@@ -1,3 +1,170 @@
...
+
+### Commit Message Format
+- Use descriptive commit messages (e.g., `fix(tracing): resolve issue with span context`)
+- Follow [Conventional Commits](https://www.conventionalcommits.org/) where possible.
+
+### Pull Request Guidelines
+- Fork the repository and create a new branch.
+- Follow the coding guidelines before submitting your PR.
+- Ensure tests pass locally before pushing.
+- Link relevant issues in the PR description.
```
spadger, line 43: Does each PR require an issue?  I have a trivial PR and will make an issue for it, but it'd be good if there were guidance in case there are some circumstances one isn't required

RichardChukwu: I'm a contributor myself, so I guess the maintainers would have better knowledge to give a response to this 

cc @brettmc @ChrisLightfootWild

brettmc: I'm happy to receive PRs without a corresponding issue, provided the PR provides some detail.

CONTRIBUTING.md
brettmc: How to set up `.env` is at the very top of [DEVELOPMENT.md](https://github.com/open-telemetry/opentelemetry-php/blob/main/DEVELOPMENT.md#development)

CONTRIBUTING.md
```diff
@@ -1,3 +1,170 @@
...
+We follow a structured workflow to ensure smooth collaboration:
+
+### Branch Naming Convention
+- **Feature branches**: `feature/<short-description>`
+- **Bugfix branches**: `fix/<short-description>`
+- **Documentation updates**: `docs/<short-description>`
+
+### Commit Message Format
+- Use descriptive commit messages (e.g., `fix(tracing): resolve issue with span context`)
+- Follow [Conventional Commits](https://www.conventionalcommits.org/) where possible.
```
brettmc, line 37: We don't currently have a standard or practise for branch naming or commit messages. I assume this has come from another repository? Is it considered best-practise for all OpenTelemetry projects?

brettmc:
A lot of this seems to be duplicate info from DEVELOPMENT.md - I don't mind which is the source of truth, but I would prefer there was just one version of the truth. Since you're aligning multiple SIGs contribution guides, maybe removing double-ups from development guide?

CONTRIBUTING.md
ChrisLightfootWild: I'm the same as @brettmc. Although there's a risk that contributors spend more time on a PR that would not have been necessary with some up-front discussion in an issue/other channel.

CONTRIBUTING.md
spadger: Oh bloody hell.  I did look for something like `development.md`, totally missed the link in the main readme and then wondered why contributing.md was so slim. My fault 🤠

CONTRIBUTING.md
RichardChukwu: Not sure if there any general format though, it can be removed if there is no standard yet for this repo though, or can be adopted this depending on what is tenable

RichardChukwu:
> A lot of this seems to be duplicate info from DEVELOPMENT.md - I don't mind which is the source of truth, but I would prefer there was just one version of the truth. Since you're aligning multiple SIGs contribution guides, maybe removing double-ups from development guide?

About this, I asked during SIG call attended and the idea was, having information that is relevant to a contributor not a user in the contributing guide was necessary, and apparently most of that info is already in the development guide.

Maybe it can be removed from the development guide instead perhaps?

brettmc:
> Maybe it can be removed from the development guide instead?

I think so - both DEVELOPMENT and CONTRIBUTING seem to serve the same purpose, so how about CONTRIBUTING completely replaces DEVELOPMENT, and remove the latter?

brettmc:
> That can work, so are you removing it?

This is your PR, you're doing it :)
Where I think this is heading is that instead of creating a new CONTRIBUTING, you're renaming DEVELOPMENT to CONTRIBUTING and aligning it to the standards you've been asked to promote.

brettmc:
> The CONTRIBUTING file already has major info from the DEVELOPMENT to create this PR though, so would it be better to just take current info as is and edit into the DEVELOPMENT guide and rename?

I think that's the same thing. End result: DEVELOPMENT no longer exists, CONTRIBUTING contains all useful info from DEVELOPMENT and is in the format you want.

PR #1502, opened 2025-02-05
A couple of code tidies left over from PR #1499

spadger:
Fixes a couple of style issues from #1499

### DIFF
--- a/src/Contrib/Grpc/GrpcTransportFactory.php
+++ b/src/Contrib/Grpc/GrpcTransportFactory.php
@@ -12,6 +12,7 @@ use function in_array;
 use InvalidArgumentException;
 use function json_encode;
 use OpenTelemetry\API\Behavior\LogsMessagesTrait;
+use OpenTelemetry\API\Common\Time\ClockInterface;
 use OpenTelemetry\Contrib\Otlp\ContentTypes;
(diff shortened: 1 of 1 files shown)


ChrisLightfootWild (approved):

brettmc:
Released a new version with latest changes...

spadger:
Thanks :)

src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -81,13 +82,12 @@ public function create(
             $opts,
             $method,
             $headers,
-            (int) ($timeout * 1000),
+            (int) ($timeout * ClockInterface::MILLIS_PER_SECOND),
```
ChrisLightfootWild, line 85: 🤦 sorry, just realised that this causes a dependency on API 1.2.1 (see [here](https://github.com/opentelemetry-php/api/blob/1.2.1/Common/Time/ClockInterface.php#L12)).

spadger:
Do you need me to do anything?

src/Contrib/Grpc/GrpcTransportFactory.php
brettmc: Ugh, yeah. Let's create a local const in GrpcTransportFactory for now.

PR #1505, opened 2025-02-08
use a local const for grpc time conversions

brettmc:
using a newly-added const in ClockInterface breaks things for some older PHP versions, so lets use a local const for now

### DIFF
--- a/src/Contrib/Grpc/GrpcTransportFactory.php
+++ b/src/Contrib/Grpc/GrpcTransportFactory.php
@@ -12,7 +12,6 @@ use function in_array;
 use InvalidArgumentException;
 use function json_encode;
 use OpenTelemetry\API\Behavior\LogsMessagesTrait;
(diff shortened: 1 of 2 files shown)


src/Contrib/Grpc/GrpcTransportFactory.php
```diff
@@ -25,6 +24,8 @@ final class GrpcTransportFactory implements TransportFactoryInterface
 {
     use LogsMessagesTrait;
 
+    public const MILLIS_PER_SECOND = 1_000;
```
ChrisLightfootWild, line 27: If this was `protected`, might we discourage wider usage? (hopefully in favour of `ClockInterface::MILLIS_PER_SECOND` where applicable)

brettmc: Yes, absolutely. Since the class is `final`, I think it can be `private`.

Issue #1436, opened 2024-11-24
[opentelemetry-php-contrib] laravel: When loading configuration files, SDK ignores laravel's .env


gmhafiz:
**Describe your environment** 

OS: linux debian 12
php: 8.2
laravel: v11
open-telemetry/opentelemetry-auto-laravel: 0.0.28
open-telemetry/sdk: 1.1.2

**Steps to reproduce**

This is after a normal laravel installation instrumented with `open-telemetry/opentelemetry-auto-laravel` package. There are two ways of running laravel. Method 1 works with `php artisan serve`. Method 2 using php-fpm does not read environment variable.

Get it and make a copy of `.env` file.
```sh
git clone git@github.com:gmhafiz/laravel-otlp.git
cd laravel-otlp
cp .env.example .env
```

# Method 1:

Run with
```sh
php artisan migrate
php artisan db:seed
php artisan serve
```

Monitor log for traceID

```sh
tail -f storage/logs/laravel.log
```

Perform a request
```sh
curl -v localhost:8000/api/users
```

# Method 2

Run laravel using nginx and php-fpm

```sh
docker build -t otlp/laravel .
docker exec -it otlp-laravel php artisan migrate
docker exec -it otlp-laravel php artisan db:seed
docker run --name otlp-laravel --rm -p 8111:80 otlp/laravel
```

Perform a request

```sh
curl -v localhost:8111/api/users
```

**What is the expected behavior?**

In both methods, traceID is present. For example.

```json
{"datetime":"2024-11-22T08:27:29.243748+00:00","traceID":"9de4bd0cfce131835f531731e950a397","level":200,"message":"random amount is ","context":[15],"extra":{"hostname":"debianDesktop","app":"Laravel","env":"local"}}
```

**What is the actual behavior?**
Method 1 has traceID, but in method 2, otel is not enabled (uses noop) and traceID becomes 0000000000000000000000000000000

**Additional context**

Even though both methods have `OTEL_PHP_AUTOLOAD_ENABLED=true` set in `.env` file, only in the first method otel is enabled.

I am unsure the mechanism of `php artisan serve` is picking up values from `.env`

In method 2, what I think happened was the SDK tries to load values using both `$_SERVER` (from EnvironmentResolver.php) and `getenv()`. And `OTEL_PHP_AUTOLOAD_ENABLED` variable is false at this point for some reason.

What I have done, or try to attempt:

[...]

PR #1523, opened 2025-03-05
PoC: Configuration\CompositeResolver and SPI discovery

ChrisLightfootWild:
Proof of concept, to cover the use-case where multi-application installs operate on shared-infra and wish to provide their own configuration.

See #1436.

---

Any suggestions on the best way to approach adding tests for these optional dependencies?

EDIT: I added a test. It's not my proudest moment, but I can't think of another way to beat the way the SDK currently bootstraps and triggers Configuration setup so early on. We could modify the `CompositeResolver` to allow it's static instance to be reset - but that would be a change which was made purely to improve testability.


src/SDK/Common/Adapter/DotEnv/VlucasPhpdotenvResolver.php
```diff
@@ -0,0 +1,44 @@
...
+namespace OpenTelemetry\SDK\Common\Adapter\DotEnv;
+
+use Dotenv\Dotenv;
+use Dotenv\Exception\InvalidFileException;
+use function getcwd;
+use Nevay\SPI\ServiceProviderDependency\PackageDependency;
+use OpenTelemetry\SDK\Common\Configuration\Resolver\ResolverInterface;
+use function str_starts_with;
+
+#[PackageDependency('vlucas/phpdotenv', '^5.0')]
```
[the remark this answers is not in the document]
Nevay: No, I was only concerned about implementing our own `.env` parsing.

src/SDK/Common/Adapter/DotEnv/VlucasPhpdotenvResolver.php
```diff
@@ -0,0 +1,44 @@
...
+
+    private function dotEnvValues(): array
+    {
+        static $env;
+
+        try {
+            $env ??= array_filter(
+                Dotenv::createArrayBacked([getcwd()])->safeLoad(),
+                // Discard any environment variables that do not start with OTEL_
+                fn (string $name) => str_starts_with($name, 'OTEL_'),
```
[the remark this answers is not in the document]
ChrisLightfootWild: Great point, I shall drop this filtering. :+1:

src/SDK/Common/Adapter/DotEnv/VlucasPhpdotenvResolver.php
```diff
@@ -0,0 +1,44 @@
...
+        return array_key_exists($variableName, $this->dotEnvValues());
+    }
+
+    private function dotEnvValues(): array
+    {
+        static $env;
+
+        try {
+            $env ??= array_filter(
+                Dotenv::createArrayBacked([getcwd()])->safeLoad(),
```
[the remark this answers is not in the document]
ChrisLightfootWild: I think that is a fair assumption, so I shall follow your lead and make the switch.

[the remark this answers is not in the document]
ChrisLightfootWild: Thanks for highlighting this, I'm hoping it's okay that I port your improvements over?! :sweat_smile:

src/SDK/composer.json
```diff
@@ -60,6 +60,9 @@
         "spi": {
             "OpenTelemetry\\API\\Instrumentation\\AutoInstrumentation\\HookManagerInterface": [
                 "OpenTelemetry\\API\\Instrumentation\\AutoInstrumentation\\ExtensionHookManager"
+            ],
+            "OpenTelemetry\\SDK\\Common\\Configuration\\Resolver\\ResolverInterface": [
+                "OpenTelemetry\\SDK\\Common\\Configuration\\Resolver\\SdkConfigurationResolver"
```
ChrisLightfootWild, line 65: Only kicks in when `open-telemetry/sdk-configuration` is available.

src/SDK/Common/Configuration/Resolver/CompositeResolver.php
```diff
@@ -18,6 +19,7 @@ public static function instance(): self
     {
         static $instance;
         $instance ??= new self([
+            ...ServiceLoader::load(ResolverInterface::class),
```
ChrisLightfootWild, line 22: This effectively replaces the below resolvers because it will use the `ServerEnvSource` & `PhpIniEnvSource` sources.

--

Should `open-telemetry/sdk-configuration` be promoted to the SDK rather than remain as an optional dependency? :thinking:

composer.json
```diff
@@ -86,9 +86,9 @@
     },
     "require-dev": {
         "ext-grpc": "*",
-        "grpc/grpc": "^1.30",
         "bamarni/composer-bin-plugin": "^1.8",
         "dg/bypass-finals": "^1.4",
+        "grpc/grpc": "^1.30",
```
ChrisLightfootWild, line 91: Sort package config re-ordered this.

brettmc:
This looks pretty solid now, would you like to rebase and we can get it merged @ChrisLightfootWild ?

src/Config/SDK/Configuration/Environment/LazyEnvSource.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace OpenTelemetry\Config\SDK\Configuration\Environment;
+
+use Closure;
+
+final class LazyEnvSource implements EnvSource
+{
+    /**
+     * @param Closure(): EnvSource|EnvSource $env
```
brettmc, line 12:
```suggestion
     * @param Closure(): EnvSource $env
```

src/SDK/Common/Configuration/Resolver/SdkConfigurationResolver.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function __construct()
+    {
+        $envSources = [];
+        $envSources[] = new ServerEnvSource();
+        $envSources[] = new PhpIniEnvSource();
+
+        /** @var EnvSourceProvider $envSourceProvider */
+        foreach (ServiceLoader::load(EnvSourceProvider::class) as $envSourceProvider) {
+            $envSources[] = new LazyEnvSource($envSourceProvider->getEnvSource(...));
```
brettmc, line 32: so the order here is $_SERVER, php.ini, then any available dotenv implementation (and lazy so that they only process if a variable is requested and not found earlier)?

src/Config/SDK/Configuration/Environment/LazyEnvSource.php
ChrisLightfootWild: I think this one is correct - an EnvSource or a closure which should return an EnvSource 🤔

brettmc (approved):

PR #1556, opened 2025-04-02
chore(security): uses pinned versions of actions

jcchavezs:


brettmc:
Hi @jcchavezs can you give some info on what this is about? Has it come from a higher-level task to apply to all SIGs, or were we doing something wrong? Are we pinning so that dependabot doesn't update our actions?

bobstrecansky:
just following up here @jcchavezs, thanks!

Issue #1535, opened 2025-03-26
SDK should NOT drop user.* attributes by default anymore


jlabedo:
Pull request #843 dropped user attributes by default, but the spec has changed and do not require sdk to drop these attributes.

See documentation change commit:
https://github.com/open-telemetry/semantic-conventions/commit/31b9301f5d154bf4dd32317c4d949b54434855b0

See current documentation: [enduser](https://github.com/open-telemetry/semantic-conventions/blob/c1d8d371e8ad7cf47b50b6e0dd976746acdaeddf/docs/attributes-registry/enduser.md) and [user](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/user.md)

ycchuang99:
@bobstrecansky I can take this up.

Before starting just want to confirm whether this modification is to add a "Deprecate" annotation to the deprecated content.

PR #1565, opened 2025-04-15
deprecate filtered attributes and reference

ycchuang99:
Deprecate FilterAttributes class and its related functions.

Closes: #1535


brettmc:
Hi @ycchuang99 sorry I hadn't seen the related issue. Since those classes are all marked `@internal`, they should not be being directly used by anybody and we don't make any stability guarantees. So, they can be removed entirely.

ycchuang99:
Hi @brettmc Thanks for the clarification! I'll go ahead and remove them. Appreciate the quick response!

brettmc (approved):

Issue #1567, opened 2025-04-15
[spec 1.44.0] Stabilize GetAll in TextMap Extract


brettmc:
https://github.com/open-telemetry/opentelemetry-specification/pull/4472

ycchuang99:
Hi @brettmc 

I like to work on this issue. Can I take this up?

PR #1570, opened 2025-04-22
implement getAll function in TextMap Extract

ycchuang99:
Closes: #1567

### DIFF
--- a/src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
+++ b/src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
@@ -40,4 +40,20 @@ final class SanitizeCombinedHeadersPropagationGetter implements PropagationGette
             $value,
         );
     }
+
+    public function getAll($carrier, string $key): array
+    {
+        $value = $this->getter->getAll($carrier, $key);
+        if ($value === []) {
+            return [];
+        }
+
+        $value = preg_replace(
+            [self::SERVER_CONCAT_HEADERS_REGEX, self::TRAILING_LEADING_SEPARATOR_REGEX],
+            [self::LIST_MEMBERS_SEPARATOR],
+            $value,
+        );
+
+        return array_values($value);
+    }
 }
--- a/src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
+++ b/src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
@@ -12,7 +12,7 @@ use function preg_replace;
  * handle edge cases where the header has a trailing ';' or an empty trace state.
  * We also need to trim trailing separators from the header, found when a header is empty.
  */
-final class SanitizeCombinedHeadersPropagationGetter implements PropagationGetterInterface
+final class SanitizeCombinedHeadersPropagationGetter implements ExtendedPropagationGetterInterface
 {
     private const LIST_MEMBERS_SEPARATOR = ',';
     private const SERVER_CONCAT_HEADERS_REGEX = '/;(?=[^,=;]*=|$)/';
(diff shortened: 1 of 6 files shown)


brettmc:
Can you think about this part of the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#getall

> For many language implementations, the GetAll function will be added after the stable release of Getter. For these languages, requiring implementations of Getter to include GetAll constitutes a breaking change since instrumentation which previously functioned would fail. Language implementations should be cognizant of this, and add GetAll in a way that retains backwards compatibility. For example, by providing a default GetAll implementation based on Get, or by creating an extended Getter type.

It's usually helpful to check Java's implementation for inspiration :)

ycchuang99:
Hi @brettmc , would you give me some advice.

I'm currently working on adding a `getAll()` method via `ExtendedPropagationGetterInterface`, but I've run into an issue with `OpenTelemetry\Context\Propagation\SanitizeCombinedHeadersPropagationGetter`.

This class uses `PropagationGetterInterface` in its constructor, and changing it to `ExtendedPropagationGetterInterface` could introduce a breaking change. At the same time, I can't guarantee that the provided instance will implement `getAll()`.

```php
final class SanitizeCombinedHeadersPropagationGetter implements PropagationGetterInterface
{
    public function __construct(private readonly PropagationGetterInterface $getter)
    {
    }
}
```

Would you recommend leaving `SanitizeCombinedHeadersPropagationGetter` as is, or adding a fallback in `getAll()` to return an empty array if the constructor argument doesn't implement `ExtendedPropagationGetterInterface`?

Or do you have any other suggestions?

brettmc:
> This class uses PropagationGetterInterface in its constructor, and changing it to ExtendedPropagationGetterInterface could introduce a breaking change.

Excellent question. I think the safest way is to continue to accept a `PropagationGetterInterface`, have the class implement `ExtendedPropagationGetterInterface`, and put a check in the `getAll` method: "if $this->getter implements ExtendedPropagationGetterInterface, call it, else return empty array`. WDYT?

ycchuang99:
Thanks! I think that makes sense, I’ll do it this way.

brettmc (approved):

src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
```diff
@@ -40,4 +40,24 @@ public function get($carrier, string $key): ?string
...
         );
     }
+
+    public function getAll($carrier, string $key): array
+    {
+        if (! $this->getter instanceof ExtendedPropagationGetterInterface) {
+            return [];
+        }
+
+        $value = $this->getter->getAll($carrier, $key);
```
Nevay, line 50:
```suggestion
        $value = $this->getter instanceof ExtendedPropagationGetterInterface
            ? $this->getter->getAll($carrier, $key)
            : (array) $this->getter->get($carrier, $key);
```

Nevay (changes requested):

src/Context/Propagation/SanitizeCombinedHeadersPropagationGetter.php
ycchuang99: Hi @Nevay, thanks for the suggestion! Just to confirm—do you think it's better to return `$this->getter->get()` when `ExtendedPropagationGetterInterface` is not implemented, instead of returning an empty array?

As mentioned in a previous comment, I feel that returning an empty array might be a better choice for me, so that `getAll()` behaves differently from `get()`.

brettmc: It is probably less surprising for a user to receive some value when there are values present for the key, let's go with this suggestion.

PR #1573, opened 2025-04-30
update resource detectors to align with spec changes

brettmc:
PR https://github.com/open-telemetry/opentelemetry-specification/pull/4461 clarifies that some attributes (service.name, telemetry.*) are mandatory. It also reserves a couple of well-known detectors names (container, host, process, service) and the attributes they should populate. Consolidate some of our resource detectors into the built-in ones:
- process-runtime into process, and deprecate
- os into host, and deprecate
- sdk-provided into sdk, and deprecate
- service.name attribute from environment into service detector

Add "mandatory attributes", which is currently service.name, and initially set this to "unknown_service:php" similar to Java's implementation. Ensure that the composite detector starts with mandatory attributes.

Ensure that "sdk" and "service" detectors are always included from ResourceInfoFactory.

The deprecated detectors are still available, but do not do anything.


bobstrecansky (approved):

src/SDK/Resource/ResourceInfoFactory.php
```diff
@@ -102,4 +85,12 @@ public static function emptyResource(): ResourceInfo
 
         return self::$emptyResource;
     }
+
+    public static function mandatoryResource(): ResourceInfo
+    {
+        return ResourceInfo::create(Attributes::create(
+            [
+            ResourceAttributes::SERVICE_NAME => 'unknown_service:php'],
+        ));
```
ChrisLightfootWild, line 94:
```suggestion
        return ResourceInfo::create(Attributes::create([
            ResourceAttributes::SERVICE_NAME => 'unknown_service:php',
        ]));
```

ChrisLightfootWild (approved):

PR #1576, opened 2025-05-02
add suppressInstrumentation context key

brettmc:
following Java's lead, we can use this key in instrumentation libraries to suppress later spans, for example if we have multiple CLIENT spans, then the highest-level one can stop lower CLIENT spans from being created

### DIFF
--- a/src/Context/ContextKeys.php
+++ b/src/Context/ContextKeys.php
@@ -22,4 +22,11 @@ final class ContextKeys
 
         return $instance ??= Context::createKey('opentelemetry-trace-baggage-key');
     }
+
+    public static function suppressInstrumentation(): ContextKeyInterface
+    {
+        static $instance;
+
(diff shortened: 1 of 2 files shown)


brettmc:
Converting to draft until I do some tests to confirm whether this idea works.

Nevay:
We'll likely want a more fine-grained solution for instrumentation suppression, see also https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api%2Fsrc%2Fmain%2Fjava%2Fio%2Fopentelemetry%2Finstrumentation%2Fapi%2Finstrumenter%2FSpanSuppressionStrategy.java. AFAIR the generic suppression context key is only used to suppress telemetry emitted by SDK conponents (which we suppress via `HookManager::disable()`).

brettmc:
Closing, refer to #1579

Issue #1563, opened 2025-04-14
stabilize logger.enabled


brettmc:
https://github.com/open-telemetry/opentelemetry-specification/issues/4208 from spec [1.44.0](https://github.com/open-telemetry/opentelemetry-specification/pull/4476)

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/api.md#enabled

> To help users avoid performing computationally expensive operations when generating a LogRecord, a Logger SHOULD provide this Enabled API.
The API SHOULD accept the following parameters:
The [Context](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/context/README.md) to be associated with the LogRecord. When implicit Context is supported, then this parameter SHOULD be optional and if unspecified then MUST use current Context. When only explicit Context is supported, accepting this parameter is REQUIRED.
[Severity Number](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/data-model.md#field-severitynumber) (optional)
This API MUST return a language idiomatic boolean type. A returned value of true means the Logger is enabled for the provided arguments, and a returned value of false means the Logger is disabled for the provided arguments.
The returned value is not always static, it can change over time. The API SHOULD be documented that instrumentation authors needs to call this API each time they [emit a LogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.44.0/specification/logs/api.md#emit-a-logrecord) to ensure they have the most up-to-date response.

PR #1580, opened 2025-05-05
stabilize logger.enabled

brettmc:
- update Logger.isEnabled signature to include context and severity number, per spec
- remove experimental tag from isEnabled
- accept event_name on a LogRecord
- add event name to otlp logs payload
- update or remove some comments that were based on the old logs-bridge specification

Closes: #1563

### DIFF
--- a/src/API/Logs/LoggerInterface.php
+++ b/src/API/Logs/LoggerInterface.php
@@ -4,18 +4,15 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\API\Logs;
 
+use OpenTelemetry\Context\ContextInterface;
+
 interface LoggerInterface
 {
-    /**
-     * This method should only be used when implementing a `log appender`
-     * @see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.32.0/specification/logs/bridge-api.md#artifact-naming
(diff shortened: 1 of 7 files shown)


src/API/Logs/LoggerInterface.php
```diff
@@ -4,18 +4,15 @@
...
 
     /**
-     * Determine if the logger is enabled. Logs bridge API authors SHOULD call this method each time they
-     * are about to generate a LogRecord, to avoid performing computationally expensive work.
-     * @experimental
+     * Determine if the logger is enabled. Instrumentation authors SHOULD call this method each time they
+     * emit a LogRecord, to ensure they have the most up-to-date response.
      */
-    public function isEnabled(): bool;
+    public function isEnabled(?ContextInterface $context = null, ?int $severityNumber = null): bool;
```
ChrisLightfootWild, line 17: Changing this would break backwards compatibility; should this PR target `2.x` instead?

brettmc: Since it was marked as `@experimental`, I _think_ that breaking is allowable.

ChrisLightfootWild (approved):

Issue #1595, opened 2025-05-16
OTLP Exporter - Protobuf serializer deprecation error


barell:
As part of protobuf [latest release v31.0](https://github.com/protocolbuffers/protobuf/releases/tag/v31.0), there is a new deprecation introduced in PHP implementation.

The deprecation is described [here](https://github.com/protocolbuffers/protobuf/commit/d82de8f147be07e7a10c5c395505f0a662a911f2).

The deprecated method is used in `open-telemetry/exporter-otlp` in `ProtobufSerializer` class.

The fix is to replace `getLabel()` with one of the methods which are available instead.

barell:
Closing as already resolved.

PR #1596, opened 2025-05-16
Use Protobuf new dedicated method instead of deprecated getLabel

barell:
#1595


Nevay:
Should already be fixed by #1593, `::traverseDescriptor()` won't be called for protobuf `^4.31`.

brettmc:
@barell which version of protobuf did the new method arrive in? If it's >= 4.31 then that code won't be called as mentioned above, but if it's available in earlier versions then we can consider this PR...

barell:
Sorry for confusion. Turned out I wasn't actually using latest `open-telemetry/exporter-otlp`. 

TLDR: We use an internal "wrapper" lib for auto-instrumentation and I didn't do composer update with the `-W` flag inside the actual application repo.

PR #1614, opened 2025-05-29
Add utility classes and tests for Timer and Attribute tracking by id …

GrzegorzDrozd:
…and object

Introduced `TimerTrackerById`, `TimerTrackerByObject`, `AttributeTrackerById`, and `AttributeTrackerByObject` classes for managing timers and attributes based on id or object references.

This will be used by auto instrumentation and metrics in contrib modules.

### DIFF
--- a/src/SDK/Metrics/Util/TimerTrackerById.php
+++ b/src/SDK/Metrics/Util/TimerTrackerById.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Metrics\Util;
+
+class TimerTrackerById
+{
+    protected array $timers = [];
+
+    public function start(string|int $id): void
+    {
+        $this->timers[$id] = microtime(true);
+    }
+
+    public function durationMs(string|int $id): float
+    {
+        if (!isset($this->timers[$id])) {
+            return 0;
+        }
+
+        return (microtime(true) - $this->timers[$id]) * 1000;
+    }
+}
--- a/src/SDK/Metrics/Util/TimerTrackerByObject.php
+++ b/src/SDK/Metrics/Util/TimerTrackerByObject.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\SDK\Metrics\Util;
+
+class TimerTrackerByObject
+{
+    protected \WeakMap $timers;
+
+    public function __construct()
+    {
+        $this->timers = new \WeakMap();
+    }
+
+    public function start(object $id): void
+    {
+        $this->timers[$id] = microtime(true);
+    }
+
+    public function durationMs(object $id): float
+    {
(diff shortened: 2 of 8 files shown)


brettmc:
- I'd like to understand _how_ this would be used by an instrumentation package. The metrics implementation is pretty extensive,  does it already have features that can be used? I'd prefer to see a metrics implementation PR in an existing package, using this code locally, before extracting it to a re-usable place
- would we have a need for metrics to be tracked across different instrumentation packages?
- instrumentation packages are actively discouraged (per spec) from depending on the SDK

ChrisLightfootWild:
For any time related operations, there's also already [clock](https://github.com/opentelemetry-php/api/blob/1.3.0/Common/Time/Clock.php) functionality in the API - so it might make sense to utilise that over the raw calls to microtime.

GrzegorzDrozd:
@brettmc Usage example is here: https://github.com/open-telemetry/opentelemetry-php-contrib/pull/391

GrzegorzDrozd:
Hi @brettmc , somehow I missed this notification :/

> * I'd like to understand _how_ this would be used by an instrumentation package. The metrics implementation is pretty extensive,  does it already have features that can be used? I'd prefer to see a metrics implementation PR in an existing package, using this code locally, before extracting it to a re-usable place

For example: TimerTrackerById would keep timers for everything that sending external requests: PDO, Redis, HttpClient, Curl, Guzzle etc. Everything that needs to get duration.

>     * would we have a need for metrics to be tracked across different instrumentation packages?

Instance of a tracker would be created in `::register` of a given contrib package. It would not be shared. 

>     * instrumentation packages are actively discouraged (per spec) from depending on the SDK
Well we can copy and paste those classes all over but that would also be a bad practice. 

Maybe some form of cross contrib code share? We could do some namespace shenanigans and have shared code just copied to a contrib module?
From this:
```json
  "autoload": {
    "psr-4": {
      "OpenTelemetry\\Contrib\\Instrumentation\\PDO\\": "src/"
    },
    "files": [
      "_register.php"
    ]
  },
```
to
```json
  "autoload": {
    "psr-4": {
      "OpenTelemetry\\Contrib\\Instrumentation\\PDO\\": "src/",
      "OpenTelemetry\\Contrib\\Instrumentation\\Shared\\": "shared/"
    },
    "files": [
      "_register.php"
    ]
  },
```
Or even put everything into one file and use `files` + copy into every contrib?

brettmc:
> Well we can copy and paste those classes all over but that would also be a bad practice.

Agreed, that sounds awful! The most likely place is its own package in contrib, or in our API. If it goes into API, then under `API\Instrumentation` feels like the right place - we already have some code there to support auto-instrumentation :thinking:

GrzegorzDrozd:
@brettmc Code is adjusted: files are moved, tests updated etc. I also replaced raw microtime with ClockInterface. I will adjust linked PR to PDO metrics tomorrow.

ChrisLightfootWild:
@bobstrecansky was this closed intentionally?

Issue #1554, opened 2025-04-02
[bug] test segfaults with xdebug 3.4.x


brettmc:
Our unit tests started to segfault about a week ago, which was narrowed down to tests/Unit/Context/ScopeTest.php (test_inactive_scope_detach) when generating coverage against xdebug 3.4.2

See upstream bug: https://bugs.xdebug.org/view.php?id=2332

Until the upstream bug is fixed, let's skip this test so that other work can progress.

brettmc:
I think this issue is still live in 3.4.3...maybe we should skip that test for `3.4.x`, or permanently with a todo + tracking issue?

brettmc:
Not tested yet, but it sounds like we're good from [3.4.5](https://github.com/xdebug/xdebug/releases/tag/3.4.5)

PR #1621, opened 2025-06-04
Allows API packages to specify file-based and env-based config

Nevay:
Moves `ComponentProvider` and related classes to API and adds env-based loader to allow API packages (instrumentations and propagators) to specify their configuration options.


ChrisLightfootWild:
Should I base #1523 on the back of this, given the environment changes here? :thinking: (I appreciate that my PR is a little stale, but I plan to pick it up again in future)

Nevay:
The two PRs should be mostly independent:
#1523 changes how `OpenTelemetry\SDK\Common\Configuration\Configuration` reads values
#1621 changes how values from `OpenTelemetry\SDK\Common\Configuration\Configuration` are read

ChrisLightfootWild:
Failing pipelines look to be related to Xdebug issues that Brett raised in #1554; still present in 3.4.3.

Noted that there is a 3.4.4 to address some issues, but looks like re-running the build hasn't picked up that extension version yet.

src/Config/SDK/Configuration/ComponentPlugin.php
```diff
@@ -2,20 +2,8 @@
 
 declare(strict_types=1);
 
-namespace OpenTelemetry\Config\SDK\Configuration;
+namespace OpenTelemetry\API\Configuration\Config;
```
brettmc, line 5: This might need a code comment pointing out that the namespace is deliberate, else somebody might "fix" it in the future?

ChrisLightfootWild (approved):

brettmc (approved):

PR #1629, opened 2025-06-12
update weaver to 0.15.2

brettmc:
when generating metrics semconv, I saw a bunch of unexpected changes to trace/resource attributes, which I assume is related to changes in weaver since the semconv version has not changed. Splitting these changes off into a separate PR to make the upcoming metrics one simpler.


src/SemConv/TraceAttributeValues.php
```diff
@@ -12,6 +12,48 @@ interface TraceAttributeValues
      * The URL of the OpenTelemetry schema for these values.
      */
     public const SCHEMA_URL = 'https://opentelemetry.io/schemas/1.32.0';
+    /**
+     * Any time before Activity.onResume() or, if the app has no Activity, Context.startService() has been called in the app for the first time.
+     *
+     * @see TraceAttributes::ANDROID_APP_STATE
+     */
+    public const ANDROID_APP_STATE_CREATED = 'created';
```
brettmc, line 20: A lot of this is not useful to PHP. We probably should have been more selective way back when we started generating semconv, but here we are. To remove them now would technically be BC, and although there are some new android, cpython etc semconvs in this PR, there are also existing ones, and I don't want to deal with it down at the level of individual items.

PR #1632, opened 2025-06-12
add Logger.Enabled event name parameter

brettmc:
extra parameter, "event name", added in spec 1.45.0 which is optional and development stability

see https://github.com/open-telemetry/opentelemetry-specification/blob/v1.45.0/specification/logs/api.md#enabled

I've left LogRecordProcessor.enabled as a todo, as we're not using it and it would require another interface to avoid BC break (plus, we're not using it and it's not marked stable)


brettmc:
I think the test failures are specific to xdebug 3.4.3, and a fix has just been released in https://xdebug.org/updates#x_3_4_4

ChrisLightfootWild (approved):

Issue #1613, opened 2025-05-29
[semconv] Add metrics to PHP Semantic Conventions generation


smaddock:
**Describe the solution you'd like**
Currently `make semconv` only generates semantic convention attributes and attribute values for Resources and Traces. We should be able to take advantage of PHP constants for Metrics as well.

I think this might just require adding `semconv_metrics` and `semconv_grouped_metrics` filters to [weaver.yaml](script/semantic-conventions/templates/registry/php/weaver.yaml) and a maybe a couple Jinga templates? Although those said they were auto-generated.

**Describe alternatives you've considered**
Manually copy-pasting metric names from the docs 😱 

**Additional context**
- https://github.com/open-telemetry/semantic-conventions
- https://github.com/open-telemetry/weaver
- https://github.com/opentelemetry-php/sem-conv

brettmc:
I've got this nearly ready. There were some unexpected changes to other semconv output when updating weaver version, so I've first submitted #1629 to reduce noise for the PR for this issue.

PR #1633, opened 2025-06-13
adding metrics semconv

brettmc:
Generate some initial metrics SemConv files. I've chosen to start small, with just `http` and `otel` groups until we need more, and thought one file per group seemed neater.

Closes: #1613


brettmc:
When this is merged, I think we should create an issue to update our metrics generation from span and log processors to use the otel semconv attributes. They didn't exist when we first implemented.

brettmc:
Any opinions on how I've split each group into its own file?

ChrisLightfootWild:
> Any opinions on how I've split each group into its own file?

I think I prefer it this way :+1:

I'm not entirely sure on the units of measure though, although I see they are coming directly from the schemas - so maybe not much we could do with them. e.g. `Unit: {connection}`

Nevay:
> Any opinions on how I've split each group into its own file?

The [spec](https://opentelemetry.io/docs/specs/semconv/non-normative/code-generation/#semantic-conventions-artifact-structure) suggests a similar structure:
> ├── SchemaUrls.code
├── attributes
│   ├── ClientAttributes.code
│   ├── HttpAttributes.code
│   └── ...
├── metrics
│   ├── HttpMetrics.code
│   └── ...
└── events
    └── ...

PR #1666, opened 2025-07-17
upgrade psalm and fix complaints

brettmc:
To fix up the big list of psalm complaints in our github actions, upgrade psalm to latest. This fixes the issues complaining about repeated attributes, but also has some new things to complain about:
- add Override attributes as required (using rector and psalm to do the work - psalm picked up a bunch that rector missed)
- add symfony/polyfill-php83 to provide Override for earlier PHP versions
- add symfony/polyfill-php84 for good measure
- be explicit about operations on int|float (if either operand is a float, cast both to float)

### DIFF
--- a/src/Config/SDK/Configuration/ConfigurationFactory.php
+++ b/src/Config/SDK/Configuration/ConfigurationFactory.php
@@ -27,6 +27,7 @@ use OpenTelemetry\Config\SDK\Configuration\Internal\ResourceCollection;
 use OpenTelemetry\Config\SDK\Configuration\Internal\TrackingEnvReader;
 use OpenTelemetry\Config\SDK\Configuration\Loader\YamlExtensionFileLoader;
 use OpenTelemetry\Config\SDK\Configuration\Loader\YamlSymfonyFileLoader;
+use RuntimeException;
 use function serialize;
 use function sprintf;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
@@ -107,7 +108,11 @@ final class ConfigurationFactory
         }
 
         $loader = new ConfigurationLoader($resources);
-        $locator = new FileLocator(getcwd());
+        $cwd = getcwd();
+        if ($cwd === false) {
+            throw new RuntimeException('Current working directory could not be determined.');
+        }
+        $locator = new FileLocator($cwd);
         $fileLoader = new DelegatingLoader(new LoaderResolver([
             new YamlSymfonyFileLoader($loader, $locator),
             new YamlExtensionFileLoader($loader, $locator),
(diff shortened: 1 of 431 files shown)


bobstrecansky (approved):

src/Contrib/Otlp/ProtobufSerializer.php
```diff
@@ -30,6 +31,7 @@
  */
 final class ProtobufSerializer
 {
+    use LogsMessagesTrait;
     private function __construct(private readonly string $contentType)
```
ChrisLightfootWild, line 35:
```suggestion
    use LogsMessagesTrait;

    private function __construct(private readonly string $contentType)
```

src/Config/SDK/Configuration/ConfigurationFactory.php
```diff
@@ -107,7 +108,11 @@ public function parseFile(
         }
 
         $loader = new ConfigurationLoader($resources);
-        $locator = new FileLocator(getcwd());
+        $cwd = getcwd();
+        if ($cwd === false) {
+            throw new RuntimeException('Current working directory could not be determined.');
+        }
```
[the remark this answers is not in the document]
brettmc: I think it would have been pretty fatal anyway, since `FileLocator` is type-hinted for `string|array`, and `getcwd()` can return false.

deptrac/stubs/Override.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+//stub Override attribute for PHP <8.3
+
+namespace {
+    if (!class_exists('Override')) {
+        #[\Attribute(\Attribute::TARGET_METHOD)]
+        final class Override {}
```
Nevay, line 8: Can this be removed as we already require `symfony/polyfill-php83`?

brettmc: Yes, it can. I had to be quite specific about paths, as deptrac tries to analyze the polyfills, and have just included polyfill-php83 for now. Also renamed stubs to polyfills for clarity.

src/SDK/Metrics/Aggregation/ExplicitBucketHistogramAggregation.php
```diff
@@ -56,10 +58,13 @@ public function record($summary, $value, AttributesInterface $attributes, Contex
...
      * @param ExplicitBucketHistogramSummary $right
      */
+    #[\Override]
     public function merge($left, $right): ExplicitBucketHistogramSummary
     {
         $count = $left->count + $right->count;
-        $sum = $left->sum + $right->sum;
+        $sum = (is_int($left->sum) && is_int($right->sum))
+            ? ($left->sum + $right->sum)
+            : ((float) $left->sum + (float) $right->sum);
```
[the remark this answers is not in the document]
brettmc: reverted these changes and ignored in deptrac for `src/SDK/Metrics/Aggregation/*`

Issue #1687, opened 2025-08-11
Error in DebugScope "missing call to Scope::detach()" notice


red-led:
In `DebugScope::detach()` code `spl_object_id($this)` used to indicate wich scope got mishandled.

But in `DebugScope::__destruct()` code `spl_object_id($this->scope)` is used, wich is confusing.

PR #1688, opened 2025-08-11
update DebugScope error message

brettmc:
Make the scope id in error messages consistent between the detached-with-error and not-detached cases, by using spl_object_id($this->scope) in the former to match the latter.

Closes: #1687

### DIFF
--- a/src/Context/DebugScope.php
+++ b/src/Context/DebugScope.php
@@ -49,18 +49,18 @@ final class DebugScope implements ScopeInterface
         if (($flags & ScopeInterface::DETACHED) !== 0) {
             trigger_error(sprintf(
                 'Scope: unexpected call to Scope::detach() for scope #%d, scope was already detached %s',
-                spl_object_id($this),
+                spl_object_id($this->scope),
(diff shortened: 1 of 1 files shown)


src/Context/DebugScope.php
```diff
@@ -49,18 +49,18 @@ public function detach(): int
         if (($flags & ScopeInterface::DETACHED) !== 0) {
             trigger_error(sprintf(
                 'Scope: unexpected call to Scope::detach() for scope #%d, scope was already detached %s',
-                spl_object_id($this),
+                spl_object_id($this->scope),
```
red-led, line 52: I think using `$this` would be more correct, cause as API user you see `DebugScope` as actual scope.

If you got some bug and start logging `spl_object_id($scope)`, you will get different id's and wold not be able to get sence of what's going on.

brettmc: yes, that makes sense. Thanks for the feedback, updated.

bobstrecansky (approved):

Nevay (approved):

PR #1718, opened 2025-09-25
Adding Azure Storage Queue as messaging value system

LuMaterna:


brettmc:
Welcome, @LuMaterna . Our semantic conventions are generated code based on the upstream https://github.com/open-telemetry/semantic-conventions, and so we don't accept direct modifications.
You can try to have the value added upstream.

PR #1766, opened 2025-11-04
fix: send min and max histogram data

rachelyangdog:
- small fix to send min and max histogram data


brettmc:
@rachelyangdog thanks for this contribution. There are a couple of tests which are failing and need to be updated since the output now includes min/max. Otherwise, lgtm!

bobstrecansky:
Sorry, approved too quickly - agree with @brettmc that the tests need to be fixed.

rachelyangdog:
@bobstrecansky Hi, thanks for looking at my PR so quickly. Let me know if there is anything else I need to update. Thanks so much!!

PR #1768, opened 2025-11-08
chore: Added comments to the duplicate event.name code in EventLogger

jerrytfleung:
Fixed duplicated code found in the emit function.

UPDATES: It is not duplicate code. It turns out it is intended to get the first placeholder for `event.name` in attributes so it won't be dropped due to the `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` limitation scenario.

### DIFF
--- a/tests/Unit/SDK/Logs/EventLoggerTest.php
+++ b/tests/Unit/SDK/Logs/EventLoggerTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\Tests\Unit\SDK\Logs;
 
+use OpenTelemetry\API\Behavior\Internal\Logging;
 use OpenTelemetry\API\Common\Time\Clock;
 use OpenTelemetry\API\Common\Time\TestClock;
 use OpenTelemetry\API\Logs\LoggerInterface;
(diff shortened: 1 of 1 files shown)


ChrisLightfootWild (approved):

jerrytfleung:
> This was done on purpose to ensure that we don't drop the `event.name` attribute.

Previously, `event.name` attribute could be replaced if `$attributes` contains another `event.name` in `setAttributes` call
```
$logRecord->setAttributes($attributes);
```
While the code set the `event.name` again after the `setAttributes` call to ensure we don't lose that attribute, so I think we don't need the first `event.name` set attributes calls. (The later always wins)

brettmc:
> This was done on purpose to ensure that we don't drop the `event.name` attribute.

I have a memory of this, too. There is a [unit test for it](https://github.com/open-telemetry/opentelemetry-php/blob/main/tests/Unit/SDK/Logs/EventLoggerTest.php#L81), and if that is passing then I think this change is OK now.

Nevay:
The linked test verifies that the `$name` parameter takes precedence over an `event.name` attribute provided in `$attributes` (= the second `$logRecord->setAttribute('event.name', $name)`).

The first `$logRecord->setAttribute('event.name', $name)` ensures that `event.name` is the first attribute entry -> is not dropped by attribute limits (unless `0`).

```php
#[BackupGlobals(true)]
public function test_event_name_is_not_dropped_first(): void
{
	$_SERVER['OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT'] = 1;
	$eventLoggerProvider = new EventLoggerProvider((new LoggerProviderBuilder())
		->addLogRecordProcessor(new SimpleLogRecordProcessor($exporter = new InMemoryExporter()))
		->build());
	$eventLogger = $eventLoggerProvider->getEventLogger('test');

	Logging::disable();
	try {
		$eventLogger->emit('my.event', attributes: ['other.attribute' => 'not.my.event']);
	} finally {
		Logging::reset();
	}
	$eventLoggerProvider->forceFlush();

	/** @var list<ReadableLogRecord> $logs */
	$logs = $exporter->getStorage()->getArrayCopy();
	$this->assertCount(1, $logs);
	$this->assertArrayHasKey('event.name', $logs[0]->getAttributes()->toArray());
}
 ```

jerrytfleung:
Closing this PR as the function call is intended to set the first entry of the attributes so that it won't get dropped due to `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` scenario

ChrisLightfootWild:
Does this warrant adding a comment to the existing code to explain the scenario that is being covered by this?

brettmc:
@jerrytfleung thanks for the update. Could you also include the test Nevay provided above in this PR?

PR #1775, opened 2025-11-14
Composer backwards compatibility for psr/log

ChrisLightfootWild:
Alternative implementation to: https://github.com/open-telemetry/opentelemetry-php/pull/1727

### DIFF
--- a/src/API/_compat.php
+++ b/src/API/_compat.php
@@ -1,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OpenTelemetry\API;
+
+use function class_exists;
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
+
+if (
+    // Provide a backwards-compatible Psr3 implementation when running under composer.
+    class_exists(InstalledVersions::class, false)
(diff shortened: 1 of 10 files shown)


ChrisLightfootWild:
I mentioned that I had an alternative PR to #1727 in today's SIG, and it was suggested to tag @brettmc & @Nevay for some thoughts on this.

I'm totally okay with just closing this if it doesn't seem like the right approach though!

src/API/_compat.php
```diff
@@ -0,0 +1,17 @@
...
+namespace OpenTelemetry\API;
+
+use function class_exists;
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
+
+if (
+    // Provide a backwards-compatible Psr3 implementation when running under composer.
+    class_exists(InstalledVersions::class, false)
+    && InstalledVersions::satisfies(new VersionParser(), 'composer/composer', '~2.0')
```
Nevay: Should check whether we are running as composer command first; `VersionParser` might not be available otherwise.

ChrisLightfootWild:
Closing in favour of https://github.com/open-telemetry/opentelemetry-php/pull/1727

PR #1864, opened 2026-01-21
Fix missing Symfony ^8.0 dependency declaration

rogierknoester:
Symfony 8 support has been added earlier but the Symfony dotenv provider did not yet declare that it supports `symfony/dotenv:^8.0`. This means that otel setups that rely on loading the configuration from a `.env.dev` won't work as the Symfony dotenv provider won't be one included. This addition should fix that.


bobstrecansky (approved):

ruudk:
I wonder how this issue could be prevented when SF 9 is released in a few years? It will probably give the same issue.

PR #1874, opened 2026-01-24
Delete Jaeger propagator

Nevay:
https://github.com/open-telemetry/opentelemetry-specification/pull/4827


Nevay:
Should we instead delete the propagator and mark the package as abandoned?

brettmc:
> Should we instead delete the propagator and mark the package as abandoned?

I'm comfortable with doing that, since it's [basically unused](https://packagist.org/packages/open-telemetry/extension-propagator-jaeger/stats)

PR #1927, opened 2026-03-25
Fix WithSpanHandler: remove hardcode, add constants instead

mdomansky:
I have a list of containers:
1 - my php app
2 - otel-collector
3 - opensearch

While using `#[WithSpan]` attribute on methods on my `php` container, all span information is successfully sent to the `otel-collector` container, but while the last sends the info to the `opensearch` database I get an exception:

```
error internal/base_exporter.go:115 Exporting failed. Rejecting data. Try enabling sending_queue to survive temporary failures. {"resource": {"[service.instance.id](http://service.instance.id/)": "c614aac8-2cfa-48f9-b109-a246a99f80ad", "[service.name](http://service.name/)": "otelcol-contrib", "service.version": "0.145.0"}, "[otelcol.component.id](http://otelcol.component.id/)": "opensearch", "otelcol.component.kind": "exporter", "otelcol.signal": "traces", "error": "not retryable error: Permanent error: {\"type\":\"mapper_parsing_exception\",\"reason\":\"object mapping for [attributes.code.function] tried to parse field [code.function] as object, but found a concrete value\",\"caused_by\":{\"type\":\"\",\"reason\":\"\",\"caused_by\":null}}", "rejected_items": 7}

```

My env:
```
OTEL_ENABLED: 'true'
OTEL_TRACES_EXPORTER: 'otlp'
```

Having researched, I found that "wrong" attribute names are hardcoded, so I updated them with constants.


ChrisLightfootWild (commented):
This is intentionally hardcoded to avoid the dependency on SemConv in the API package.

mdomansky:
> This is intentionally hardcoded to avoid the dependency on SemConv in the API package.

Why do I get the issue like `code.function` is wrong while `code.function.name` is correct?

ChrisLightfootWild:
> > This is intentionally hardcoded to avoid the dependency on SemConv in the API package.
> 
> Why do I get the issue like `code.function` is wrong while `code.function.name` is correct?

I believe `code.function` is correct for the emitted version of telemetry, `1.25.0`. See https://github.com/open-telemetry/semantic-conventions/blob/v1.25.0/model/general.yaml#L55-L65

Might be worth asking the otel collector devs on this one. The schema file links to https://github.com/open-telemetry/semantic-conventions/pull/1624 from the `1.30.0` semantic conventions. It contains the following:

```yaml
...trimmed
  1.30.0:
    all:
      changes:
...trimmed
        - rename_attributes:
            attribute_map:
              code.function: code.function.name
              code.filepath: code.file.path
              code.lineno: code.line.number
              code.column: code.column.number
```

My understanding is that we are emitting valid telemetry for the schema version we have declared, and that the collector should rename the attributes during ingestion.

bobstrecansky:
@open-telemetry/collector-approvers, can you weigh in here?

mx-psi:
The error on this PR's description seems specific to the opensearch exporter, feel free to file an issue on the opentelemetry-collector-contrib bug tracker and the code owners will take a look

mdomansky:
Updated the schema version and updated span attribute names according to the scheme. Left hardcoded attribute names.

src/API/Instrumentation/WithSpanHandler.php
```diff
@@ -36,10 +36,11 @@ public static function pre(mixed $target, array $params, ?string $class, string
             ->tracer()
             ->spanBuilder($name)
             ->setSpanKind($kind)
-            ->setAttribute('code.function', $function)
+            ->setAttribute('code.function.name', $function)
             ->setAttribute('code.namespace', $class)
```
Nevay:
```suggestion
            ->setAttribute('code.function.name', $class !== null ? $class . '::' . $function : $function)
```

See semconv [v1.31.0
](https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md#v1310)

src/API/Instrumentation/WithSpanHandler.php
```diff
@@ -36,10 +36,10 @@ public static function pre(mixed $target, array $params, ?string $class, string
...
             ->spanBuilder($name)
             ->setSpanKind($kind)
-            ->setAttribute('code.function', $function)
-            ->setAttribute('code.namespace', $class)
-            ->setAttribute('code.filepath', $filename)
-            ->setAttribute('code.lineno', $lineno)
+            ->setAttribute('code.function.name', $class !== null ? $class . '::' . $function : $function)
+            ->setAttribute('code.file.path', $filename)
+            ->setAttribute('code.line.number', $lineno)
+            ->setAttribute('code.column.number', $lineno)
```
ChrisLightfootWild: This doesn't look right. I don't think we should use the line number as column number.

mdomansky: Thanks. Fixed.

src/API/Instrumentation/WithSpanHandler.php
```diff
@@ -36,10 +36,9 @@ public static function pre(mixed $target, array $params, ?string $class, string
             ->tracer()
             ->spanBuilder($name)
             ->setSpanKind($kind)
-            ->setAttribute('code.function', $function)
-            ->setAttribute('code.namespace', $class)
-            ->setAttribute('code.filepath', $filename)
-            ->setAttribute('code.lineno', $lineno)
+            ->setAttribute('code.function.name', $class !== null ? $class . '::' . $function : $function)
```
Nevay: Could unify this with the span `$name` logic above:
```php
$functionName = $class !== null ? $class . '::' . $function : $function;
$name = $span_args['name'] ?? $functionName;
```

Nevay (approved):

src/API/Instrumentation/WithSpanHandler.php
mdomansky: Updated to the `->setAttribute('code.function.name', $name)` due to `$name` setting above correct way I believe.

Nevay: We cannot use `$name` for `code.function.name`, it might be overridden via `$span_args`.

Using the snippet above:
`$name` -> span name provided to `->spanBuilder()`
`$functionName` -> `code.function.name` attribute

mdomansky: Updated

intuibase (approved):

brettmc:
It looks like all reviewers are happy with this change, but we're blocked on `zizmor` results... I'll close and reopen this PR, which should re-run all actions.

brettmc:
@mdomansky sorry this has been waiting so long. I've just kicked off the checks and they're failing, I think a rebase will probably fix it given the age of this PR.

PR #1935, opened 2026-04-09
feat: limit OTLP HTTP response body to 4 MiB

bobstrecansky:
Prevents unbounded memory usage when a misconfigured or malicious collector sends an oversized response body.

- Add ResponseBodySizeLimit::MAX_BYTES constant (4 MiB)
- Add PsrUtils::readBodyWithSizeLimit() and decode()
- Update PsrTransport::handleResponse() to use the size-capped reader
- Add TransportResponseException for typed HTTP error handling
- Add ResponseBodySizeLimitTest (10 test cases) 

Fixes #1932

### DIFF
--- a/src/SDK/Common/Export/Http/PsrTransport.php
+++ b/src/SDK/Common/Export/Http/PsrTransport.php
@@ -4,155 +4,137 @@ declare(strict_types=1);
 
 namespace OpenTelemetry\SDK\Common\Export\Http;
 
-use function assert;
-use BadMethodCallException;
-use function explode;
-use function in_array;
-use OpenTelemetry\SDK\Common\Export\TransportInterface;
+use Http\Discovery\Psr17FactoryDiscovery;
 use OpenTelemetry\SDK\Common\Future\CancellationInterface;
 use OpenTelemetry\SDK\Common\Future\CompletedFuture;
 use OpenTelemetry\SDK\Common\Future\ErrorFuture;
 use OpenTelemetry\SDK\Common\Future\FutureInterface;
 use Psr\Http\Client\ClientInterface;
-use Psr\Http\Client\NetworkExceptionInterface;
 use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\StreamFactoryInterface;
-use RuntimeException;
-use function strtolower;
 use Throwable;
-use function time_nanosleep;
-use function trim;
 
 /**
- * @psalm-template CONTENT_TYPE of string
- * @template-implements TransportInterface<CONTENT_TYPE>
+ * PSR-7/PSR-18 HTTP transport for OTLP/HTTP exporters.
+ *
+ * ### Response body size limiting (issue #1932)
+ *
+ * All response body reads are funnelled through
+ * {@see PsrUtils::readBodyWithSizeLimit()}, which caps consumption at 4 MiB.
+ * This prevents a misconfigured or malicious collector from causing unbounded
+ * memory growth in the PHP process.
  */
-final class PsrTransport implements TransportInterface
+final class PsrTransport
 {
-    private bool $closed = false;
+    private ClientInterface $client;
+    private RequestFactoryInterface $requestFactory;
+    private StreamFactoryInterface $streamFactory;
+    private string $endpoint;
+    private string $contentType;
 
-    /**
-     * @psalm-param CONTENT_TYPE $contentType
-     */
+    /** @var array<string,string> */
+    private array $headers;
+    private string $compression;
+
     public function __construct(
-        private readonly ClientInterface $client,
-        private readonly RequestFactoryInterface $requestFactory,
-        private readonly StreamFactoryInterface $streamFactory,
-        private readonly string $endpoint,
-        private readonly string $contentType,
-        private readonly array $headers,
-        private readonly array $compression,
-        private readonly int $retryDelay,
-        private readonly int $maxRetries,
+        ClientInterface $client,
+        RequestFactoryInterface $requestFactory,
+        StreamFactoryInterface $streamFactory,
+        string $endpoint,
+        string $contentType,
+        array $headers = [],
+        string $compression = 'none'
     ) {
-    }
-
-    #[\Override]
-    public function contentType(): string
-    {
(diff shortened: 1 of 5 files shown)


src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -4,155 +4,137 @@
...
+
+    // -------------------------------------------------------------------------
+    // Private helpers
+    // -------------------------------------------------------------------------
+
+    private function buildRequest(string $payload): RequestInterface
+    {
+        $body = $payload;
+
+        if ($this->compression === 'gzip') {
```
Nevay, line 87:
```suggestion
        if ($this->compression === 'gzip' && extension_loaded('zlib')) {
```

Same for the other `if ($this->compression === 'gzip')` below; can we merge these two ifs?

Alternatively check in constructor.

src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -4,155 +4,137 @@
...
-        private readonly array $compression,
-        private readonly int $retryDelay,
-        private readonly int $maxRetries,
+        ClientInterface $client,
+        RequestFactoryInterface $requestFactory,
+        StreamFactoryInterface $streamFactory,
+        string $endpoint,
+        string $contentType,
+        array $headers = [],
+        string $compression = 'none'
```
Nevay, line 48:
```suggestion
        ?string $compression = null,
```

src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -4,155 +4,137 @@
...
-        if ($appliedEncodings) {
-            $request = $request->withHeader('Content-Encoding', $appliedEncodings);
-        }
-        foreach ($this->headers as $header => $value) {
-            $request = $request->withAddedHeader($header, $value);
-        }
+            ->withBody($stream)
+            ->withHeader('Content-Type', $this->contentType);
 
-        for ($retries = 0;; $retries++) {
```
Nevay, line 80: Is the removal of retries intended?

src/SDK/Common/Export/Http/PsrUtils.php
```diff
@@ -4,178 +4,101 @@
...
-    public static function encode(string $value, array $encodings, ?array &$appliedEncodings = null): string
-    {
-        for ($i = 0, $n = count($encodings); $i < $n; $i++) {
-            if (!$encoder = self::encoder($encodings[$i])) {
-                unset($encodings[$i]);
-
-                continue;
-            }
+        // (3) Read at most $maxRead bytes.
+        $body = $response->getBody()->read($maxRead);
```
Nevay, line 60: Has to call `read` in a loop.
> Fewer than $length bytes may be returned if underlying stream call returns fewer bytes.

src/SDK/Common/Export/Http/PsrUtils.php
```diff
@@ -4,178 +4,101 @@
...
+     *
+     * @return string Decoded payload bytes, ready for proto-unmarshalling.
      */
-    public static function decode(string $value, array $encodings): string
+    public static function decode(ResponseInterface $response): string
     {
-        if ($value === '') {
-            return $value;
-        }
+        $body = self::readBodyWithSizeLimit($response);
```
Nevay, line 75: Should decompress the response body stream while reading it instead of buffering its content and then decompressing the entire body at once.

src/SDK/Common/Export/Http/PsrUtils.php
```diff
@@ -4,178 +4,101 @@
...
+                throw new \RuntimeException(
+                    'Failed to gzip-decode OTLP response body.'
+                );
             }
 
-            /** @phan-suppress-next-line PhanPossiblyUndeclaredVariable */
-            if ($result === false) {
-                throw new LogicException();
-            }
+            return $decoded;
```
Nevay, line 99: The limit must apply to the decompressed body, see [OTLP/HTTP Response](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response):
> The client MUST limit the size of the response body when parsing it, including after decompression, to mitigate possible excessive memory usage caused by a misconfigured or malicious server.

src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -4,155 +4,137 @@
...
+ *
+ * ### Response body size limiting (issue #1932)
+ *
+ * All response body reads are funnelled through
+ * {@see PsrUtils::readBodyWithSizeLimit()}, which caps consumption at 4 MiB.
+ * This prevents a misconfigured or malicious collector from causing unbounded
+ * memory growth in the PHP process.
  */
-final class PsrTransport implements TransportInterface
+final class PsrTransport
```
Nevay, line 29: This change seems unintended?

Nevay (changes requested):

src/SDK/Common/Export/Http/PsrTransport.php
```diff
@@ -4,155 +4,137 @@
...
-        private readonly ClientInterface $client,
-        private readonly RequestFactoryInterface $requestFactory,
-        private readonly StreamFactoryInterface $streamFactory,
-        private readonly string $endpoint,
-        private readonly string $contentType,
-        private readonly array $headers,
-        private readonly array $compression,
-        private readonly int $retryDelay,
-        private readonly int $maxRetries,
+        ClientInterface $client,
```
ChrisLightfootWild, line 42: Why has the constructor property promotion been removed here? 🤔

src/SDK/Common/Export/Http/TransportResponseException.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    private int $statusCode;
+    private string $responseBody;
+
+    public function __construct(int $statusCode, string $responseBody, string $message = '')
+    {
+        parent::__construct($message !== '' ? $message : sprintf('HTTP %d', $statusCode));
+        $this->statusCode = $statusCode;
+        $this->responseBody = $responseBody;
+    }
```
ChrisLightfootWild, line 26:
```suggestion
{
    public function __construct(
        private int $statusCode,
        private string $responseBody,
        string $message = '',
    ) {
        parent::__construct($message ?: sprintf('HTTP %d', $statusCode));
    }
```

PR #1937, opened 2026-04-18
Added PHP 8.5 to build matrix for docker base images.

ChrisLightfootWild:
- Added 8.5 to build matrix
 - Alphabetical sort of php packages in Dockerfile

### DIFF
--- a/.github/workflows/publish-otel-php-base-docker-image.yml
+++ b/.github/workflows/publish-otel-php-base-docker-image.yml
@@ -16,7 +16,6 @@ permissions:
 jobs:
   push_to_registry:
     name: OpenTelemetry PHP base docker image creation
-    continue-on-error: true
     strategy:
       matrix:
         php-version:
--- a/.github/workflows/publish-otel-php-base-docker-image.yml
+++ b/.github/workflows/publish-otel-php-base-docker-image.yml
@@ -16,9 +16,15 @@ permissions:
 jobs:
   push_to_registry:
     name: OpenTelemetry PHP base docker image creation
+    continue-on-error: true
     strategy:
       matrix:
-        php-version: ['8.1', '8.2', '8.3', '8.4']
+        php-version:
+          - "8.1"
+          - "8.2"
+          - "8.3"
+          - "8.4"
+          - "8.5"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-otel-php-base-docker-image.yml
```diff
@@ -16,9 +16,15 @@ permissions:
 jobs:
   push_to_registry:
     name: OpenTelemetry PHP base docker image creation
+    continue-on-error: true
```
bobstrecansky: (question): Are we sure we want `continue-on-error` here?

docker/Dockerfile
```diff
@@ -1,41 +1,45 @@
 FROM composer:2.9@sha256:743aebe48ca67097c36819040633ea77e44a561eca135e4fc84c002e63a1ba07 AS composer
 FROM debian:trixie@sha256:55a15a112b42be10bfc8092fcc40b6748dc236f7ef46a358d9392b339e9d60e8
+
+ARG PHP_VERSION=8.3
```
bobstrecansky: (nit): What was the reasoning behind choosing 8.3 as a default here?

bobstrecansky (approved):
Two small questions, no blockers.

docker/Dockerfile
ChrisLightfootWild: It was 8.3 already, I just moved it in the file. We can change it as part of this too.

.github/workflows/publish-otel-php-base-docker-image.yml
ChrisLightfootWild: Happy to revert this one. Just any step failure was cancelling the sibling  workflows from the build matrix.

docker/Dockerfile
ChrisLightfootWild: Updated to current stable version, 8.5.

.github/workflows/publish-otel-php-base-docker-image.yml
ChrisLightfootWild: Reverted.

bobstrecansky (approved):

PR #1975, opened 2026-05-27
Enforce W3C Baggage size and entry-count caps in Parser

### LINKED ISSUE #1974: Baggage: extract has no size or entry-count cap (hardening)
### Summary

`BaggagePropagator::extract` in `open-telemetry/api` does not enforce
the W3C Baggage size limits when parsing an inbound `baggage` HTTP
header. The parser at [`src/API/Baggage/Propagation/Parser.php:25-66`](https://github.com/open-telemetry/opentelemetry-php/blob/74d43f20674aa0ec3a0af783edb0b5e4af8d0f8a/src/API/Baggage/Propagation/Parser.php#L25-L66)
walks the entire `explode(',', $this->baggageHeader)` result with no
cap on header byte size, no cap on entry count, and no per-entry
length cap. Every list-member triggers two further `explode` calls,
two `urldecode` allocations, two `str_replace` passes, and a
`BaggageBuilder::set` hash insert.

The W3C Baggage specification ([`#limits`](https://www.w3.org/TR/baggage/#limits))
recommends 180 list-members and 8192 bytes total; the PHP SDK enforces
neither.

This is the PHP sibling of the same hardening that landed in the other
SDKs:

- java (1.62.0): GHSA-rcgg-9c38-7xpx
- go: `maxMembers=64`, `maxBytes=8192`
- dotnet: `MaxBaggageLength=8192`, `MaxBaggageItems=180`
- cpp: `kMaxSize=8192`, `kMaxKeyValuePairs=180`
- rust: GHSA-w9wp-h8wv-79jx
- js: GHSA-8988-4f7v-96qf

### Affected versions

`open-telemetry/api` <= 1.9.0 and current `main`.

### Context

This was originally filed as a private security advisory
([GHSA-7f5h-mpj3-v3hf](https://github.com/open-telemetry/opentelemetry-php/security/advisories/GHSA-7f5h-mpj3-v3hf),
now closed). Per the project security policy, the maintainers prefer
to handle this through the standard issue/PR workflow as a hardening
bug rather than as a security vulnerability. Filing here as
requested.

### Reproduction

Against `composer require open-telemetry/api:1.9.0` on PHP 8.5:

```php
require __DIR__ . '/vendor/autoload.php';

use OpenTelemetry\API\Baggage\Propagation\BaggagePropagator;
use OpenTelemetry\Context\Propagation\ArrayAccessGetterSetter;

$entries = 200000;
$pairs = [];
for ($i = 0; $i < $entries; $i++) {
    $pairs[] = "k{$i}=v{$i}";
}
$header = implode(',', $pairs);

$carrier = ['baggage' => $header];
$startMemory = memory_get_usage(true);
$startTime = microtime(true);
BaggagePropagator::getInstance()->extract(
    $carrier,
    ArrayAccessGetterSetter::getInstance(),
    Context::getCurrent(),
);
printf(
    "elapsed_ms=%.1f delta_rss_kb=%.0f\n",
    (microtime(true) - $startTime) * 1000,
    (memory_get_usage(true) - $startMemory) / 1024,
);
```

A 200000-entry header (~3 MiB) takes ~1.5s of CPU and grows the
worker RSS by ~47 MiB on PHP 8.5.6. Linear scaling. With
`pm.max_children=4` a small handful of concurrent attacker requests
pin the entire FPM worker pool.

### Suggested fix

Mirror the caps already shipped by the other patched SDKs:

1. Add two private constants to `Parser`:

   ```php
   private const MAX_BAGGAGE_BYTES = 8192;
   private const MAX_BAGGAGE_ENTRIES = 180;
   ```

2. Reject the whole header up-front when
   `strlen($header) > MAX_BAGGAGE_BYTES` (matches cpp's `FromHeader`
   gate; avoids running the per-entry allocation loop at all).

3. Track an entry counter inside the parse `foreach` and `break` once
   `MAX_BAGGAGE_ENTRIES` is reached.

I will follow up with a PR shortly.

### References

- W3C Baggage limits: https://www.w3.org/TR/baggage/#limits
- Java sibling fix: https://github.com/open-telemetry/opentelemetry-java/security/advisories/GHSA-rcgg-9c38-7xpx
- JS sibling fix: https://github.com/open-telemetry/opentelemetry-js/security/advisories/GHSA-8988-4f7v-96qf
- Rust sibling fix: https://github.com/open-telemetry/opentelemetry-rust/security/advisories/GHSA-w9wp-h8wv-79jx
- Python (Jaeger) sibling fix: https://github.com/open-telemetry/opentelemetry-python/security/advisories/GHSA-f5m9-jfrj-jjj7

tonghuaroot:
Fixes #1974.

### What

`BaggagePropagator::extract` walked the entire `explode(',', $header)`
result with no cap on header byte size, no cap on entry count, and no
per-entry length cap. This change adds two private constants to
`Parser` and enforces them inside `parseInto`.

### Change

- `Parser::MAX_BAGGAGE_BYTES = 8192`
- `Parser::MAX_BAGGAGE_ENTRIES = 180`
- `parseInto` returns early when
  `strlen($header) > MAX_BAGGAGE_BYTES` (cheap, prevents the per-entry
  allocation loop running at all).
- `parseInto` `break`s once `MAX_BAGGAGE_ENTRIES` accepted entries
  have been processed.

Values are taken from the W3C Baggage spec
([`#limits`](https://www.w3.org/TR/baggage/#limits)) and match what
the other SDKs already do:

| SDK    | max bytes | max entries |
| ------ | --------- | ----------- |
| java (1.62.0) | 8192 | 180 |
| go     | 8192      | 64          |
| dotnet | 8192      | 180         |
| cpp    | 8192      | 180         |

### Tests

Two regression tests added in `tests/Unit/API/Baggage/Propagation/ParserTest.php`:

- `test_parse_into_rejects_header_exceeding_w3c_byte_limit` builds a
  ~10 KiB header from 1024 pairs and asserts the builder is never
  invoked.
- `test_parse_into_caps_entries_at_w3c_list_member_limit` builds a
  500-pair header (under 8192 bytes) and asserts the builder receives
  exactly 180 `set` calls.

`vendor/bin/phpunit tests/Unit/API/Baggage` passes 63/63 locally on
PHP 8.5.6.

### Static analysis

- `vendor/bin/phpstan analyse src/API/Baggage/Propagation/Parser.php`
  no errors.
- `vendor-bin/rector/vendor/bin/rector process ... --dry-run` no
  changes.
- `vendor-bin/php-cs-fixer/vendor/bin/php-cs-fixer fix ... --dry-run`
  no changes.

### Context

Originally filed as a private security advisory
([GHSA-7f5h-mpj3-v3hf](https://github.com/open-telemetry/opentelemetry-php/security/advisories/GHSA-7f5h-mpj3-v3hf),
now closed). Per the project security policy the maintainers prefer
to handle this through the standard issue/PR workflow as a hardening
bug.

### References

- W3C Baggage limits: https://www.w3.org/TR/baggage/#limits
- Java sibling: https://github.com/open-telemetry/opentelemetry-java/security/advisories/GHSA-rcgg-9c38-7xpx
- JS sibling: https://github.com/open-telemetry/opentelemetry-js/security/advisories/GHSA-8988-4f7v-96qf

### DIFF
--- a/src/API/Baggage/Propagation/Parser.php
+++ b/src/API/Baggage/Propagation/Parser.php
@@ -8,6 +8,7 @@ use function explode;
 use OpenTelemetry\API\Baggage\BaggageBuilderInterface;
 use OpenTelemetry\API\Baggage\Metadata;
 use function str_replace;
+use function strlen;
 use function trim;
 use function urldecode;
 
@@ -17,6 +18,10 @@ final class Parser
     private const EXCLUDED_VALUE_CHARS = [' ', '"', ',', ';', '\\'];
     private const EQUALS = '=';
 
+    /** @see https://www.w3.org/TR/baggage/#limits */
+    private const MAX_BAGGAGE_BYTES = 8192;
+    private const MAX_BAGGAGE_ENTRIES = 180;
+
     public function __construct(
         private readonly string $baggageHeader,
     ) {
@@ -24,7 +29,15 @@ final class Parser
 
     public function parseInto(BaggageBuilderInterface $baggageBuilder): void
     {
+        if (strlen($this->baggageHeader) > self::MAX_BAGGAGE_BYTES) {
+            return;
+        }
+
+        $entries = 0;
         foreach (explode(',', $this->baggageHeader) as $baggageString) {
+            if ($entries >= self::MAX_BAGGAGE_ENTRIES) {
+                break;
+            }
             if (empty(trim($baggageString))) {
                 continue;
             }
@@ -62,6 +75,7 @@ final class Parser
             }
 
             $baggageBuilder->set($key, $value, $metadata);
+            $entries++;
         }
     }
 }
--- a/tests/Unit/API/Baggage/Propagation/ParserTest.php
+++ b/tests/Unit/API/Baggage/Propagation/ParserTest.php
@@ -100,4 +100,42 @@ class ParserTest extends TestCase
             'missing value' => ['key1='],
         ];
     }
+
+    public function test_parse_into_rejects_header_exceeding_w3c_byte_limit(): void
+    {
+        // https://www.w3.org/TR/baggage/#limits, max total bytes 8192
+        $pairs = [];
+        for ($i = 0; $i < 1024; $i++) {
+            $pairs[] = "k{$i}=v{$i}";
+        }
+        $header = implode(',', $pairs);
+        $this->assertGreaterThan(8192, strlen($header));
+
+        $parser = new Parser($header);
+
+        $this->builder
+            ->expects($this->never())
+            ->method('set');
+
+        $parser->parseInto($this->builder);
+    }
+
+    public function test_parse_into_caps_entries_at_w3c_list_member_limit(): void
+    {
+        // https://www.w3.org/TR/baggage/#limits, max list-members 180
+        $pairs = [];
+        for ($i = 0; $i < 500; $i++) {
+            $pairs[] = "k{$i}=v{$i}";
+        }
+        $header = implode(',', $pairs);
+        $this->assertLessThanOrEqual(8192, strlen($header));
+
+        $parser = new Parser($header);
+
+        $this->builder
+            ->expects($this->exactly(180))
+            ->method('set');
+
+        $parser->parseInto($this->builder);
+    }
 }


bobstrecansky (approved):

bobstrecansky:
I had a couple of non-blockers, adding here in case you want to fast-follow @tonghuaroot:


1. Boundary tests — `tests/Unit/API/Baggage/Propagation/ParserTest.php` tests fire well past the limits (1024 pairs, 500 pairs) but don't pin > vs >=. Add cases for exactly 8192 bytes (accept), 8193 bytes (reject), and 180 vs 181 entries to lock the off-by-one.

2. Entry-counter intent comment — `Parser.php:36` — `$entries` increments only on accepted pairs, not raw list-members. A future maintainer could move the  increment and silently change semantics. One-line comment recommended.

3. Silent drop observability — when the byte cap trips the entire header is discarded with no signal. Matches W3C "MAY truncate or discard" and sibling SDKs, but optional self-diagnostics emission would help debugging.

tonghuaroot:
Thanks for the thoughtful review @bobstrecansky! I've opened a fast-follow to pick up your suggestions: #1976.

- **Boundary tests** — added four cases that pin the exact edge the existing tests skipped over: 8192 bytes accepted / 8193 rejected (the byte cap is inclusive, `strlen() > MAX_BAGGAGE_BYTES`), and 180 entries accepted / 181st dropped (`$entries >= MAX_BAGGAGE_ENTRIES`). They lock the off-by-one against the real implemented semantics.
- **Intent comment** — added a one-liner noting that `$entries` counts only *accepted* pairs (incremented at the bottom of the loop), so a future change won't silently relocate the counter and alter the cap semantics.
- **Silent-drop observability** — deferred by design. Discarding an over-cap header matches the W3C spec wording ("MAY truncate or discard") and the behaviour of the sibling SDKs, so I left self-diagnostics out for now; happy to revisit if you'd like it tracked separately.

`phpunit --filter ParserTest` is green (17 tests, 43 assertions).

PR #1976, opened 2026-05-28
Add W3C Baggage boundary tests and entry-counter comment (follow-up to #1975)

tonghuaroot:
Fast-follow to #1975, addressing @bobstrecansky's non-blocking review suggestions.

## What's changed

**1. Boundary tests (`tests/Unit/API/Baggage/Propagation/ParserTest.php`)**

The existing caps tests fire well past the limits (1024 pairs / 500 pairs), which proves the caps engage but does not pin the `>` vs `>=` edge. These four cases lock the off-by-one against the *actual implemented* semantics:

- `test_parse_into_accepts_header_at_exactly_w3c_byte_limit` — exactly 8192 bytes is **accepted** (the cap is inclusive: `strlen() > MAX_BAGGAGE_BYTES`).
- `test_parse_into_rejects_header_one_byte_over_w3c_byte_limit` — 8193 bytes discards the whole header.
- `test_parse_into_accepts_exactly_w3c_list_member_limit` — 180 valid list-members are all accepted.
- `test_parse_into_caps_at_w3c_list_member_limit_plus_one` — the 181st valid member is dropped once `$entries` reaches 180 (`$entries >= MAX_BAGGAGE_ENTRIES`).

**2. Intent comment (`src/API/Baggage/Propagation/Parser.php`)**

One line documenting that `$entries` counts only *accepted* pairs (incremented at the bottom of the loop), not raw comma-separated list-members, so the counter isn't silently relocated in a future refactor.

The silent-drop observability idea is intentionally deferred — discarding an over-cap header matches the W3C spec ("MAY truncate or discard") and the behaviour of the sibling SDKs, so it's not wired in here.

## Testing

`vendor/bin/phpunit --filter ParserTest tests/Unit/API/Baggage/Propagation/ParserTest.php`

```
OK (17 tests, 43 assertions)
```

### DIFF
--- a/src/API/Baggage/Propagation/Parser.php
+++ b/src/API/Baggage/Propagation/Parser.php
@@ -33,6 +33,8 @@ final class Parser
             return;
         }
 
+        // Counts only ACCEPTED entries (incremented at the bottom of the loop), not raw
+        // comma-separated list-members, so empty/invalid pairs do not consume the W3C cap.
         $entries = 0;
         foreach (explode(',', $this->baggageHeader) as $baggageString) {
             if ($entries >= self::MAX_BAGGAGE_ENTRIES) {
--- a/tests/Unit/API/Baggage/Propagation/ParserTest.php
+++ b/tests/Unit/API/Baggage/Propagation/ParserTest.php
@@ -138,4 +138,75 @@ class ParserTest extends TestCase
 
         $parser->parseInto($this->builder);
     }
+
+    public function test_parse_into_accepts_header_at_exactly_w3c_byte_limit(): void
+    {
+        // The byte cap is inclusive: strlen() > 8192 is rejected, so exactly 8192 is accepted.
+        // Single valid pair padded to exactly 8192 bytes with a non-excluded value char.
+        $prefix = 'key=';
+        $header = $prefix . str_repeat('a', 8192 - strlen($prefix));
+        $this->assertSame(8192, strlen($header));
+
+        $parser = new Parser($header);
+
+        $this->builder
+            ->expects($this->once())
+            ->method('set');
+
+        $parser->parseInto($this->builder);
+    }
+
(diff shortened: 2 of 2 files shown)


bobstrecansky (approved):

bobstrecansky:
Fascinating that this failed just PHP 8.1.   Did a quick copilot review:

```
The failing test test_storage_keeps_constant_memory_on_one_active_reader is asserting that memory usage should increase by less than 2000 bytes during 10,000 iterations, but it's actually increasing by 65,536 bytes.

Root Cause: The DeltaStorage class is not properly cleaning up or releasing memory for inactive readers. When only reader 0 is active (calling collect(0)), the storage should not accumulate data for reader 1 (which is never collected).

The Issue at Line 127:

PHP
$this->assertLessThan(2000, memory_get_usage() - $memory);
The test expects constant memory with only one active reader, but memory is growing by 65,536 bytes (approximately 64KB), suggesting that metrics data is being retained for the inactive reader instead of being released after collection.

Fix for DeltaStorage:

The DeltaStorage class needs to implement logic that removes metrics from readers that have collected them. When a reader collects a delta, if that reader is the only one who hasn't collected yet, the metric should be purged from storage.

Key changes needed in the DeltaStorage implementation:

Track which readers have collected each metric version
Remove metrics once all active readers have collected them
Ensure proper cleanup to prevent memory accumulation
The test validates that with only one active reader consistently collecting, the memory footprint remains stable. The current implementation is likely keeping historical metric data that should be discarded once a reader has accessed it.

Verification: After fixing the storage cleanup logic, the memory growth during 10,000 iterations with only one active reader should remain under 2000 bytes, confirming that old data is properly released.
```

Rerunning to test.

bobstrecansky:
Looks like it might have just been a GHA flake, merging.

PR #1904, opened 2026-03-16
Support Protobuf 5.0

knobik:
Added support for google protobuf v5.0, fixes issues with php 8.5

PR #1983, opened 2026-06-11
Allow google/protobuf 5

derrabus:
#1904 enabled the usage of google/protobuf 5, but apparently the composer.json file of the subtree split was forgotten. Because of that, downstream projects can't use `open-telemetry/gen-otlp-protobuf` together with google/protobuf 5.


ChrisLightfootWild (approved):
Thanks @derrabus

PR #1985, opened 2026-06-11
Fix(SDK): Auto root span start timestamps precision

zigzagdev:
### Content
Fixed a precision loss bug in `AutoRootSpan::create()` where the start timestamp was calculated incorrectly when `REQUEST_TIME_FLOAT` is not available.

### Root Cause
The original code cast `microtime(true)` to `int` before multiplying via `NANOS_PER_SECOND`:

```php
$startTime = array_key_exists('REQUEST_TIME_FLOAT', $request->getServerParams())
    ? $request->getServerParams()['REQUEST_TIME_FLOAT'] : (int) microtime(true);          // ← precision loss here
->setStartTimestamp((int) ($startTime * ClockInterface::NANOS_PER_SECOND))
```

This caused up to 999,999,999ns (~1 second) of error because the sub-second precision was discarded before nanosecond conversioning.

- `GET /?foo=bar` with `microtime(true)` = `1718123456.789` → cast to `1718123456` → multiplied by `1e9` → last 9 digits always `000000000`

### What I have done
Multiply first, then cast to `int` to preserve sub-second precision like:

```php
$startTimeInNanos = array_key_exists('REQUEST_TIME_FLOAT', $request->getServerParams())
? (int) ($request->getServerParams()['REQUEST_TIME_FLOAT'] * ClockInterface::NANOS_PER_SECOND)
    : (int) (microtime(true) * ClockInterface::NANOS_PER_SECOND)
->setStartTimestamp($startTimeInNanos)
```

### Test Results
screenshot 2026-06-11 23 33 52

### DIFF
--- a/src/SDK/Trace/AutoRootSpan.php
+++ b/src/SDK/Trace/AutoRootSpan.php
@@ -43,12 +43,12 @@ class AutoRootSpan
             Version::VERSION_1_25_0->url(),
         );
         $parent = Globals::propagator()->extract($request->getHeaders());
-        $startTime = array_key_exists('REQUEST_TIME_FLOAT', $request->getServerParams())
-            ? $request->getServerParams()['REQUEST_TIME_FLOAT']
-            : (int) microtime(true);
+        $startTimeInNanos = array_key_exists('REQUEST_TIME_FLOAT', $request->getServerParams())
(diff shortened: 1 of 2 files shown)


ChrisLightfootWild (approved):
Thanks for the PR @zigzagdev 💪

PR #1990, opened 2026-06-25
fix(sdk): wire meterProvider for span self-observability metrics and add otel.sdk.span.ended

intuibase:
SdkAutoloader was not passing the meterProvider to TracerProviderBuilder, so TracerSharedState always received null and never created the otel.sdk.span.started / otel.sdk.span.live counters even when OTEL_PHP_INTERNAL_METRICS_ENABLED=true.

TracerProviderFactory (used directly in examples) had the same gap — it was not forwarding a meterProvider to either SpanProcessorFactory or TracerProvider.

Also implements otel.sdk.span.ended (defined in SemConv but previously unrecorded): counter is created in TracerSharedState alongside spanStartedCounter and incremented in Span::end() with the otel.span.sampling_result attribute.

### DIFF
--- a/src/SDK/Logs/Processor/BatchLogRecordProcessor.php
+++ b/src/SDK/Logs/Processor/BatchLogRecordProcessor.php
@@ -19,6 +19,7 @@ use OpenTelemetry\SDK\Logs\LogRecordProcessorInterface;
 use OpenTelemetry\SDK\Logs\ReadWriteLogRecord;
 use OpenTelemetry\SemConv\Incubating\Attributes\OtelIncubatingAttributes;
 use OpenTelemetry\SemConv\Incubating\Metrics\OtelIncubatingMetrics;
+use OpenTelemetry\SemConv\Version;
 use SplQueue;
 use Throwable;
 
@@ -106,7 +107,7 @@ class BatchLogRecordProcessor implements LogRecordProcessorInterface
                 OtelIncubatingAttributes::OTEL_COMPONENT_NAME => (new \ReflectionClass($this->exporter))->getShortName(),
             ];
 
-            $meter = $meterProvider->getMeter('io.opentelemetry.sdk', schemaUrl: 'https://opentelemetry.io/schemas/1.32.0');
+            $meter = $meterProvider->getMeter('io.opentelemetry.sdk', schemaUrl: Version::VERSION_1_32_0->url());
             $meter
                 ->createObservableUpDownCounter(
                     OtelIncubatingMetrics::OTEL_SDK_PROCESSOR_LOG_QUEUE_CAPACITY,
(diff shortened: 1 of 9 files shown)


ChrisLightfootWild (approved):

src/SDK/Trace/TracerSharedState.php
Nevay: Should set schema url.

src/SDK/Trace/TracerSharedState.php
```diff
@@ -46,13 +47,19 @@ public function __construct(
                 '{span}',
                 'The number of created spans',
             );
+            $this->spanEndedCounter = $meter->createCounter(
+                OtelIncubatingMetrics::OTEL_SDK_SPAN_ENDED,
+                '{span}',
+                'The number of spans which have been finished by a Span.end() call',
+            );
```
Nevay: Should not record `otel.sdk.span.ended`. It was replaced by `otel.sdk.span.started` in `1.36.0`.

> Use `otel.sdk.span.started` minus `otel.sdk.span.live` to derive this value.

Nevay (changes requested):

src/SDK/Logs/Processor/BatchLogRecordProcessor.php
```diff
@@ -106,7 +107,7 @@ public function __construct(
                 OtelIncubatingAttributes::OTEL_COMPONENT_NAME => (new \ReflectionClass($this->exporter))->getShortName(),
             ];
 
-            $meter = $meterProvider->getMeter('io.opentelemetry.sdk');
+            $meter = $meterProvider->getMeter('io.opentelemetry.sdk', schemaUrl: Version::VERSION_1_32_0->url());
```
Nevay: Must be >= `1.36.0`

Nevay (approved):

src/SDK/Logs/Processor/BatchLogRecordProcessor.php
```diff
@@ -106,7 +107,7 @@ public function __construct(
                 OtelIncubatingAttributes::OTEL_COMPONENT_NAME => (new \ReflectionClass($this->exporter))->getShortName(),
             ];
 
-            $meter = $meterProvider->getMeter('io.opentelemetry.sdk');
+            $meter = $meterProvider->getMeter('io.opentelemetry.sdk', schemaUrl: Version::VERSION_1_36_0->url());
```
ChrisLightfootWild, line 110: Need to bump this to `open-telemetry/sem-conv: ^1.36`, https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/composer.json#L25C9-L25C43

intuibase: good point, created PR 
https://github.com/open-telemetry/opentelemetry-php/pull/1994

PR #1991, opened 2026-06-26
fix: preserve debug log level priority in Logging::level()

jorgsowa:
array_search returns 0 for debug (first in the LEVELS array) and false when not found; the old ?: 1 collapsed both to 1 (info), silently ignoring debug-level messages. Use !== false to distinguish the two.


ChrisLightfootWild:
@jorgsowa thanks for this. Would you be able to take a look at the failing unit test please?

jorgsowa:
@ChrisLightfootWild test adjusted.

ChrisLightfootWild (approved):

PR #1997, opened 2026-07-01
ci: promote PHP 8.5 to the stable test matrix, add PHP 8.6 as experimental

jorgsowa:
PHP 8.5 is a stable release for almost a year, so add it to the php-version matrix alongside 8.4 and drop the experimental/continue-on-error machinery that previously let its failures pass silently. Also remove the --ignore-platform-reqs composer args, no longer needed.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -49,7 +49,7 @@ jobs:
           echo "spi=${{steps.spi.outcome}}"
           echo "::endgroup::"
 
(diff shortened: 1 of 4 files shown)


ChrisLightfootWild:
@jorgsowa perhaps 8.6 can take its place in the matrix instead, so we can run the next-gen in the pipeline?

jorgsowa:
@ChrisLightfootWild I'm hopeless towards Phan. It's now running the newest version but still lot of false positives. Should I ignore it?

jorgsowa:
Psalm and Phan fail on PHP 8.5, but because of upstream bugs, not codebase errors, I ignore their results.

jorgsowa:
Fix for PHPStan is extracted to separate PR https://github.com/open-telemetry/opentelemetry-php/pull/2007

jorgsowa:
Finally, all checks passed

ChrisLightfootWild (approved):

intuibase (approved):
LGTM

PR #2004, opened 2026-07-08
fix(sdk): don't emit placeholder service.version in Composer detector

### LINKED ISSUE #1320: service.version being set by Composer
@julianocosta89 opened a slack thread to discuss seeing the `service.version` resource attribute set, even though it was not being set by default.

After getting more clarity in the #otel-community-demo channel, @puckpuck mentioned the following:

```
I did some sleuthing. This is set in the Composer detector here: https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Resource/Detectors/Composer.php#L23-L26
Within Composer if you don't have a version set you will get the Default version which matches what we see in the otel-demo https://github.com/composer/composer/blob/main/src/Composer/Package/RootPackage.php#L22
I wonder if we should be using the version from Composer if it is set to the Default (unset)?
```

\
We should probably determine how we want to handle this `service.version` in the future.  Other SIGs do not set this by default, so perhaps maybe we shouldn't have composer set this value if possible.

PuvaanRaaj:
## Problem

Closes #1320.

When a Composer root package has no explicit `version`, Composer substitutes the placeholder pretty version `1.0.0+no-version-set` (`Composer\Package\RootPackage::DEFAULT_PRETTY_VERSION`). The `Composer` resource detector passed `pretty_version` straight through, so any application that never declares a version — e.g. the [OpenTelemetry demo](https://github.com/open-telemetry/opentelemetry-demo) — ended up with a misleading `service.version = 1.0.0+no-version-set`.

Likewise, when no package `name` is set Composer substitutes the placeholder `__root__`, which should not surface as `service.name`.

Other language SDKs do not set `service.version` by default, as discussed in the issue.

## Fix

- Skip the `service.version` attribute when Composer reports the `1.0.0+no-version-set` placeholder.
- Skip the `service.name` attribute when Composer reports the `__root__` placeholder.
- A real, explicitly-set name/version is still reported as before.

The detector reads `InstalledVersions::getRootPackage()` directly (no constructor injection). Tests drive the placeholder paths by swapping the installed-versions dataset via `InstalledVersions::reload()`, restoring the real data in `tearDown()`.

## Tests

- `test_reports_service_version_when_set` — explicit version is emitted.
- `test_omits_service_version_when_composer_reports_placeholder` — `1.0.0+no-version-set` is dropped.
- `test_omits_service_name_when_composer_reports_placeholder` — `__root__` is dropped.
- Existing detector tests still pass; no state leaks across the suite.

### DIFF
--- a/src/SDK/Resource/Detectors/Composer.php
+++ b/src/SDK/Resource/Detectors/Composer.php
@@ -18,17 +18,42 @@ use OpenTelemetry\SemConv\Version;
  */
 final class Composer implements ResourceDetectorInterface
 {
+    /**
+     * Placeholder version that Composer reports for a root package with no explicit
+     * version (Composer\Package\RootPackage::DEFAULT_PRETTY_VERSION).
+     */
+    private const UNSET_VERSION = '1.0.0+no-version-set';
+
+    /**
+     * @param array{name?: string, pretty_version?: ?string}|null $rootPackage overrides
+     *        the detected Composer root package; used for testing
+     */
+    public function __construct(private readonly ?array $rootPackage = null)
+    {
+    }
+
     #[\Override]
     public function getResource(): ResourceInfo
     {
-        if (!class_exists(InstalledVersions::class)) {
-            return ResourceInfoFactory::emptyResource();
+        $rootPackage = $this->rootPackage;
+        if ($rootPackage === null) {
+            if (!class_exists(InstalledVersions::class)) {
+                return ResourceInfoFactory::emptyResource();
+            }
+            $rootPackage = InstalledVersions::getRootPackage();
         }
 
         $attributes = [
-            ServiceAttributes::SERVICE_NAME => InstalledVersions::getRootPackage()['name'],
-            ServiceAttributes::SERVICE_VERSION => InstalledVersions::getRootPackage()['pretty_version'],
+            ServiceAttributes::SERVICE_NAME => $rootPackage['name'] ?? null,
         ];
+
+        // Only report service.version when the root package has an explicit version.
+        // Composer substitutes a placeholder ("1.0.0+no-version-set") when none is set,
+        // and other OpenTelemetry SDKs do not set service.version by default. See #1320.
+        $version = $rootPackage['pretty_version'] ?? null;
+        if ($version !== null && $version !== self::UNSET_VERSION) {
+            $attributes[ServiceAttributes::SERVICE_VERSION] = $version;
+        }
 
         return ResourceInfo::create(Attributes::create($attributes), Version::VERSION_1_38_0->url());
     }
--- a/src/SDK/Resource/Detectors/Composer.php
+++ b/src/SDK/Resource/Detectors/Composer.php
@@ -19,40 +19,38 @@ use OpenTelemetry\SemConv\Version;
 final class Composer implements ResourceDetectorInterface
 {
     /**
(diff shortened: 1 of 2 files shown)


PuvaanRaaj:
@open-telemetry/php-approvers small SDK fix for #1320 (Composer resource detector) with unit tests — review welcome when convenient. 🙏

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -18,18 +18,43 @@
...
+            if (!class_exists(InstalledVersions::class)) {
+                return ResourceInfoFactory::emptyResource();
+            }
+            $rootPackage = InstalledVersions::getRootPackage();
         }
 
         $attributes = [
-            ServiceAttributes::SERVICE_NAME => InstalledVersions::getRootPackage()['name'],
-            ServiceAttributes::SERVICE_VERSION => InstalledVersions::getRootPackage()['pretty_version'],
+            ServiceAttributes::SERVICE_NAME => $rootPackage['name'] ?? null,
```
Nevay: Should skip setting `service.name` if `__root__`.

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -18,18 +18,43 @@
...
 final class Composer implements ResourceDetectorInterface
 {
+    /**
+     * Placeholder version that Composer reports for a root package with no explicit
+     * version (Composer\Package\RootPackage::DEFAULT_PRETTY_VERSION).
+     */
+    private const UNSET_VERSION = '1.0.0+no-version-set';
+
+    /**
+     * @param array{name?: string, pretty_version?: ?string}|null $rootPackage overrides
```
Nevay:
```suggestion
     * @param array{name: string, pretty_version: string, ...}|null $rootPackage overrides
```

If we keep the parameter; to match `InstalledVersions::getRootPackage()`.

IMO we should instead `InstalledVersions::reload()` the test data to keep the implementation clean.

src/SDK/Resource/Detectors/Composer.php
PuvaanRaaj: Done — now also skips service.name when the root package name is '__root__'.

src/SDK/Resource/Detectors/Composer.php
PuvaanRaaj: Reworked as suggested: dropped the $rootPackage constructor param and read InstalledVersions::getRootPackage() directly; tests now inject via InstalledVersions::reload(). Cleaner. Thanks!

Nevay (approved):

src/SDK/Resource/Detectors/Composer.php
```diff
@@ -18,17 +18,40 @@
...
+
     #[\Override]
     public function getResource(): ResourceInfo
     {
         if (!class_exists(InstalledVersions::class)) {
             return ResourceInfoFactory::emptyResource();
         }
+        $rootPackage = InstalledVersions::getRootPackage();
+
+        $attributes = [];
```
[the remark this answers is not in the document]
PuvaanRaaj: Updated the PR description to match the implementation — the constructor-injection mention is gone; it now describes reading InstalledVersions::getRootPackage() directly with tests driven via InstalledVersions::reload(). Thanks.

bobstrecansky (approved):

PR #2007, opened 2026-07-08
build: upgrade phpstan to 2.x to resolve PHP 8.5 attribute false positives

jorgsowa:
PHPStan 1.x cannot correctly resolve Attribute::IS_REPEATABLE under PHP 8.5 (support only landed in 2.1.x), causing false "not repeatable" errors for CoversClass, UsesClass, Depends, and PackageDependency.

Bumps phpstan/phpstan to ^2.1 and its phpunit/mockery extensions to ^2.0, and baselines the pre-existing findings newly surfaced by PHPStan 2.x's stricter analysis for later cleanup.


jorgsowa:
> CI runs PHPStan across PHP 8.1–8.5 plus lowest-deps builds, and the reported errors differ between those jobs. With unmatched-error reporting turned on, any baseline or ignore entry that doesn't match in a particular matrix job would fail that job.

I don't think Copilot comments are relevant.

bobstrecansky (approved):

PR #2011, opened 2026-07-17
chore: docker image housekeeping

thompson-tomo:
Explicitly pin docker version to improve info/context available to renovate.

The jaegar tracing image has been switched as existing image is eol https://github.com/jaegertracing/jaeger/issues/6321

### DIFF
--- a/docker-compose.collector.yaml
+++ b/docker-compose.collector.yaml
@@ -12,8 +12,6 @@ services:
     - 9411:9411
   jaeger:
     image: jaegertracing/jaeger:2.19.0@sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b
-    environment:
-      COLLECTOR_ZIPKIN_HOST_PORT: 9412
     ports:
     - 9412:9412
     - 16686:16686
--- a/docker-compose.collector.yaml
+++ b/docker-compose.collector.yaml
@@ -7,7 +7,7 @@ services:
     depends_on:
       - collector
   zipkin:
-    image: openzipkin/zipkin-slim3:6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
+    image: openzipkin/zipkin-slim:3:6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
     ports:
     - 9411:9411
   jaeger:
--- a/docker-compose.collector.yaml
+++ b/docker-compose.collector.yaml
@@ -7,18 +7,18 @@ services:
     depends_on:
       - collector
   zipkin:
-    image: openzipkin/zipkin-slim@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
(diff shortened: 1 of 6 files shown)


PuvaanRaaj (commented):
Reviewed the Docker image housekeeping changes and validated the changed image references with docker manifest inspect. Two references in the current head do not resolve and would break the affected Compose workflows.

docker-compose.fiber-ffi.yaml
```diff
@@ -1,7 +1,7 @@
 version: '3.8'
 services:
   web:
-    image: nginx:alpine@sha256:54f2a904c251d5a34adf545a72d32515a15e08418dae0266e23be2e18c66fefa
+    image: nginx:1.30.4-alpine@sha256:sha256:59d10bca5c674965ef4ff884715000dd60ef5567c36663523f108eec8e4105d4
```
PuvaanRaaj: This reference contains a duplicated sha256 prefix. I verified that docker manifest inspect fails for the current value. It should use a single digest prefix: nginx:1.30.4-alpine@sha256:59d10bca5c674965ef4ff884715000dd60ef5567c36663523f108eec8e4105d4.

docker-compose.collector.yaml
```diff
@@ -7,18 +7,16 @@ services:
     depends_on:
       - collector
   zipkin:
-    image: openzipkin/zipkin-slim@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
+    image: openzipkin/zipkin-slim3:6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
```
PuvaanRaaj: This changes the repository name from openzipkin/zipkin-slim to openzipkin/zipkin-slim3 and the resulting reference does not resolve. I verified that docker manifest inspect fails here, while openzipkin/zipkin-slim:3.6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c resolves successfully.

thompson-tomo:
@PuvaanRaaj Issues have been fixed

PuvaanRaaj (commented):
Follow-up review on the updated head: the nginx digest reference is corrected, but the revised Zipkin tag still has an invalid separator.

docker-compose.collector.yaml
```diff
@@ -7,18 +7,16 @@ services:
     depends_on:
       - collector
   zipkin:
-    image: openzipkin/zipkin-slim@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
+    image: openzipkin/zipkin-slim:3:6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c
```
PuvaanRaaj: Thanks for the quick update. This value is still not a valid Docker reference because it contains two tag separators (:3:6.1). The resolving form is openzipkin/zipkin-slim:3.6.1@sha256:fc29b9862c14f4fc241cf0259c585e1cfd4e9b18f01ad36d86adaabb7d01815c.

thompson-tomo:
Apologies didn't notice that the : had gone after the 3. Has been fixed.

bobstrecansky (approved):

PR #2041, opened 2026-08-24
Remove static state in SystemClock

SamMousa:
In the library code base `SystemClock` is used as a singleton (via `Clock::getDefault()`). So this change will not actually impact much.

Practically it simplifies `SystemClock` by removing its internal static state, which I consider to be a code smell.

- We maintain internal monotonicity by computing the reference time at construction
- The library maintains monotonicity by only using the singleton instance

We lose monotonicity between different instances, I do not think this is a bad thing, it could actually be a good thing.
For long running PHP processes (for example ReactPHP or other scenarios where a single process handles many requests) we could choose to instantiate a system clock for each request. Having a single reference time at process start means we can drift away from real time because we're not getting resynced with NTP.

FYI: While the code was edited by opencode, this description is handwritten.

### DIFF
--- a/src/API/Common/Time/SystemClock.php
+++ b/src/API/Common/Time/SystemClock.php
@@ -12,11 +12,14 @@ use function microtime;
  */
 final class SystemClock implements ClockInterface
 {
-    private static int $referenceTime = 0;
+    private readonly int $referenceTime;
 
     public function __construct()
     {
-        self::init();
+        $this->referenceTime = self::calculateReferenceTime(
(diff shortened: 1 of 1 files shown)


SamMousa:
I'll get the CLA thing fixed, already contacted them to enroll our organisation etc.

bobstrecansky:
Once your CLA is signed we can review!  Thank you @SamMousa !

ChrisLightfootWild (approved):

PR #2049, opened 2026-09-03
Update semconv for otel schema v1.39, 1.40, 1.41,1.42,1.43,1.44 based on schem and migrate incu…

DaanBiesterbos:
## Semantic Conventions: update to v1.44.0

Updates the OpenTelemetry Semantic Conventions from **v1.38.0** to **v1.44.0** and upgrades the Weaver code generator from `v0.20.0` to `v0.25.1`.

---

### Generator changes

- Bumped `SEMCONV_VERSION` from `1.38.0` to `1.44.0` in `semconv.sh`
- Bumped `OTEL_WEAVER_IMG_VERSION` from `v0.20.0` to `v0.25.1`
- Removed `feature_flag`, `jsonrpc`, and `rpc` from the `excluded_namespaces` list in `weaver.yaml`, enabling generation of their respective attribute files

---

### New files

| File | Reason |
|------|--------|
| `Attributes/ContainerAttributes.php` | Container attributes graduated to stable |
| `Attributes/DeploymentAttributes.php` | Deployment attributes graduated to stable |
| `Incubating/Attributes/FeatureFlagIncubatingAttributes.php` | `feature_flag` namespace unexcluded |
| `Incubating/Attributes/JsonrpcIncubatingAttributes.php` | `jsonrpc` namespace unexcluded (renamed from `rpc.jsonrpc.*`) |
| `Incubating/Attributes/RpcIncubatingAttributes.php` | `rpc` namespace unexcluded |
| `Incubating/Attributes/OracleIncubatingAttributes.php` | New namespace in spec |
| `Incubating/Attributes/OracleCloudIncubatingAttributes.php` | New namespace in spec |
| `templates/.../ErrorIncubatingAttributes_deprecations.php.partial` | Backward-compat stub for deprecated `error.message` |

---

### Attribute graduations (experimental to stable)

The following attributes were promoted to stable in the spec. Stable constants added to the `Attributes\*` namespace; incubating constants updated to `@stable`:

- `otel.event.name`
- `service.instance.id`
- `service.namespace`
- `telemetry.distro.name`
- `telemetry.distro.version`
- `telemetry.sdk.language` enum values (incl. new `kotlin`)
- `container.id`, `container.image.name`, `container.image.repo_digests`, `container.image.tags`
- `deployment.environment.name` (incl. new enum values: `production`, `staging`, `test`, `development`)

---

### New attributes (spec v1.39-v1.44)

- `service.criticality` + enum values (`critical`, `high`, `medium`, `low`)
[...]

PR #2050, opened 2026-09-05
[DRAFT] Update opentelemetry configuration (v1.1)

DaanBiesterbos:
Add support for current [v1.1 configuration](https://github.com/open-telemetry/opentelemetry-configuration) format.

Why?
I am responsible for the implementation of opentelemetry on about 40 applications for one of my clients. Due to mixed infrastructures (k8s, bare metal, vm's) and existing pipelines it will be troublesome to use environment variables. The configuration format is a better fit. The library currently supports _only_ the v1.0-rc2 format. I can't ship software build on RC flagged formats that runs on all applications.

This MR provides backward compatible support for v1.1, using the schema version in the file for validation and normalization. The current semconv defintions are outdated (missing everything above 1.38 up until v1.44).  

Added some minor changes in the SDK to implement the added configurations.
The changes to the SDK require a bit more testing. I don't mind to dive into it. But I first need to see that someone is actively cooperating with me. I'd like to get in touch so my efforts are not in vane.

This MR is rather big because it depends on  [#2049](https://github.com/open-telemetry/opentelemetry-php/pull/2049)
We'll need to merge this before we merge this one.


src/SDK/Resource/Detectors/Container.php
```diff
@@ -0,0 +1,66 @@
...
+use OpenTelemetry\SemConv\Attributes\ContainerAttributes;
+use OpenTelemetry\SemConv\Version;
+
+/**
+ * Detects container resource attributes by reading the container ID from the
+ * cgroup filesystem available on Linux.
+ *
+ * @see https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/container.md
+ */
+final class Container implements ResourceDetectorInterface
```
Nevay: Already implemented in https://github.com/opentelemetry-php/contrib-detector-container.

Must not be part of the SDK https://opentelemetry.io/docs/specs/otel/resource/sdk/#detecting-resource-information-from-the-environment:
> Custom resource detectors related to generic platforms (e.g. Docker, Kubernetes) or vendor specific environments (e.g. EKS, AKS, GKE) MUST be implemented as packages separate from the SDK.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -393,13 +409,21 @@ public function createPlugin(array $properties, Context $context): SdkBuilder
...
         }
 
-        // TODO Allow injecting log record attributes factory
+        $logRecordLimitsBuilder = new LogRecordLimitsBuilder();
+        if ($properties['logger_provider']['limits']['attribute_count_limit'] !== null) {
+            $logRecordLimitsBuilder->setAttributeCountLimit($properties['logger_provider']['limits']['attribute_count_limit']);
+        }
+        if ($properties['logger_provider']['limits']['attribute_value_length_limit'] !== null) {
+            $logRecordLimitsBuilder->setAttributeValueLengthLimit($properties['logger_provider']['limits']['attribute_value_length_limit']);
+        }
```
Nevay: Must use `$properties['attribute_limits']` if the respective limit is not set in `$properties['logger_provider']['limits']`.

Should not use the fallback to env variable configuration in `LogRecordLimitsBuilder::build()`: either add a parameter to `LogRecordLimitsBuilder`s constructor to disable the env fallback or set the limits unconditionally.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -420,14 +444,38 @@ public function getConfig(ComponentProviderRegistry $registry, NodeBuilder $buil
...
                 ->booleanNode('disabled')->defaultFalse()->end()
+                ->enumNode('log_level')
+                    ->values([
+                        'trace', 'trace2', 'trace3', 'trace4',
+                        'debug', 'debug2', 'debug3', 'debug4',
+                        'info', 'info2', 'info3', 'info4',
+                        'warn', 'warn2', 'warn3', 'warn4',
+                        'error', 'error2', 'error3', 'error4',
+                        'fatal', 'fatal2', 'fatal3', 'fatal4',
+                    ])
```
Nevay: Could use `array_map(strtolower(...), array_column(Severity::cases(), 'name'))` instead.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -680,21 +732,178 @@ private function getLoggerProviderConfig(ComponentProviderRegistry $registry, No
...
+
     private function getPropagatorConfig(ComponentProviderRegistry $registry, NodeBuilder $builder): ArrayNodeDefinition
     {
         $node = $builder->arrayNode('propagator');
         $node
             ->beforeNormalization()
                 ->ifArray()
                 ->then(static function (array $value): array {
-                    $existing = [];
+                    // Normalize composite entries: bare string "name" → ["name" => null]
```
Nevay: This normalization should not be needed; `composite` cannot contain `string` values.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -65,8 +71,9 @@ final class OpenTelemetrySdk implements ComponentProvider
 {
     /**
      * @param array{
-     *     file_format: '1.0-rc.2',
+     *     file_format: '1.0-rc.2'|'1.1',
```
Nevay:
```suggestion
     *     file_format: '1.0'|'1.1',
```

IMO we should drop the compatibility layer for `1.0-rc.2` and support only `1.0` and `1.1`.

This PR will be released as either `0.10.0` or `1.0.0`; users using `1.0-rc.2` can continue using `0.9.0`.

DaanBiesterbos: Fair enough. How about using semver checking? We can parse the version. And we can trigger a deprecation notice for RC versions? That way it won't break existing deployments and we can still support 1.0 and 1.1?

DaanBiesterbos: https://github.com/open-telemetry/opentelemetry-configuration/releases

I don't see 0.9, and it seems we never supported it. I was considering to add support for 0.4 to support the 0.* range but since there are many changes and support was never there we can just ignore it?

src/SDK/Resource/Detectors/Container.php
DaanBiesterbos: I will remove it. One question though. 

If I am correct, we would need to add the provider to  https://github.com/opentelemetry-php/contrib-detector-container. 

And we need this in composer.json:

```json
{
       "config": {                                                                                                                                                                         
             "allow-plugins": {                                                                                                                                                              
                 "php-http/discovery": true,                                                                                                                                                 
                 "tbachert/spi": true                                                                                                                                                        
             }                                                                                                                                                                               
         },                                                                                                                                                                                  
         "extra": {                                                                                                                                                                          
             "spi": {                                                                                                                                                                        
                 "OpenTelemetry\\API\\Configuration\\Config\\ComponentProvider": [                                                                                                           
                     "OpenTelemetry\\Contrib\\Resource\\Detector\\Container\\ContainerDetectorComponentProvider"                                                                             
                 ]                                                                                                                                                                           
             }                                                                                                                                                                               
         }  
}
```

As far as I can see we'd need to add this to any custom detector we want to support in the configuration?
We can't read OTEL_PHP_DETECTORS so I assume we need another issue to integrate the custom detectors.

DaanBiesterbos:
@Nevay  I've pushed the requested changes.  Instead of removing support for the RC version I added a deprecation error. This way I don't introduce breaking changes. Removing 1.0-RC.2 could break existing deployments otherwise.

src/SDK/Resource/Detectors/Container.php
Nevay: We can use a `#[PackageDependency('open-telemetry/detector-container', '^1.0')]` attribute on the [`Container`](https://github.com/open-telemetry/opentelemetry-php/pull/2050/changes/528e4ce0b04d19cd01e3d6d641c16bad6c1f0b08#diff-96e5cf9b687dac05c4580f2376dcb20f11bcb903c7c159b3e823657452c65187) `ComponentProvider`, similar to https://github.com/open-telemetry/opentelemetry-php/blob/main/src/Config/SDK/ComponentProvider/Propagator/TextMapPropagatorB3.php#L19.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
Nevay: > I don't see 0.9

I was referring to the version of the [`open-telemetry/sdk-configuration`](https://packagist.org/packages/open-telemetry/sdk-configuration) package rather than the `file_format` version. I should have been more explicit.

> How about using semver checking? We can parse the version. And we can trigger a deprecation notice for RC versions?

I'm fine with keeping a BC-layer if we release this as `open-telemetry/sdk-configuration` `0.10.0`.
(Though I'm not sure if it's worth the additional complexity. We immediately dropped support for the previous version when upgrading from `file_format` `0.3` to `0.4` and from `0.4` to `1.0-rc.2` / upgrading from `open-telemetry/sdk-configuration` `^0.9` to `0.10.0` still requires users to explicitly bump the dependency.)

I don't think we should maintain support for `file_format` versions `<` stable `1.0` in our `open-telemetry/sdk-configuration` `1.0` release.

DaanBiesterbos: True. But when would we release 1.0.0?  As far as I can see this is this only breaking change. Not doing this would break things overnight otherwise. I am hoping we can ship this as a minor release or else I need to use RC in about ~40 applications only to change them soon after.

Would it work for you if I create another merge request to remove support. Which can be merged into the 1.0.0 release?

DaanBiesterbos:
@Nevay Hi Tobias, can I mark the issues resolved?

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
ChrisLightfootWild: `v0.10.0` with BC, and then tagging a `v1.0` in a follow-up sounds like a plan.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -680,21 +730,173 @@ private function getLoggerProviderConfig(ComponentProviderRegistry $registry, No
...
+        if (array_key_exists('disabled', $value) && !array_key_exists('enabled', $value) && SemVer::gte($version, '1.1')) {
+            trigger_error(sprintf(
+                '"%s.disabled" is deprecated since file_format "1.1"; use "enabled: %s" instead.',
+                $path,
+                $value['disabled'] ? 'false' : 'true',
+            ), \E_USER_DEPRECATED);
+        }
+
+        if (array_key_exists('enabled', $value)) {
+            if (SemVer::lt($version, '1.1')) {
```
Nevay, line 863: `disabled` was changed to `enabled` in `1.0`, not `1.1`.

src/Config/SDK/ComponentProvider/OpenTelemetrySdk.php
```diff
@@ -680,21 +730,173 @@ private function getLoggerProviderConfig(ComponentProviderRegistry $registry, No
...
             ->beforeNormalization()
                 ->ifArray()
                 ->then(static function (array $value): array {
-                    $existing = [];
+                    $normalized = [];
                     foreach ($value['composite'] ?? [] as $item) {
-                        $existing[] = key($item);
+                        $normalized[] = $item;
                     }
+                    $value['composite'] = $normalized;
```
Nevay: This loop can be simplified to `$value['composite'] ??= [];`

src/Config/SDK/Configuration/OtelConfigVersion.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Use {@see self::fromVersion()} to resolve a raw file_format string (which may include
+ * a patch component or a pre-release suffix) to a known version constant.
+ *
+ * @internal
+ */
+enum OtelConfigVersion: string
+{
+    case V1_0 = '1.0';
+    case V1_1 = '1.1';
```
Nevay, line 18:
```suggestion
    case V1_0 = '1.0';
    case V1_1 = '1.1';
    case V1_2 = '1.2';
```

Seems like most of the `file_format` `1.2` changes are already incorporated.
https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CHANGELOG.md#v120---2026-09-11

DaanBiesterbos: Fair point. I am currently busy with my taxes. But when I have time I will add 1.2 and look a bit deeper to what is missing and what should be supported per version.

PR #2059, opened 2026-09-19
Fix OTLP gRPC exporter env-based configuration issues: certificates and signal specific endpoints

Nevay:
Fixes:
- environment exporter factories do not pass certificate environment variables (`OTEL_EXPORTER_OTLP_CERTIFICATE`/`_CLIENT_KEY`/`_CLIENT_CERTIFICATE`) to the transport factory
- environment exporter factories do not append gRPC method to signal-specific gRPC endpoints: appending the gRPC method to the endpoint is a PHP-specific implementation detail of the `TransportFactory` abstraction and does not violate "the URL is used as-is for them, without any modifications"

Additionally adds support for:
- `OTEL_EXPORTER_OTLP_INSECURE` and its signal-specific variants

---

The following issue descriptions and the test suite (['Nevay/otel-sdk-integration-tests](https://github.com/Nevay/otel-sdk-integration-tests)) that identified these issues were generated by Qwen 3.8.

This PR resolves all [15 gRPC failures](https://github.com/Nevay/otel-sdk-integration-tests/blob/main/sdk-support-overview.md#upstream-tracking-official-sdk-failures-by-root-cause).
The OTLP/HTTP-based `TlsTest`s are still failing due to the missing `$cacert`/`$cert`/`$key` support in `PsrTransportFactory`.

---

## Issue 1: OTLP exporter ignores the certificate environment variables (HTTP and gRPC)

**Describe your environment**

- PHP 8.4.25 (Linux x86_64)
- `ext-grpc` 1.84.0 (C-core based), `grpc/grpc` 1.82.0
- `open-telemetry/sdk` 1.15.0 (latest release as of 2026-09-16)
- `open-telemetry/exporter-otlp` 1.4.0 (latest release as of 2026-09-16)
- `open-telemetry/transport-grpc` 1.2.0 (latest release as of 2026-09-16)

**Steps to reproduce**

Run an OTLP/gRPC exporter against a server that presents a certificate not
signed by a public CA (e.g. a self-signed test certificate), configuring the
CA via the standard environment variables:

```sh
OTEL_TRACES_EXPORTER=otlp
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_EXPORTER_OTLP_ENDPOINT=https://127.0.0.1:4317
OTEL_EXPORTER_OTLP_CERTIFICATE=/path/to/ca.pem
```

```php
$tracer = \OpenTelemetry\API\Globals::tracerProvider()->getTracer('test');
$span = $tracer->spanBuilder('probe')->startSpan();
$span->end();
// export happens on shutdown / batch flush
```

Per the OTLP exporter specification, `OTEL_EXPORTER_OTLP_CERTIFICATE` (and the
signal-specific `OTEL_EXPORTER_OTLP_<SIGNAL>_CERTIFICATE`) configures the CA
used to verify the collector's certificate. The export should succeed when the
collector's certificate chains to the configured CA.

The export fails during the TLS handshake, and the configured CA is never
used:

```
OpenTelemetry: [error] Export failure [exception] failed to connect to all
addresses; last error: UNKNOWN: ipv4:127.0.0.1:4317: Ssl handshake failed
(TSI_PROTOCOL_FAILURE): SSL_ERROR_SSL:
error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED:
self signed certificate
```

Root cause: the signal exporter factories build the transport with exactly
five arguments and never read or forward the certificate variables. In
`exporter-otlp/SpanExporterFactory.php` (`buildTransport()`):

```php
return $factory->create($endpoint, $contentType, $headers, $compression, $timeout);
```

while `transport-grpc/GrpcTransportFactory::create()` accepts the values as
optional parameters:

```php
public function create(
    string $endpoint,
    string $contentType = ContentTypes::PROTOBUF,
    array $headers = [],
    $compression = null,
    float $timeout = 10.,
    int $retryDelay = 100,
    int $maxRetries = 3,
    ?string $cacert = null,
    ?string $cert = null,
    ?string $key = null,
): TransportInterface { ... }
```

Because `$cacert`/`$cert`/`$key` are always `null`, the factory ends up
calling `ChannelCredentials::createSsl(null, null, null)`, and the C-core
channel falls back to the system's default root store — so any collector that
does not present a publicly signed certificate is rejected. The same pattern
exists in `MetricExporterFactory` and `LogsExporterFactory`.

**Additional context**

- The transport layer itself works correctly when the CA is supplied: calling
  `GrpcTransportFactory::create()` directly with `$cacert` set (or using
  `ChannelCredentials::createSsl($caPem, null, null)` manually) completes the
  TLS handshake and exports successfully against both a standard gRPC server
  (opentelemetry-collector v0.160.0) and an amphp-based HTTP/2 server. So
  `ChannelCredentials::createSsl()` is not misbehaving — with a proper CA it
  trusts that CA, and with `null` it reasonably falls back to the default
  roots. The gap is entirely in the SDK wiring: the environment variable
  values never reach the transport factory.
- The identical wiring gap exists for OTLP/HTTP: `OtlpHttpTransportFactory::create()`
  also accepts `$cacert`/`$cert`/`$key`, but none of the three signal exporter
  factories reads the `OTEL_EXPORTER_OTLP_<SIGNAL>_CERTIFICATE`,
  `_CLIENT_CERTIFICATE`, or `_CLIENT_KEY` variables for any protocol. A single
  fix in the signal factories addresses both protocols.
- Pinned in this suite by `tests/TlsTest.php`: every TLS test (generic and
  per-signal, CA and client certificate) fails on the official SDK because no
  certificate variable is applied, while the same tests pass against
  `tbachert/otel-sdk`.

---

## Issue 2: Per-signal OTLP gRPC endpoints abort SDK initialization

**Describe your environment**

Same as above (PHP 8.4.25, `ext-grpc` 1.84.0, `open-telemetry/sdk` 1.15.0,
`open-telemetry/exporter-otlp` 1.4.0, `open-telemetry/transport-grpc` 1.2.0).

**Steps to reproduce**

Configure a per-signal gRPC endpoint in the standard `host:port` form:

```sh
OTEL_TRACES_EXPORTER=otlp
OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=https://127.0.0.1:4317
```

Any use of the tracer provider (e.g. `Globals::tracerProvider()`) triggers
SDK initialization, which fails.

Per the OTLP exporter specification, the gRPC endpoint option MUST accept a
URL with an `http`/`https` scheme — and the specification's own default
endpoint for gRPC is `http://localhost:4317`, which contains no path. SDK
initialization should succeed and exports should be sent to
`/opentelemetry.proto.collector.trace.v1.TraceService/Export` on that host.

Initialization aborts with (re-verified against the current releases on
2026-09-16):

```
opentelemetry: [warning] error during opentelemetry initialization: endpoint
has to contain scheme, host and path
#0 .../exporter-otlp/SpanExporterFactory.php(55):
   OpenTelemetry\Contrib\Grpc\GrpcTransportFactory->create('https://127.0.0.1:...', ...)
```

Root cause: in `SpanExporterFactory::getEndpoint()`, a per-signal endpoint is
returned as-is and never reaches the branch that appends the gRPC method:

```php
private function getEndpoint(string $protocol): string
{
    if (Configuration::has(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT)) {
        // returned as-is: for gRPC, no service method is appended here
        return Configuration::getString(Variables::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT);
    }
    $endpoint = /* generic endpoint or default */;
    if ($protocol === Protocols::GRPC) {
        // only reached when no per-signal endpoint is configured
        return $endpoint . OtlpUtil::method(Signals::TRACE);
    }
    ...
}
```

`GrpcTransportFactory::create()` then rejects the path-less URL because it
takes the RPC method from the endpoint's path and requires one:

```php
$parts = parse_url($endpoint);
if (!isset($parts['scheme'], $parts['host'], $parts['path'])) {
    throw new InvalidArgumentException('Endpoint has to contain scheme, host and path');
}
```

The fix belongs in the signal exporter factories, not in the transport
factory: the gRPC transport is signal-agnostic (registered once under the
`grpc` protocol, shared by all three signals) and takes the RPC method from
the endpoint's path, so it cannot choose one itself. The signal factories
already know their method (`OtlpUtil::method(Signals::TRACE)` etc.) and append
it for the generic endpoint; they should do the same when a per-signal gRPC
endpoint has no path (an explicitly provided path can stay as-is and become
the RPC method). The as-is behavior is shared by all three signal exporter
factories (`SpanExporterFactory`, `MetricExporterFactory`,
`LogsExporterFactory`).

**Additional context**

- With the generic endpoint (`OTEL_EXPORTER_OTLP_ENDPOINT`) the method path
  is appended and initialization succeeds; only per-signal configuration is
  broken.
- A standard deployment (e.g. a collector on `host:4317` configured via
  `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`) is therefore unusable with the gRPC
  protocol as of this version.
- Pinned in this suite by `tests/GrpcTest.php`: every test that configures a
  per-signal gRPC endpoint fails on the official SDK with the initialization
  abort above, while the same tests pass against `tbachert/otel-sdk`.


src/Contrib/Otlp/LogsExporterFactory.php
```diff
@@ -67,16 +75,31 @@ private function getCompression(): string
...
+                $insecure = OtlpUtil::getBoolVar(
+                    Variables::OTEL_EXPORTER_OTLP_LOGS_INSECURE,
+                    Variables::OTEL_EXPORTER_OTLP_INSECURE,
+                ) ?? false;
+                $endpoint = $insecure
+                    ? 'http://' . $endpoint
+                    : 'https://' . $endpoint;
+            }
+
             return $endpoint . OtlpUtil::method(Signals::LOGS);
```
Nevay, line 98: We might want to skip appending the gRPC method when a path is already present to maintain backwards compatibility for users who have included the gRPC method in their endpoint as a workaround for the initialization failure.

brettmc: Agreed, I think it would be worth adding that, with a note that it can be removed in a future breaking version

brettmc (approved):

